PR #13, opened 2018-11-01
Add script + ansible logic to unpack upgrade image

prakashsurya:
This change adds a script to this repository that can be used to unpack
an upgrade image, such that the upgrade-scripts contained in that image
can then be used to upgrade the appliance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,11 @@ package:
 	done
 
 shellcheck:
-	shellcheck etc/delphix-platform/ansible/apply
+	shellcheck \
+		etc/delphix-platform/ansible/apply \
+		etc/delphix-platform/upgrade/unpack-image
 
 shfmtcheck:
-	! shfmt -d etc/delphix-platform/ansible/apply | grep .
+	! shfmt -d \
+		etc/delphix-platform/ansible/apply \
+		etc/delphix-platform/upgrade/unpack-image | grep .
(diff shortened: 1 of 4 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/180/)

prakashsurya:
Copied this script to an existing VM, and ran the following:
```
$ sudo /etc/delphix-platform/upgrade/unpack-image internal-dev.upgrade.tar.gz
Progress increment: 16:34:46:085881594+0000, 10, Extracting upgrade image
Progress increment: 16:36:04:311588304+0000, 40, Verifying upgrade image
Progress increment: 16:36:04:320422666+0000, 100, Unpacking successful
```

prakashsurya:
cc: @dtzeng @pandar00 @grwilson

I think we should move the existing `dx_unpack` script out of the virtualization package, and into this new platform package. This way, we can use this script and the appliance-build upgrade scripts to upgrade an appliance regardless of if it has the virtualization package installed (e.g. our dcenter variant). Let me know what y'all think of this idea.

This change adds the new "unpack" script to this repository, and then after this lands, I can update the virtualization repository to remove its unpack script.

prakashsurya:
> Downside is that upgrade related scripts are distributed into different repos but I think the benefit you mentioned is greater.

Yea, it's a little awkward to have this "unpack" script live here, where all the other upgrade related scripts live in the appliance-build repository.

I'm open to trying to find a way to have them all reside together, but we do have a special requirement for the "unpack" script, in that it needs to be installed on the appliance prior to downloading/extracting the upgrade image (at which point, the other upgrade related scripts become available, since they're stored in the image itself).

I don't really like how we accomplish this on illumos, but I'm open to considering it for linux if y'all think that'd be a better solution.

PR #14, opened 2018-11-02
Be explicit about using the "delphix" username

prakashsurya:
This change modifies how the appliance user is configured by the
delphix-platform service. Previously we allowed the username and the
user's password to be configured using environment variables. While this
seemed like a good idea at the time, as it allowed us flexibility while
we were early in our Linux conversion work, it makes things considerably
more difficult on upgrade and migration.

Thus, this change removes this flexibility and now hardcodes the delphix
user into the delphix-platform ansible role, as well as the associated
cloud-init file (that also needs to know the appliance's username).

Additionally, the delphix-plaform service will no longer attempt to
configure the appliance user's password. Instead, configuration of the
user's password will (now) need be accomplished either by the initial
appliance build process, or by the upgrade and/or migration logic.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/189/)

see also: https://github.com/delphix/appliance-build/pull/144

prakashsurya:
cc: @pzakha @dtzeng @pandar00 @grwilson

PR #16, opened 2018-11-09
Pass in the name of the root filesystem to ansible playbooks

sebroy:
This change passes in the name of the root filesystem to ansible
playbooks to enable them to make modifications to root filesystem
properties without implementing root filesystem discovery logic within
ansible.

I tested this by applying this diff to an already installed apply script, and
modifying the virtualization package's ansible role to set a zfs property
on the filesystem.

### DIFF
--- a/etc/delphix-platform/ansible/apply
+++ b/etc/delphix-platform/ansible/apply
@@ -19,6 +19,7 @@ set -o pipefail
 
 SOURCE_DIRECTORY="$(readlink -f "${BASH_SOURCE%/*}")"
 DLPX_ANSIBLE_DIRECTORY="${DLPX_ANSIBLE_DIRECTORY:-$SOURCE_DIRECTORY}"
+ROOT_FILESYSTEM=$(zfs list -Ho name /)
 
 function find_playbooks() {
 	if ! find "$DLPX_ANSIBLE_DIRECTORY" \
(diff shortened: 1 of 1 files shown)


etc/delphix-platform/ansible/apply
```diff
@@ -19,6 +19,7 @@ set -o pipefail
 
 SOURCE_DIRECTORY="$(readlink -f "${BASH_SOURCE%/*}")"
 DLPX_ANSIBLE_DIRECTORY="${DLPX_ANSIBLE_DIRECTORY:-$SOURCE_DIRECTORY}"
+ROOT_FILESYSTEM=$(df / --output=source | tail +2)
```
grwilson: Could we use `zfs list -Ho name  /` instead?

sebroy: Yes, I like that better.

grwilson:
This isn't the same as the verify container but I am running with nspawn and I get this:

pzakha:
bors delegate+

etc/delphix-platform/ansible/apply
```diff
@@ -34,6 +35,7 @@ function apply_playbook() {
 	if ! ansible-playbook -vvv \
 		-c "$DLPX_ANSIBLE_CONNECTION" \
 		-i "$DLPX_ANSIBLE_INVENTORY" \
+		-e "root_filesystem=${ROOT_FILESYSTEM}" \
```
prakashsurya: I'm a late to the party, but we might want to start prefixing ansible variable names with `dlpx_` or `delphix_` since (AFAIK) there isn't any namespacing built into the variable names, so we could potentially conflict with other variables. E.g. for built in variables, iirc, they're prefixed with `ansible_`.. so in this case, something like `delphix_root_filesystem` is what I'm thinking.

This already landed, and is probably fine, but just wanted to make a note if we end up adding more variables in the future.

PR #17, opened 2018-11-09
Get delphix-platform ready for linux-pkg

pzakha:
This change makes the repository ready for linux-pkg and also
allows creating per-platform flavours of the package.

Note that this is a breaking change, and should be pushed along with changes in other repositories.
See main review http://reviews.delphix.com/r/44644 for more details.

**ADDITIONAL DETAILS**
delphix-platform repository now builds 2 packages for each platform:

 - delphix-platform-PLATFORM
 - delphix-kernel-KVERS

Where PLATFORM is something like "aws" or "generic" and KVERS is something like
"4.15.0-38-generic" or "4.15.0-1023-aws".

delphix-kernel-KVERS will provide virtual-package delphix-kernel-PLATFORM.
It will depend on the ubuntu kernel image and on all our kernel modules.
e.g.:
Package: delphix-kernel-4.15.0-1023-aws
Virtual Package: delphix-kernel-aws
Depends: zfs-modules-4.15.0-1023-aws, connstat-modules-4.15.0-1023-aws,
         linux-image-4.15.0-1023-aws

delphix-platform-PLATFORM will now depend on delphix-kernel-PLATFORM and on
linux-image-PLATFORM.
e.g.:
Package: delphix-platform-aws
Depends: delphix-kernel-aws, linux-image-aws

During the build, we will just include the proper version of delphix-platform
for the platform on which we are building, e.g.: delphix-platform-aws. This
should install the only version of delphix-kernel-aws available, e.g.
delphix-kernel-4.15.0-1023-aws, as part of the dependencies of delphix-platform.

Following the aws example, during upgrade, we will install a new version of
delphix-entire-aws, which will depend on a newer version of delphix-kernel-KVERS.
Since delphix-kernel-aws is a virtual package, we can have
delphix-kernel-4.15.0-1023-aws and delphix-kernel-4.18.0-1005-aws installed
together without any conflicts. After the system has rebooted, we can uninstall
the older version, delphix-kernel-4.15.0-1023-aws, and its dependencies.

Note that until appliance-build supports making platform-dependent images, we
only build packages for platform 'generic' in linux-pkg. Since
dlephix-platform-generic provides virtual package 'delphix-platform', the
package list in appliance-build doesn't need to be modified.


pzakha:
The `configure.sh` script must be executed prior to building the packages to have the appropriate fields substituted in for the KVERS and PLATFORM tokens.
You can look at the `make package` step in the Makefile.
For `linux-pkg`, we do something similar in the `build()` hook of `packages/delphix-platform/config.sh`:
 https://github.com/delphix/linux-pkg/pull/1/files#diff-765ff05e8e5bdc8bbd0e76ecf776e9fa

prakashsurya:
Why do we want to add the complexity of hypervisor/kernel specific packages to this repository, vs. adding hypervisor/kernel specific "stuff" outside of this repository? Did I overlook that explanation?

pzakha:
> Why do we want to add the complexity of hypervisor/kernel specific packages to this repository, vs. adding hypervisor/kernel specific "stuff" outside of this repository? Did I overlook that explanation?

@prakashsurya `delphix-platform` has dependencies on the kernel and ZFS, which are going to be versioned by platform, so we need to put some sort of logic here to pick the right packages. This also seems like a good place to add platfrom-specific changes in the future.

prakashsurya:
> `delphix-platform` has dependencies on the kernel and ZFS

This repository has no more dependencies on the kernel/zfs than packages like virtualization (probably less), which doesn't have this complexity. I'm not sure I agree with this approach based on this motivation.

> This also seems like a good place to add platfrom-specific changes in the future.

I'm not sure I agree with the idea that we should add complexity to solve a problem that doesn't yet exist; and I'm not sure this is the best way to solve the problem even if it did exist (e.g. we have a similar problem on illumos, and have solved it without hypervisor specific packages).

pzakha:
All this logic feels fairly straight-forward to me, but if you have ideas to simplify it I'm open to suggestions.

prakashsurya:
> All this logic feels fairly straight-forward to me

It's not that the code itself isn't straight-forward, it's the motivation for adding it that's less clear...

> but if you have ideas to simplify it I'm open to suggestions.

Remove it? I'm not trying to be confrontational (sorry if it comes off that way), it's just not clear to me why this specific repository needs specific packages that depend on a specific kernel and/or hypervisor. It's simply dropping some (kernel/hypervisor independent) files into `/etc`.. there's nothing in it that depends on a specific kernel, or on a specific hypervisor. Further, the PR description doesn't mention anything about the problem per-platform packages solves.

I might not be seeing the "bigger picture" where this becomes more clear, so I'm open to discuss over DM or something.

Simply looking at this repository in isolation, it's just not really clear why a single platform-independent package is insufficient. E.g. I view this similar to something like `ubuntu-minimal`, which is also platform-independent.

pzakha:
We need a place to put the proper kernel dependencies for each platform, so that for instance appliance-build can install the "aws" kernel modules for the "aws" platform. If we just specify that `delphix-platform` depends on `zfs-modules` then we will need to have some mechanism that installs `zfs-modules-aws` before we try to install delphix-platform, otherwise apt will not know which zfs modules package to install when trying to satisfy delphix-platform's dependency. 
I figured this would be a good candidate package as it's essentially the "base-files" package for the delphix appliance. I can provide you more details over a DM.

pzakha:
I've addressed the review feedback and rebased on top of existing changes.
In this new iteration, delphix-platform will be built by the linux-pkg framework using:
```
./scripts/docker-run.sh make package VERSION="<passed from linux-pkg>'
```
The `delphix-kernel` package is now an independent package, with a review available here:
https://github.com/delphix/delphix-kernel/pull/1

Makefile
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+VERSION := 0.0.1
```
prakashsurya: Can we still set the version to `$(date '+%Y.%m.%d.%H')` by default?

It's probably also worth adding a comment to explain that we do it this way so an external build system can easily configure the version. Otherwise it'd be hard to know why we don't just use the string "0.0.1" below in the "dch" command.

Makefile
```diff
@@ -23,18 +25,12 @@
...
-
-	@for ext in buildinfo changes deb; do \
-		mv -v ../delphix-platform_*_amd64.$$ext artifacts; \
-	done
+	rm -rf artifacts
+	mkdir artifacts
+	dch --create --package delphix-platform -v $(VERSION) \
+			"Automatically generated changelog entry."
+	dpkg-buildpackage -b -uc -us
+	mv ../delphix-platform*.deb artifacts/
```
prakashsurya: why don't we move the other files (i.e. the non-`.deb` files) into the artifacts directory anymore? It's not clear why the prior `mv` commands needed to be changed.

Makefile
```diff
@@ -23,18 +25,12 @@
...
-	dpkg-buildpackage
-
-	@for ext in dsc tar.xz; do \
-		mv -v ../delphix-platform_*.$$ext artifacts; \
-	done
-
-	@for ext in buildinfo changes deb; do \
-		mv -v ../delphix-platform_*_amd64.$$ext artifacts; \
-	done
+	rm -rf artifacts
```
prakashsurya: why do we need to remove the `artifacts` directory and recreated it?

I think I liked it better when the directory existed in the repository, and `make package` didn't try to remove existing build products (i.e. how it currently works). We could add a `make clean` target if we need to clear that out (and `git clean -dxf` is insufficient).

Is there a specific reason for changing this?

pzakha: I felt like it simplified things since we don't need an extra `.gitignore` file in there, but it's not something I had a strong opinion so I reverted it back to what it was.

Makefile
pzakha: `linux-pkg` doesn't care about the other deliverables since we do not consume them during the build. That said I've reverted back the logic to what it was for the independent build. `linux-pkg` just doesn't use the other artifacts.

pzakha:
ab-pre-push is here: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/278/
Those new changes can be pushed without breaking the existing build.

pzakha:
Ah nvm, the added dependencies on delphix-extra and delphix-kernel will break the build.

PR #19, opened 2018-11-20
Remove the ubuntu welcome banner, help info, and last login when ssh-ing

waynefranklin:
Tested by scp-ing updated main.yml onto target vm and rebooting.

### DIFF
--- a/etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -168,6 +168,20 @@
     replace: '#\1'
 
 #
+# Ssh leads to the CLI, not bash, so let's remove all the linuxy shell goodies,
+# like last-login, "welcome to ubuntu", and help messages. This makes linux and
+# illumos look the same, too.
+#
+- replace:
+    dest: /etc/ssh/sshd_config
+    regexp: '^#?[\s]*PrintLastLog.*$'
+    replace: 'PrintLastLog no'
+- replace:
+    dest: /etc/pam.d/sshd
+    regexp: '^(session[\s]+optional[\s]+pam_motd\.so.*)$'
+    replace: '#\1'
+
+#
 # Enable SNMP client tools to load MIBs by default.
 #
 - replace:


etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -167,6 +167,22 @@
     regexp: '^(Subsystem.*sftp.*)'
     replace: '#\1'
 
+#
+# Remove the last-login message when ssh-ing.
+#
+- replace:
+    dest: /etc/ssh/sshd_config
+    regexp: '#PrintLastLog yes'
```
prakashsurya: I think we want to use something like this: `^#?PrintLastLog` .. This way, it'll modify the line if the field is commented out or not, and regardless of the current setting.

Otherwise, I think we could wind up with two lines, the one we insert and the line that would have already existed.. e.g. if the field isn't commented out (i.e. doesn't start with `#`), or has a value other than `yes`.

What do you think?

etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -198,3 +214,4 @@
     line: '{{ item.key }}="{{ item.value }}"'
   with_items:
     - { key: 'JAVA_HOME', value: '/usr/lib/jvm/oracle-java8-jdk-amd64' }
+
```
prakashsurya: is this additional line intentional?

etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -167,6 +167,22 @@
     regexp: '^(Subsystem.*sftp.*)'
     replace: '#\1'
 
+#
+# Remove the last-login message when ssh-ing.
```
prakashsurya: Rather than simply repeating what the code already states, it'd be more useful to explain why we want to do this configuration/removal.

etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -167,6 +167,22 @@
...
+#
+- replace:
+    dest: /etc/ssh/sshd_config
+    regexp: '#PrintLastLog yes'
+    replace: 'PrintLastLog no'
+
+#
+# Remove the ubuntu motd banner and related dross.
```
prakashsurya: Similar to above, rather than simply repeating what the code already states, it'd be more useful to explain why we want to do this configuration/removal.

etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
waynefranklin: No. I'll take care of it.

waynefranklin:
Updated the commit per comments. Let me know how it looks now.

waynefranklin:
Thanks, guys. And I'll do the `git-ab-pre-push` next time.

bors r+

PR #21, opened 2018-11-27
Apply configuration once, to reduce bootup latency

prakashsurya:
Problem
=======

Using "systemd-analyze" to inspect the bootup latency, we can quickly
see the "delphix-platform" adds considerable latency to bootup time.

Since the "delphix-mgmt" service depends on "delphix-platform", any time
required to execute the "delphix-platform" service directly affects the
latency for the "delphix-mgmt" to come up after a reboot.

Here's an example from a system without this change:

    $ systemd-analyze blame | head -n2
        1min 20.155s delphix-mgmt.service
             18.660s delphix-platform.service

From this, we see it takes over 18s to execute the "delphix-platform"
service; this translates to 18 seconds of additional downtime for the
"delphix-mgmt" service.

Solution
========

To address this issue, this change modifies the "delphix-platform"
service to only run the ansible configuration a single time. This will
allow the service to configure the system during the first boot (which
will obviously entail some bootup latency), but will prevent the service
from running the same ansible configuration on subsequent boots
(eliminating the additional bootup latency caused by this service for
these subsequent boots).

Here's an example from a system with this change:

    $ sudo systemd-analyze blame | head -n2
        1min 22.222s delphix-mgmt.service
             10.496s rtslib-fb-targetctl.service

    $ sudo systemd-analyze blame | grep delphix-platform
               123ms delphix-platform.service

In this example, the "delphix-platform" now adds a very minimal amount
of additionally latency to the bootup time, because the service doesn't
have to re-apply the ansible configuration that's already been applied.

Caveats
=======

The caveat to this approach being, "delphix-platform" service will no
longer automatically apply new configuration changes after an upgrade of
the Delphix appliance.

Since we actually do want re-apply the ansible configuration after an
upgrade (e.g. because the ansible logic may have been changed due to the
upgrade), the logic/scripts we use to orchestrate the upgrade will need
to be aware of this solution, and force it to run (e.g. by removing the
/etc/delphix-platform/ansible-done file as part of the upgrade).

### DIFF
--- a/etc/delphix-platform/ansible/apply
+++ b/etc/delphix-platform/ansible/apply
@@ -42,6 +42,36 @@ function apply_playbook() {
 	fi
 }
 
+#
+# To reduce the bootup latency and the latency to start services that
+# depend on this service, we only want to apply the configuration a
+# single time. This way, the configuration will be applied on the first
+# boot, but not on subsequent boots. This will allow the bootup time for
+# the subsequent boots to be quicker.
+#
+# Thus, if this "ansible-done" file is present, this signifies that the
+# ansible configuration has already been applied, so there's no need to
+# apply this configuration again.
+#
+# We don't use Systemd's "ConditionPathExists" directive to disable the
+# service after it's been run once, for a couple of reasons:
+#
+# 1. That would cause the service to enter a "failed" state when the
+#    file exists. This isn't the behavior we want, since it then makes
+#    it more difficult for other services to depend on this service
+#    running successfully.
+#
+# 2. When the file exists, restarting "delphix-platform" would not
+#    result in restarting services that depend on "delphix-platform". We
+#    rely on restarts to "delphix-platfrom" causing restarts to all
+#    dependent services, so we need to ensure this functionality
+#    continues to work, even if the ansible configuration doesn't have
+#    to be re-applied.
+#
+if [[ -f /etc/delphix-platform/ansible-done ]]; then
+	exit 0
+fi
+
 if [[ -z "$DLPX_ANSIBLE_DIRECTORY" ]]; then
 	echo "DLPX_ANSIBLE_DIRECTORY is unset." >&2
 	exit 1
@@ -65,3 +95,22 @@ fi
 find_playbooks | while read -r playbook; do
 	apply_playbook "$playbook"
 done
+
+#
+# This script is executed as part of the appliance-build process so that
+# configuration that's required before the first boot (e.g. bootloader)
+# can be applied using the same mechanism as configuration that can be
+# applied after the first boot.
+#
+# Thus, when this script is executed in this way, via appliance-build,
+# we need to be careful not to try to create this file. Otherwise we'd
+# modify the root filesystem of the system running the build, rather
+# than the Delphix appliance's root filesystem that's being configured.
+#
+# We detect if this script is being executed via appliance-build or not,
+# by inspecting this environment variable. It will be set to "chroot"
+# only by appliance-build; otherwise it should contain the value "local".
+#
+if [[ "$DLPX_ANSIBLE_CONNECTION" != "chroot" ]]; then
+	touch /etc/delphix-platform/ansible-done
+fi
--- a/lib/systemd/system/delphix-platform.service
+++ b/lib/systemd/system/delphix-platform.service
@@ -22,7 +22,7 @@ Before=rsync.service
 [Service]
 Type=oneshot
 ExecStart=/etc/delphix-platform/ansible/apply
-ExecReload=/bin/true
+ExecReload=/bin/rm -f /etc/delphix-platform/ansible-done
 RemainAfterExit=yes
 
 #


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/260/)

etc/delphix-platform/ansible/apply
```diff
@@ -42,6 +42,42 @@ function apply_playbook() {
...
+#    continues to work, even if the ansible configuration doesn't have
+#    to be re-applied.
+#
+# Also worth noting, consumers of this service (e.g. upgrade, factory
+# reset, etc.) are aware of this file and it's implications. Thus, if we
+# ever need to change this file's path, we'll likely have to also update
+# all of the consumers of this service; since these consumers may rely
+# on removing this file, in order to force this service to re-run.
+#
+if [[ -f /etc/delphix-platform/ansible-done ]]; then
```
sebroy: I think this would be right right place to check for DLPX_FACTORY_RESET and remove the ansible-done file if the variable is true.

etc/delphix-platform/ansible/apply
```diff
@@ -42,6 +42,42 @@ function apply_playbook() {
 	fi
 }
 
+#
+# To reduce the bootup latency and the latency to start services that
+# depend on this service, we only want to apply the configuration a
+# single time. This way, the configuration will be applied on the first
```
sebroy: The ansible roles also need to be applied on upgrade (since the ansible logic can change at that point). How will that get accomplished?

prakashsurya: I'll open a PR to update the upgrade-scripts to remove this file when we do an upgrade. I can make that change, and open the appliance-build PR prior to landing this; so that can be evaluated before we land any of the changes involved with this approach.

etc/delphix-platform/ansible/apply
prakashsurya: I disagree; the `DLPX_FACTORY_RESET` variable is an app-gate concept, so leaking that into this repository doesn't seem like a good design. Currently, this repository knows nothing about that variable.

Further, I think we need a way to re-run the ansible configuration, without invoking the app-gate's factory reset logic. E.g. on upgrade, we want to run the ansible code without invoking factory reset. If we only had a single tunable, `DLPX_FACTORY_RESET`, I'm not sure how we could invoke the logic unrelated to factory reset, without also invoking the factory reset logic (and breaking the customer's engine).

What do you think?

etc/delphix-platform/ansible/apply
prakashsurya: I've opened this: https://github.com/delphix/appliance-build/pull/160

etc/delphix-platform/ansible/apply
sebroy: I understand your point of view.

My perspective on factory reset when making the suggestion was that I viewed setting DLPX_FACTORY_RESET on delphix-platform as the only way to perform a factory reset of the entire appliance today. The variable is exposed by the delphix-platform service insofar as there is no other way to do a factory reset.

I think your point is that this design isn't a good one, and we should separately re-design factory reset so that it doesn't use this environment variable, and also doesn't rely on being driven by packages' ansible roles. Is that right? I'm fine with that, let's create a jira issue for that.

prakashsurya: My point is this, delphix-platform doesn't have a concept of "factory reset". This concept is created by the app-gate, and doesn't exist outside of the app-gate. Any appliance we create that doesn't have the app-gate on it (e.g. dc1), has no concept of "factory reset", even if it has the delphix-platform installed.

Further, the way the app-gate implements factory reset on Linux is not very elegant, or extensible (as you've mentioned). Thus, I'd like to ignore how it's currently implemented (i.e. forget about the specific variable) for now, and think more broadly w.r.t. the concepts we're trying to accomplish.

If we want to extend delphix-platform to have a concept of "factory reset" (which is what you're suggesting, by checking for the `DLPX_FACTORY_RESET` variable here), such that that other consumers can make use of it (e.g. app-gate uses this new concept to implement it's own factory reset logic, masking uses it do the same, etc.), I'm OK with that. But, it's more work that would need to be scoped and implemented (since I'm not sure just porting the current implementation is the best way to do it).

Additionally, in my mind, that functionality is orthogonal to implementing this "configure once" concept. We could implement "configure once" without implementing "factory reset" in delphix-platform (which I'm trying to do in this PR), and could also implement "factory reset" without implementing "configure once".

What I think you're suggesting is that we implement both "configure once" and "factory reset" in delphix-platform, because once the two are implemented in delphix-plaform, consumers of delphix-platform become more simple (e.g. the app-gate logic you commented on could be more simple).

I'm OK going that route, it's just more work (both design and implementation) than I was expecting to tackle with this PR. So, if we want to implement both concepts simultaneously in delphix-platform because ultimately that'll lead to a better overall solution, I'm OK with that, but it'll have to be something that is left to be implemented later (if at all).

I see two ways to proceed:

1. Close this PR and wait for "configure once" and "factory reset" to be implemented in delphix-platform at the same time.

2. Move forward with this PR, and only tackle the "configure once" semantics which I was going for; potentially opening a JIRA issue to tackle "factory reset" in delphix-platform later. As pointed out in the app-gate review, this might leave is with a sub-optimal solution given the current "factory reset" implementation.

I'm happy to go with either approach, honestly. As long as we're all on the same page as to which we'd like to go with. Which would you prefer?

What I don't want to do is implement a poor solution for factory reset in delphix-platform, which is then difficult to modify and improve later, because it's consumed by multiple different projects (and would require a multi-repository and/or multi-team flag day to reconcile). I think we might end up in this situation if we just simply checked for `DLPX_FACTORY_RESET` here.

sebroy: Yep, I think we can move on, we also hashed this out in the dlpx-app-gate review. I'm fine with the change as-is and our handling factory reset changes separately.

prakashsurya:
`git-ab-pre-push -v internal-dev` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/269/)

debian/postinst
```diff
@@ -22,6 +22,8 @@ case $1 in
 	systemctl enable systemd-networkd.service
 
 	systemctl disable walinuxagent.service
+
+	systemctl reload delphix-platform.service
```
sebroy: Should this happen before we re-enable the service at line 21?

prakashsurya: I don't think so, because enabling the service doesn't cause the service to start running. It just makes it eligible to run the next time the system boots up.

So, we install the package, enable the service, then ensure the "ansible-done" file is removed, then wait for the service to be started by something else (e.g. the upgrade scripts, or a reboot, or whatever).

Does this make sense?

sebroy: Yes, thanks for clarifying.

PR #23, opened 2018-12-04
Don't modify journald.conf, use ".d" directory

prakashsurya:


prakashsurya:
cc: @grwilson @shartse

etc/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -91,7 +91,7 @@
 # also can make the collection of support bundles more useful.
 #
 - lineinfile:
-    dest: /etc/systemd/journald.conf
+    dest: /etc/systemd/journald.conf.d/override.conf
```
sebroy: Is it not preferable to ship a packaged override.conf file for this rather than ansible so that we can make changes to that file without adding ansible logic to undo existing changes?

prakashsurya: definitely, thanks! idk why I didn't think of that to begin with.

prakashsurya:
For testing, I built and installed this package on a DE, manually removed settings in `/etc/systemd/journald.conf`, restarted the `systemd-journald` service, and verified the max setting applied:
```
$ sudo systemctl status systemd-journald | tail -n 1
Dec 04 20:04:16 ip-10-110-221-169 systemd-journald[6963]: System journal (/var/log/journal/456b088d36064ca18d52744a2e6b8520) is 252.5M, max 250.0M, 0B free.
```

I also bumped the max to 500M as another test, and verified this new max applied (i.e. to verify the value in this new file was taking effect):
```
$ sudo systemctl status systemd-journald | tail -n 1
Dec 04 20:07:10 ip-10-110-221-169 systemd-journald[8825]: System journal (/var/log/journal/456b088d36064ca18d52744a2e6b8520) is 252.5M, max 500.0M, 247.4M free.
```

bors r+

PR #24, opened 2018-12-13
Create platform-specific packages

jgallag88:
Create an instance of the delphix-platform package for each platform
that we support. The package depends on the appropriate version of
delphix-kernel, and contains other platform-specific dependencies where
appropriate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,11 @@
 # limitations under the License.
 #
 
+ALL_PLATFORMS := aws azure esx gcp kvm
+
 #
 # The version field defaults to a timestamp. Note that it can be
-# overridden by running: make package VERSION="<custom version>"
+# overridden by running: make packages VERSION="<custom version>"
 #
 VERSION := $(shell date '+%Y.%m.%d.%H')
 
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -28,17 +30,20 @@ VERSION := $(shell date '+%Y.%m.%d.%H')
 
 check: shellcheck shfmtcheck
 
-package:
+packages: $(addprefix package-,$(ALL_PLATFORMS))
```
prakashsurya: since we're no longer supporting "make package", and replacing it with "make packages".. we also need to update the README to reflect this.. as well as the comment towards the top of this file, and the one in `debian/rules`:
```
± g g "make package"
Makefile:# overridden by running: make package VERSION="<custom version>"
README.md:    $ ./scripts/docker-run.sh make package
debian/rules:   # infinite loop would occur: make package -> make package -> ...
```
Additionally, `.travis.yml` should be updated to reflect this.

This will also require a flag-day with the devops-gate, since we run `make package`, which will need to change to `make packages`.

Makefile
```diff
@@ -28,18 +30,18 @@ VERSION := $(shell date '+%Y.%m.%d.%H')
...
```
pzakha: We'll also want to update `linux-pkg` to call `make packages` instead of `make package`

jgallag88: 👍 https://github.com/delphix/linux-pkg/pull/4

debian/rules
```diff
@@ -62,6 +67,14 @@ DEPENDS += ubuntu-dbgsym-keyring,
 #
 DEPENDS += delphix-extra,
 
+# Platform-specific dependencies
+DEPENDS.aws =
+DEPENDS.azure = walinuxagent,
```
grwilson: Do you need to make changes to the `postinst` file since it will try to disable a service that may or may not be installed or does this error out silently and just works?

jgallag88: The `postinst` script does fail, but it turns out the failure is ignored by live-build. Rather than have this error in the build, I've just removed walinuxagent for now, since my understanding is that it is unused even on Azure. Once we get it working correctly on Azure we can add the package back.

jgallag88: I see, so while we aren't using the daemon, we still want the package for the files it provides. I can add this back in.

grwilson: I think the correct answer is to remove the "disable" of the waagent from the script since it will only be installed on the VHD variant and it can be left in an enabled state. We still need to fix the cloudinit portion of it but it's not worse than the current state of things. What do you think?

prakashsurya: I've just recently learned about systemd's concept of "vendor presets". We could leverage this; we can set services to be disabled by default even if they don't exist (e.g. use the same preset value for all cloud platforms). See [here](https://www.freedesktop.org/software/systemd/man/systemd.preset.html) for more details.

Arguably we should do this for all services, and stop calling `systemctl` from the `postinst` script (I'm working on doing this for the virtualization package right now).

prakashsurya: Here's an example of what I mean (it's untested): https://github.com/delphix/delphix-platform/pull/34

jgallag88: @grwilson I'm not sure what the consequences of leaving it enabled are, but if we can do that, that seems like the simplest option. If not, I can easily disable it using the method Prakash mentioned, or by just disabling it on Azure, making use of DPKG_MAINTSCRIPT_PACKAGE in the post install script to figure out what platform we are on.

grwilson: I think leaving it enabled should be fine since we still have to do some work to get the azure agent working correctly. Leaving it enabled won't make things worse.

jgallag88:
bors +r

PR #32, opened 2019-01-11
Better adhere to the Filesystem Hierarchy Standard

prakashsurya:
This change moves where some files are stored, in an attempt to better
adhere to the Filesystem Hierarchy Standard. Thus, executable scripts
have been moved out of "/etc" and into "/usr/bin". Further, files that
represent state rather than configuration for the delphix-platform
service (e.g. it's ansible configuration files), have also been moved
out of "/etc" and into "/var/lib/delphix-platform".

Additionally, since other software might still attempt to install
ansible configuration into the old location, "/etc/delphix-platform", we
still temporarly look in that directory and apply any configuration from
there that might exist. Eventually, after updating all consumers, we'll
remove this backwards compatibility logic; this helps avoid flag days.


prakashsurya:
cc: @pzakha

debian/rules
```diff
@@ -125,4 +125,5 @@ override_dh_auto_test:
 override_dh_auto_install:
```
pzakha: FYI, you can replace this by a `debian/install` file with:
```
etc/
lib/
usr/
var/
```

var/lib/delphix-platform/ansible/apply
```diff
@@ -29,6 +29,20 @@ function find_playbooks() {
...
+	#
+	# Temporarily, we need to maintain backwards compatibility with
+	# software that still installs ansible configuration into the
+	# old directory. Once all consumers can be updated to install
+	# into the new directory, we can remove this logic.
+	#
+	if ! find "/etc/delphix-platform/ansible" \
+		-maxdepth 2 \
+		-mindepth 2 \
+		-name playbook.yml | sort -n; then
```
pzakha: This will result in 2 lists that are sorted locally, i.e. everything in `/etc/delphix-platform/ansible` will be executed after DLPX_ANSIBLE_DIRECTORY. Is this going to cause issues?

prakashsurya: > Is this going to cause issues?

No.

Long term, virtualization will get updated to install in `/var/lib/delphix-platform` and we'll remove this backwards compatibility layer.

Short term, the delphix-platform logic will be in `/var/lib/delphix-platform` using a prefix of `10-`, and the virtualization logic will be in `/etc/delphix-platform` using a prefix of `20-`. So, currently delphix-platform configuration is run first, while virtualization configuration is run second; this is due to their respective prefix numbers (`10-` and `20-`). Thus, the ordering of this won't practically be a problem, since we'll continue to execute the same configuration in the same order.

That is, unless I'm overlooking something.

debian/prerm
```diff
@@ -25,6 +25,7 @@ case $1 in
 	# so we can't reliably restore the system back to that prior state.
 	#
 
+	systemctl reload delphix-platform.service
```
pzakha: Can you comment on this?

prakashsurya: Sure.. When a package is uninstalled, it'll attempt to remove all files/directories that that package owned; in this case, it'll attempt to remove `/var/lib/delphix-platform`.
                                                                                
If the package manager detects a file in that directory that isn't owned by the package, it won't remove the directory (presumably, in fear of removing some unrelated file). In this case, it's possible for the `/var/lib/delphix-platform/ansible-done` file to exist, which isn't owned by the package (even though this package's service will create the file). So, to remedy this, and ensure the `/var/lib/delphix-platform` directory is removed on uninstall, we should ensure that `ansible-done` file is removed first.
                                                                                
This is an existing issue, it's just not practically a problem since we generally don't remove the `delphix-platform` package. I just noticed it when doing some manual testing of this PR.
                                                                                
Another solution would be to store the `ansible-done` file in some other directory, and leave it there if it exists on uninstall; but I think it's better to just remove it.

prakashsurya: Sure.. When a package is uninstalled, it'll attempt to remove all files/directories that that package owned; in this case, it'll attempt to remove `/var/lib/delphix-platform`.
                                                                                
If the package manager detects a file in that directory that isn't owned by the package, it won't remove the directory (presumably, in fear of removing some unrelated file). In this case, it's possible for the `/var/lib/delphix-platform/ansible-done` file to exist, which isn't owned by the package (even though this package's service will create the file). So, to remedy this, and ensure the `/var/lib/delphix-platform` directory is removed on uninstall, we should ensure that `ansible-done` file is removed first.
                                                                                
This is an existing issue, it's just not practically a problem since we generally don't remove the `delphix-platform` package. I just noticed it when doing some manual testing of this PR.
                                                                                
Another solution would be to store the `ansible-done` file in some other directory, and leave it there if it exists on uninstall; but I think it's better to just remove it.

pzakha: I see, so the real purpose of this is to remove the `ansible-done` file.
Do you think we should just do that explicitly here, instead of relying on `systemctl reload`, since the link between the two isn't obvious when reading `man systemctl`:
```
       reload PATTERN...
           Asks all units listed on the command line to reload their configuration. Note that this will reload the service-specific configuration, not the unit configuration file of systemd. If you want systemd
           to reload the configuration file of a unit, use the daemon-reload command. In other words: for the example case of Apache, this will reload Apache's httpd.conf in the web server, not the
           apache.service systemd unit file.
```

prakashsurya: Yea, we could do that.

I thought that might leak the abstraction, but since it's all still contained in the same package, I think it's reasonable just to use `rm -f` here. That'll be more reliable if/when removing the package from a `chroot`, too.

prakashsurya: I made this change in my latest revision.

PR #33, opened 2019-01-17
Use cron.daily to manage "/var/log/delphix-upgrade"

prakashsurya:


prakashsurya:
This is related to the changes being proposed here: https://github.com/delphix/appliance-build/pull/196

prakashsurya:
> Other logs in /var/log are managed and kept in check using logrotate. Is there a reason why we are implementing this using this custom mechanism rather than logrotate?

Yes. AFAIK, logrotate can't easily be used to simply delete old files. I did some google searching, looked at the manpage for `logrotate.conf`, and wasn't able to determine how to configure logrotate to simply delete files older than X amount of days.

From what I could gather, logrotate is only capable of rotating files, and then deleting the files that it has created (i.e. the rotated files). It can't delete files that will be created by some other software, and thus can't delete the `.tar.gz` and `.json` files without first rotating them. We might be able to configure it to rotate our files to `.1` first, and then delete those `.1` files when they're "old enough", but I didn't find a set of options that I was happy with (I tried various permutations of `maxage`, `rotate`, `maxsize`, etc.), at least compared to how simple this cron script is.

I'm not super familar with logrotate though, so if you (or anybody else) can help me accomplish this with logrotate, I'm OK using that instead.

shartse:
I had basically the same frustrations as Prakash when I was working on rotating `pg_log` - it's annoying hard to say "delete any files older than X". I settled on using cron as well.

prakashsurya:
@sebroy are you OK with this change as is?

sebroy:
Yes, I'm OK with it, just needed that little bit of extra context on the design choice.

PR #36, opened 2019-01-23
LX-1495 Tune TCP

grwilson:


grwilson:
Tony has the changes but I didn't want to wait for the performance analysis since the goal was to match the illumos settings. We can address further enhancement when we start the performance investigations.

etc/sysctl.d/50-override.conf
```diff
@@ -55,6 +55,14 @@ net.ipv4.tcp_keepalive_time = 15
 net.ipv4.tcp_keepalive_intvl = 15
 net.ipv4.tcp_keepalive_probes = 4
 
+#
+# Increase the global network and TCP buffer sizes.
+#
+net.core.rmem_max = 16777216
+net.core.wmem_max = 16777216
+net.ipv4.tcp_wmem = 4096 16777216 16777216
```
pzakha: Why is there 3 values there?

etc/sysctl.d/50-override.conf
```diff
@@ -55,6 +55,14 @@ net.ipv4.tcp_keepalive_time = 15
 net.ipv4.tcp_keepalive_intvl = 15
 net.ipv4.tcp_keepalive_probes = 4
 
+#
+# Increase the global network and TCP buffer sizes.
```
prakashsurya: This comment isn't very useful; it just re-states what the "code" already tells us. It'd be better to comment on _why_ we're setting these tunables to these values, since we can't determine that from the code.

grwilson: For this round it was just to be in parity with the illumos tunables so that's why I left vague. We will revisit this when we get to the performance epic.

PR #37, opened 2019-01-24
Move sysctl configuration from "/etc" into "/lib"

prakashsurya:


prakashsurya:
On an existing `dlpx-trunk` based VM, I moved the `/etc/sysctl.d/50-override.conf` file to `/lib/sysctl.d/50-override.conf` and verified the file is still consumed by `sysctl`; e.g.
```
$ sudo sysctl --system | grep Applying | grep 50-override.conf
* Applying /lib/sysctl.d/50-override.conf ...
```

I'll wait for #36 to land, pull that into this change, and then land this.

PR #38, opened 2019-01-24
LX-1628 Assign specific UIDs and GIDs for Delphix users

pzakha:
This is required for the migration project.

A follow-up change would be to chgrp all the files in `/mds` to `postgres`. We just need to determine when is the right time post-migration to do that.

## TESTING
- http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/391/
- Perform migration. In Linux: stop postgres service (failed), import domain0, start postgres service (success)
- Check that domain0 files created in Illumos are actually owned by `delphix` in Linux.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -20,6 +20,27 @@ configure)
 	systemctl enable auditd.service
 	systemctl enable delphix-platform.service
 	systemctl enable systemd-networkd.service
+
+	if ! id -u postgres >/dev/null; then
+		# When installing postgres, a postgres user is created unless it
+		# already exists. To have a consistent UID accross installations
+		# and to facilitate migration from Illumos we create a postgres
+		# user with the same UID as in Illumos. The Ubuntu postgres
+		# package requires that the postgres user is also part of the
+		# postgres group. Again, for consistency, we create the postgres
+		# group with a pre-determined GID, which the same as the
+		# postgres UID. Note that we put this code here instead of the
+		# delphix-platform service as we need it to be executed before
+		# the postgres package gets installed.
+		addgroup postgres --gid 65437
+		adduser --home /var/lib/postgresql --no-create-home \
+			--shell /bin/bash --ingroup postgres \
+			--gecos "PostgreSQL administrator" --uid 65437 \
+			--disabled-password postgres
+	elif [[ $(id -u postgres) -ne 65437 ]]; then
+		echo "ERROR: postgres id $(id -u postgres) != 65437" >&2
+		exit 1
+	fi
 	;;
 esac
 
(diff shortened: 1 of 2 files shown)


debian/postinst
```diff
@@ -22,6 +22,31 @@ case $1 in
...
+		    --disabled-password postgres
+		# To facilitate migration, we also add postgres to group 10,
+		# which is the GID of group staff on Illumos. domain0/mds is
+		# owned by delphix:staff on Illumos.
+		usermod -a -G 10 postgres
+		# Finally, we also add postgres to the Linux staff group, so
+		# that it can access files created by the delphix user.
+		usermod -a -G staff postgres
+	elif [[ $(id -u postgres) -ne 65437 ]]; then
+		echo "ERROR: postgres id $(id -u postgres) != 65437" >&2
```
prakashsurya: this message says "ERROR", but we don't actually return an error.. so package installation/upgrade will report "success".. is this intentional?

debian/postinst
```diff
@@ -22,6 +22,31 @@ case $1 in
 	systemctl enable systemd-networkd.service
 
 	systemctl disable walinuxagent.service
+
+	if ! id -u postgres > /dev/null; then
```
prakashsurya: can you also comment on why we do this configuration here, rather then some place else (e.g. the delphix-platform service)? I we have to do it here for a good reason, so it'd be good to explain that reason, so it's not accidentally moved without realizing the implications.

debian/postinst
```diff
@@ -22,6 +22,31 @@ case $1 in
...
+		# postgres UID.
+		addgroup postgres --gid 65437
+		adduser --home /var/lib/postgresql --no-create-home \
+		    --shell /bin/bash --ingroup postgres \
+		    --gecos "PostgreSQL administrator" --uid 65437 \
+		    --disabled-password postgres
+		# To facilitate migration, we also add postgres to group 10,
+		# which is the GID of group staff on Illumos. domain0/mds is
+		# owned by delphix:staff on Illumos.
+		usermod -a -G 10 postgres
```
prakashsurya: is the plan to eventually remove this?

it's not clear to me if we want all files to use GID 10, or GID "linux staff" after a migration.

pzakha: Yes, the plan would be to remove this some time after the migration has been completed and files/datasets have been chowned appropriately. I agree that there is no value in keeping this around forever.

debian/postinst
pzakha: Good catch, I'll fix this.

debian/postinst
pzakha: Okay, I think this logic belongs more into the upcoming `delphix-migration` service. This way we won't have postgres beloging to group 10 or 50 for fresh Linux installs.

prakashsurya:
it might be good to rebase and squash this to a single commit before landing.

PR #41, opened 2019-01-28
Don't run zpool command during appliance-build

prakashsurya:
We need to be careful with our delphix-platform service's Ansible code,
and ensure we don't execute commands that would need the ZFS modules
during appliance-build. We accomplish this by causing these tasks not to
run when it detects Ansible is being run in a "chroot" environment.
Unfortunately, a prior change did not properly guard against a "zpool"
invocation, which is causing builds to fail; this change adds the guard,
which should fix the build failures.


prakashsurya:
cc: @jgallag88 @pzakha

prakashsurya:
bor r+

Issue #44, opened 2019-01-29
"apply" script does not handle errors correctly


basil:
The `apply_playbook` function contains logic that attempts to fail the script with `exit 1` when `ansible-playbook` fails:

```bash
function apply_playbook() {
	if ! ansible-playbook -vvv \
		-c "$DLPX_ANSIBLE_CONNECTION" \
		-i "$DLPX_ANSIBLE_INVENTORY" \
		-e "root_filesystem=${ROOT_FILESYSTEM}" \
		"$1"; then
		echo "Failure when applying playbook." 2>&1
		exit 1
	fi
}
```

However, `apply_playbook` is invoked as follows:

```bash
find_playbooks | while read -r playbook; do
        apply_playbook "$playbook"
done
```

This spawns a sub-shell for the body of the loop, meaning that `exit 1` exits the subshell with a non-zero exit status rather than the higher-level shell (i.e., the one running `apply`). As such, the higher-level shell running `apply` drives on and exits cleanly, even though an error occurred.

A suggested fix is to switch to process substitution, as follows:

```bash
while read -r playbook; do
        apply_playbook "$playbook"
done < <(find_playbooks)
```

This avoids the subshell. Note that from the perspective of the invoked process, `<(...)` resolves to a named pipe, as this demonstrates:

```bash
$ echo <(true)
/dev/fd/63
```

basil:
bq. This spawns a sub-shell for the body of the loop, meaning that exit 1 exits the subshell with a non-zero exit status rather than the higher-level shell (i.e., the one running apply).

After running a few more experiments, I don't think this is actually the case. `exit 1` should still work in the presence of a subshell, as this example demonstrates:

```bash
printf "1\n2\n3\n" | while read foo; do
	echo "$foo"
	exit 1
done
```

As such I think my original analysis is wrong.

prakashsurya:
Is this a duplicate of https://github.com/delphix/appliance-build/issues/215? @jgallag88 is proposing a fix here: https://github.com/delphix/delphix-platform/pull/45.

basil:
Yeah, I think it is a duplicate.

prakashsurya:
Closing, as I believe this is fixed.

PR #45, opened 2019-01-29
Propagate Ansible errors in 'apply' script

jgallag88:
The 'apply_playbook' function runs in a sub-shell because it is called
in a pipeline, so the 'exit 1' in that function only exits from the
sub-shell. We need to check the exit status of the sub-shell from the
main shell in order to propagate errors.

Fixes https://github.com/delphix/appliance-build/issues/215

### DIFF
--- a/var/lib/delphix-platform/ansible/apply
+++ b/var/lib/delphix-platform/ansible/apply
@@ -27,7 +27,7 @@ function find_playbooks() {
 		-mindepth 2 \
 		-name playbook.yml | sort -n; then
 		echo "Failure when finding playbooks." 2>&1
-		exit 1
+		return 1
 	fi
 
 	#
@@ -36,13 +36,12 @@ function find_playbooks() {
 	# old directory. Once all consumers can be updated to install
 	# into the new directory, we can remove this logic.
 	#
-	if ! find "/etc/delphix-platform/ansible" \
+	find "/etc/delphix-platform/ansible" \
 		-maxdepth 2 \
 		-mindepth 2 \
-		-name playbook.yml | sort -n; then
-		echo "Failure when finding playbooks." 2>&1
-		exit 1
-	fi
+		-name playbook.yml | sort -n
+
+	return 0
 }
 
 function apply_playbook() {
@@ -106,7 +105,8 @@ if [[ "$(find_playbooks | wc -l)" -eq "0" ]]; then
 	exit 1
 fi
 
-find_playbooks | while read -r playbook; do
+playbooks=$(find_playbooks) || exit 1
+for playbook in $playbooks; do
 	apply_playbook "$playbook"
 done
 


jgallag88:
Tested by doing an appliance-build with this change and with https://github.com/delphix/delphix-platform/pull/41 reverted. The build failed as expected:
```
14:46:47     }, 
14:46:47     "msg": "non-zero return code", 
14:46:47     "rc": 1, 
14:46:47     "start": "2019-01-28 22:46:47.405942", 
14:46:47     "stderr": "The ZFS modules are not loaded.\nTry running '/sbin/modprobe zfs' as root to load them.", 
14:46:47     "stderr_lines": [
14:46:47         "The ZFS modules are not loaded.", 
14:46:47         "Try running '/sbin/modprobe zfs' as root to load them."
14:46:47     ], 
14:46:47     "stdout": "", 
14:46:47     "stdout_lines": []
14:46:47 }
14:46:47 	to retry, use: --limit @/opt/appliance-build/live-build/build/internal-dev-aws/binary/var/lib/****-platform/ansible/10-****-platform/playbook.retry
14:46:47 
14:46:47 PLAY RECAP *********************************************************************
14:46:47 binary                     : ok=4    changed=3    unreachable=0    failed=1   
14:46:47 
14:46:47 Failure when applying playbook.
14:46:47 E: config/hooks/80-platform-configuration.binary failed (exit non-zero). You should check for errors.
14:46:47 P: Begin unmounting filesystems...
14:46:47 P: Saving caches...
14:46:48 Reading package lists...
14:46:48 Building dependency tree...
14:46:48 Reading state information...
14:46:48 + kill -9 1022
14:46:48 + [[ ! -f binary/SHA256SUMS ]]
14:46:48 + exit 1
14:46:48 
14:46:48 > Task :live-build:buildInternalDevAws FAILED
14:46:48 
14:46:48 FAILURE: Build failed with an exception.
```

linux-pkg build job: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/pre-push/34/console
appliance-build-orchestrator job: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/436/console

var/lib/delphix-platform/ansible/apply
```diff
@@ -36,13 +36,12 @@ function find_playbooks() {
...
+	find "/etc/delphix-platform/ansible" \
 		-maxdepth 2 \
 		-mindepth 2 \
-		-name playbook.yml | sort -n; then
-		echo "Failure when finding playbooks." 2>&1
-		exit 1
-	fi
+		-name playbook.yml | sort -n
+
+	return 0
```
prakashsurya: is this `return` call necessary?

var/lib/delphix-platform/ansible/apply
```diff
@@ -36,13 +36,12 @@ function find_playbooks() {
 	# old directory. Once all consumers can be updated to install
 	# into the new directory, we can remove this logic.
 	#
-	if ! find "/etc/delphix-platform/ansible" \
+	find "/etc/delphix-platform/ansible" \
```
prakashsurya: why don't we change the `find` call a few lines up?

var/lib/delphix-platform/ansible/apply
```diff
@@ -106,7 +105,8 @@ if [[ "$(find_playbooks | wc -l)" -eq "0" ]]; then
 	exit 1
 fi
 
-find_playbooks | while read -r playbook; do
+playbooks=$(find_playbooks) || exit 1
+for playbook in $playbooks; do
```
prakashsurya: @basil's proposed solution looks a little cleaner to me:
```
while read -r playbook; do
        apply_playbook "$playbook"
done < <(find_playbooks)
```
since it avoids creating a new variable.

But I'm OK with this too; either way, it's not a big deal, as long as the script properly reports errors, now.

prakashsurya:
@basil do you also what to give this a look, since you've looked into this failure mode as well?

var/lib/delphix-platform/ansible/apply
jgallag88: Since we know that delphix-platform ships a playbook, the first invocation `find` should always find something. It seems worth asserting that this is the case.

Once we move the playbook provided by virtualization, though,  `/etc/delphix-platform/ansible` won't exist, so the second invocation of `find` will fail. It will also fail when we do the appliance-build because virtualization hasn't been installed yet when we run this script. That should be fine though, so we shouldn't check for errors in the second call.

basil:
I'm a bit confused now, because based on the empirical example in my second comment in #44 I would expect `exit 1` to work from within a subshell (where by "work" I mean "exit the parent shell"). So my understanding must be incomplete. I agree with you that the use of `exit` and `return` from the same function seems confusing to me. If anything, it raises the question: "which one should I use when modifying this code, and why?"

var/lib/delphix-platform/ansible/apply
jgallag88: Process substitution does look cleaner, but we still have `find_playbooks` running in a sub-shell, so we have to check that sub-shell's exit status. I just wasn't aware of a simple way to do that.

var/lib/delphix-platform/ansible/apply
jgallag88: No, it isn't really necessary. I just wanted to point out that we were deliberately ignoring the exit status of `find`. Maybe this just makes it more confusing though. I can remove it.

basil:
> This behavior seems consistent with the idea that the `exit` only exits from the sub-shell. The shell running `while read foo; do echo "$foo"; exit 1; done` exits, but the parent shell (my interactive one) doesn't.

Yeah, that does seem consistent. But I can't explain the behavior when it's running from a separate script:

```
$ cat test.bash
#!/bin/bash
printf "1\n2\n3\n" | while read foo; do echo "$foo"; exit 1; done
$ ./test.bash
1
```

In this case I would expect the first shell spawned by my interactive shell to spawn a subshell (which would exit 1) but then ignore its result. Yet the shell spawned by my interactive shell seems to observe its subshell's result and exit 1. I can't explain this based on my current understanding.

Regardless of the above, your change still looks good to me, since you have proven it works empirically.

basil:
By the way, I suspect the code in this example wouldn't work if there are any paths with spaces that are returned by the `find` command. I'm not sure that is important enough to warrant changing this code any further, though.

jgallag88:
> Yet the shell spawned by my interactive shell seems to observe its subshell's result and exit 1.

I think it's because, by default, the exit status of the last command in a pipeline is the exit status of the overall pipeline, and the exit status of the last command in the script is the exit status of the script.

prakashsurya:
Another potential solution, could be to use `kill 0` instead of `exit`, when trying to abort on failure..

> If pid equals 0, then sig is sent to every process in the process group of the calling process.

```
$ cat test.sh
#!/bin/bash -e
printf "1\n2\n3\n" | while read n; do
        echo $n
        kill 0
done
echo done

$ bash test.sh
1
Terminated
```

I'm OK with what's here, but just wanted to mention this (potential) alternate solution.

basil:
> I think it's because, by default, the exit status of the last command in a pipeline is the exit status of the overall pipeline

Sure, but then wouldn't the same logic apply to the original code that was exhibiting the bug? I still seem to be missing something. I suppose I would understand the original bug and the fix a lot more clearly if we could distill it into a minimal reproducer. But complete understanding shouldn't be a prerequisite to you pushing this change, since you've clearly demonstrated that the problem is no longer present with your new code.

prakashsurya:
@basil in your reproducer in #44:
```
printf "1\n2\n3\n" | while read foo; do
	echo "$foo"
	exit 1
done
```
the `exit` doesn't terminate the parent.. e.g. if you and another `echo` after the loop, that'll still run..
```
printf "1\n2\n3\n" | while read foo; do
	echo "$foo"
	exit 1
done
echo done
```
this reproduces the issue, right?

basil:
> this reproduces the issue, right?

Yep, it sure does. Thanks! Now I think I understand everything that's going on here. So my original understanding was correct, but my original attempt at a reproducer was flawed.

PR #46, opened 2019-01-29
End support of "/etc/delphix-platform/ansible" directory

prakashsurya:
Previously, in commit 37809a574e9e209a4a58020b7155a13aea85d13d we moved
the location of the delphix-platform service's ansible files from "/etc"
to "/var/lib". Since, at that time, consumers of the delphix-platform
service would still install files in "/etc", we couldn't completely drop
support for the "/etc/delphix-platform/ansible" directory.

Now that consumers have been updated to stop installing files into the
"/etc/delphix-platform/ansible" directory, we can safely drop support
for this directory, and support only the "/var/lib" directory.


prakashsurya:
see also: http://reviews.delphix.com/r/46533/

PR #48, opened 2019-02-14
bcc-tools product dependency

brad-lewis:
With the analytics scripts such as the nfs collector using bcc it needs to be included in the product dependencies


prakashsurya:
@brad-lewis do we still want this?

PR #49, opened 2019-02-15
Configure Azure agent

jgallag88:


jgallag88:
**Testing**

 - linux-pkg: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/pre-push/57/console
 - appliance-build-orchestrator (for azure and aws): http://collins.d.delphix.com:37977/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/70/console
 - blackbox pre-checkin tests against Azure VM with these changes: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-smoke/65883/consoleFull

jgallag88:
cc @grwilson

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -234,3 +234,31 @@
     line: '{{ item.key }}="{{ item.value }}"'
   with_items:
     - { key: 'JAVA_HOME', value: '/usr/lib/jvm/oracle-java8-jdk-amd64' }
+
+- shell: dpkg-query -Wf '${Package}' 'delphix-platform-*' | sed 's/delphix-platform-//'
+  register: determine_platform
```
prakashsurya: We have precedence for setting variables via the `apply` script; e.g. see [here](https://github.com/delphix/delphix-platform/blob/master/var/lib/delphix-platform/ansible/apply#L38)

Would it be better to set the variable in `apply` rather than doing it here? If we do it that way, then any other roles (e.g. the app-gate) could also make use of it (not sure if it'll eventually need to or not).

i.e. we could do something like this:
```
PLATFORM=$(dpkg-query -Wf '${Package}' 'delphix-platform-*' | cut -d- -f 3 )

function apply_playbook() {
	if ! ansible-playbook -vvv \
...
                -e "platform=${PLATFORM}" \
```

and then it'd be used like what you're already doing:
```
when: platform is defined and platform == "azure"
```

perhaps we could even throw an error if `platform` isn't defined, since that probably means there's a bug rather than silently ignoring the issue:
```
- fail:
    msg: 'variable "platform" is not defined'
  when: platform is not defined
```
or
```
- assert:
    that:
      - platform is defined
```

If you'd rather go with what you have already implemented, that's OK with me too.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -234,3 +234,31 @@
     line: '{{ item.key }}="{{ item.value }}"'
   with_items:
     - { key: 'JAVA_HOME', value: '/usr/lib/jvm/oracle-java8-jdk-amd64' }
+
+- shell: dpkg-query -Wf '${Package}' 'delphix-platform-*' | sed 's/delphix-platform-//'
```
prakashsurya: Should we create a script/command that encapsulated this code? Since it's used in the upgrade scripts as well, I'm just wondering if wrapping it would make things a little more maintainable.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -234,3 +234,31 @@
...
+    #
+    - { key: 'Extensions.Enabled', value: 'n' }
+    #
+    # This controls auto-updating of the extension handler (not the
+    # provisioning handler or daemon). It seems safer to disable this
+    # rather than to allow Microsoft to push new code to our engines,
+    # especially since we aren't using any extensions.
+    #
+    - { key: 'AutoUpdate.Enabled', value: 'n' }
+  when: determine_platform.stdout == "azure"
```
prakashsurya: Can you also leave a comment somewhere to remind the reader that the "azure agent" package is only included for the azure version of the platform package..?

When I first saw this I wondered if we should disable the service when the platform is not azure, but after looking more closely seems that the agent won't even be installed for non-azure platforms, is that right? A quick comment might help get that point across, and save the reader some time and/or confusion.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
jgallag88: That sounds reasonable to me, I'll do that.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
jgallag88: Yeah, I think that's reasonable. Alternatively, we could install a file that has the platform as its contents, if that would be simpler for consumers. I'm not sure where such a file would go on the filesystem though. Something along the lines of
```
$ cat /etc/delphix/platform
azure
```

prakashsurya: Yea, I think that'd be much bette.

I'm not sure how we'd implement that, though, give the current build system. How would we provide that file? I don't think we have a way to customize files provided by the debian package(s) based on the platform, do we?

grwilson: I think most of the `/var/opt/delphix/server.conf` is no longer needed but we could also append the platform information to that file.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -234,3 +234,31 @@
...
+- lineinfile:
+    dest: /etc/waagent.conf
+    regexp: '^{{ item.key }}='
+    line: '{{ item.key }}={{ item.value }}'
+  with_items:
+    #
+    # We use cloud-init rather than the Azure agent to handle any
+    # provisioning logic that we need.
+    #
+    - { key: 'Provisioning.Enabled', value: 'n' }
```
grwilson: Do we need to do anything to prevent users from being added to the delphix engine?

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
prakashsurya: Just FYI, we're adding another occurance of `dpkg-query` + `sed` to get the platform in [this review](http://reviews.delphix.com/r/46631); so this would bring the duplication to delphix-platform, upgrade scripts, and now the app-gate.

prakashsurya: In that review, it's also using `dpkg-query -Wf '${Version}' "delphix-entire-$PLATFORM"` to get the "appliance version". If we end up writing a script to query the "appliance platform", it might be worth also adding a script to query the "appliance version" and perhaps the "appliance variant". At that point though, it might be better left for a follow up change since there's some feature creep now.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
jgallag88: Creating a new user is done as part of provisioning, so I think disabling provisioning should be sufficient. I am planning on asking for access to the Azure console so that I can confirm this though.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
jgallag88: I've added a file `/etc/delphix/info/platform`. Let me know what you think

grwilson: FYI, my gcp changes will build off of this too.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -234,3 +234,31 @@
     line: '{{ item.key }}="{{ item.value }}"'
   with_items:
     - { key: 'JAVA_HOME', value: '/usr/lib/jvm/oracle-java8-jdk-amd64' }
+
+#
+# Configure the Azure agent. Only run this on Azure, since that is the
+# only platform that has the Azure agent installed.
+#
+- lineinfile:
+    dest: /etc/waagent.conf
```
grwilson: This is more a question on our ansible style -- should we be using `path` or `dest` here?

prakashsurya: Looking [here](https://docs.ansible.com/ansible/latest/modules/lineinfile_module.html), it does seem like `dest` has been replaced with `path`, but `dest`, `destfile`, and `name` all still work. I'd opt for moving to `path` since that seems like the approach the docs imply is best, and it might save us some work later if/when the aliases are removed.

Makefile
```diff
@@ -36,9 +36,8 @@ package-%:
 	@rm -f debian/changelog
 	dch --create --package delphix-platform -v $(VERSION) \
 			"Automatically generated changelog entry."
-	# Add a file which can be queried to find out the platform on a
-	# running engine.
-	echo "$*" >platform
+	sed "s/@@PLATFORM@@/$*/" var/lib/delphix-appliance/platform.in \
```
grwilson: Any reason not to use TARGET_PLATFORM here too?

debian/install
```diff
@@ -1,4 +1,6 @@
 etc/
 lib/
 usr/
-var/
+var/lib/delphix-appliance/platform
+var/lib/delphix-platform/
+var/target
```
grwilson: Where is var/target used?

prakashsurya: see here: https://github.com/delphix/delphix-platform/commit/0fc62878f835e2f04b1d4659a1055235ea56b84e

Makefile
prakashsurya: +1 .. I also noticed the inconsistency, but didn't bother to mention it.

var/lib/delphix-platform/ansible/apply
```diff
@@ -20,6 +20,7 @@ set -o pipefail
 SOURCE_DIRECTORY="$(readlink -f "${BASH_SOURCE%/*}")"
 DLPX_ANSIBLE_DIRECTORY="${DLPX_ANSIBLE_DIRECTORY:-$SOURCE_DIRECTORY}"
 ROOT_FILESYSTEM=$(zfs list -Ho name /)
+PLATFORM=$(cat /etc/delphix/info/platform)
```
grwilson: This needs to be updated to the new file location

jgallag88:
Added `Provisioning.UseCloudInit=n` to the values that we are setting to prevent timeouts when provisioning. I also did manual testing to verify that the user specified in the Azure portal when provisioning is not created, and that extensions specified in the Azure portal, such as 'Custom Script For Linux', were not run.

ab-pre-push: http://collins.d.delphix.com:37977/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/80/

var/lib/delphix-platform/ansible/apply
```diff
@@ -21,6 +21,13 @@ SOURCE_DIRECTORY="$(readlink -f "${BASH_SOURCE%/*}")"
 DLPX_ANSIBLE_DIRECTORY="${DLPX_ANSIBLE_DIRECTORY:-$SOURCE_DIRECTORY}"
 ROOT_FILESYSTEM=$(zfs list -Ho name /)
 
+#
+# If we are running as part of live-build, the root of the appliance's
+# filesystem won't be /
+#
+APPLIANCE_ROOT_DIR="$SOURCE_DIRECTORY/../../../.."
```
grwilson: I was talking with Prakash about this yesterday and I wasn't sure if we needed to have a more global concept of the ROOT for the appliance build. Also this probably needs to be inside an if-block and only execute when we $DLPX_ANSIBLE_CONNECTION == chroot.

grwilson: Ok, when we do upgrade, I wasn't sure if it's called from the same level in the directory tree. If you've verified this then we wouldn't need to guard this with `chroot` check.

prakashsurya: Would it be better to expose this as an environment variable? Then appliance-build can pass in that information via the variable, and if it's not provided it can default to `/` (or perhaps the service can also provide it, and this script would error if it's empty)?

This just strikes me as hack, and easily broken if the location of this script changes.

Thoughts? I'm OK with this if folks would rather do this, but just wanted to bring up another approach.

This also doesn't solve the issue for the `zfs list /` call above, but I'm not sure we can fix that will the same solution (although we should probably use the same "appliance root" in the `zfs list`, even if it still doesn't work).

jgallag88: While I don't love the change that I proposed, I also don't love adding another environment variable that has to be set, since the caller of this script lives in another repository. I think it would be good to keep the interface between the two repositories as simple where possible to prevent flag days.

I think ideally we would enter the chroot before calling the `apply` script, so that it doesn't need to be aware of the difference scenarios in which it can be called. However, I know that Ansible does need to be aware when it is running in a chroot, so it would need to be able to detect that itself. I'm not sure whether that's possible.

prakashsurya: Yea... the fact that this script is called outside of the chroot is to hack around the fact that ansible can't be called directly from the chroot.

OK, I'm good with this solution. I can't think of anything better at the moment.

grwilson: I agree. It's hacky but can live with this until we have a broader need to make it more robust.

PR #51, opened 2019-02-20
LX-1770 Add our custom td-agent to delphix-platform

prashks:
This adds our custom td-agent package built via linux-pkg framework to `delphix-platform`
and depends on LX-1645 to build the custom td-agent package.

* appliance-build-orchestrator run that uses this and corresponding linux-pkg changes   : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/506/console
* Created a VM based on the resulting image and the custom `td-agent` is installed by default as expected

```
psreenivasa@dlpxdc$ dc clone-latest dlpx-psreenivasa-master pks-lx-1645-new
pks-lx-1645-new - dlpx-psreenivasa-master
psreenivasa@dlpxdc$

delphix@ip-10-110-215-46:~$ uname -a
Linux ip-10-110-215-46 4.15.0-1032-aws #34-Ubuntu SMP Thu Jan 17 15:18:09 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
delphix@ip-10-110-215-46:~$
delphix@ip-10-110-215-46:~$
delphix@ip-10-110-215-46:~$ sudo apt show td-agent
Package: td-agent
Version: 3.3.0-delphix-2019.02.19.22
Status: install ok installed
Priority: extra
Section: misc
<<<snip>>>
```

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -61,11 +61,16 @@ DEPENDS += ansible, \
 DEPENDS += ubuntu-dbgsym-keyring,
 
 #
-# This package contains a consolidation for third-party packages built
-# by delphix as well as build info for all packages built by the linux-pkg
-# framework.
+# This package provides the build information for all packages built by
+# the linux-pkg framework.
 #
 DEPENDS += delphix-extra,
+
+#
+# The following packages provide additional set of packages that are
+# common part of the Delphix platform.
+#
+DEPENDS += td-agent,
 
 # Platform-specific dependencies
 DEPENDS.aws =


debian/rules
```diff
@@ -61,12 +61,17 @@ DEPENDS += ansible, \
...
+# This package provides the build information for all packages built by
+# the linux-pkg framework.
 #
 DEPENDS += delphix-extra,
 
+#
+# The following packages provide additional set of packages that are
+# common part of the Delphix platform.
+#
+DEPENDS += td-agent,
```
prakashsurya: why not put this in the list of packages above? the “minimal” set of packages..

related, why have it be a dependency of the platform package rather than the virtualization package?

prakashsurya:
Apart from my two questions, LGTM. 

bors delegate+

debian/rules
prashks: @prakashsurya I didn't put it in the "minimal" set because per its comments, its for packages "necessary to create a working system" and this td-agent while wanted isn't essential for a working system

Regarding second comment, while this package is used by the virtualization/appstack, it is not limited to that - other components of delphix platform like masking, orbital integration etc will eventually use it.
And logically this (td-agent/fluentd) forms an essential piece of the delphix platform for any subsystem to push out Insight/diagnostic data out of DDP.

prashks:
@sebroy Like I mentioned in the JIRA description, while this package is used by the virtualization/appstack, it is not limited to that - other components of delphix platform like masking, orbital integration etc will eventually use it.
And logically this (td-agent/fluentd) forms an essential piece of the delphix platform for any subsystem to push out Insight/diagnostic data out of DDP.
Hence this belongs to delphix-platform package.

debian/rules
prakashsurya: Sorry, now it makes sense. I thought the comment said something about the list creating the "minimal platform"; I didn't actually re-read it last night.

I'm OK creating a new "section" like you did here. I'd vote for pulling `delphix-extra` into this new section as well (or not, up to you).

> other components of delphix platform like masking, orbital integration etc will eventually use it.

These other components could also depend on the td-agent.. e.g. masking and virtualization packages could both depend on td-agent; but that's a little redundant since masking already depends on td-agent. Same for any orbital component; presumably that'll have a package that could depend on td-agent in the future as well.

I'm not opposed to adding the dependency here, it's just the argument for doing it here vs. the actual consumers of the package feels is a little weak (to me, at least).

prashks:
On further discussion, going with the rationale of moving the dependency to the consumers of the package.
Hence closing this out for now.

PR #54, opened 2019-02-22
Fix cloud-init configuration priority

jgallag88:
Currently our setting for datasource_list is being overridden by a
configuration file generated by the cloud-init package. This results in
cloud-init executing whatever userdata it is passed, even on the
external-standard variant. We can make sure our setting gets used by
renaming the configuration file so that it is read after the one
generated by cloud-init.

### DIFF
--- a/etc/cloud/cloud.cfg.d/91-delphix-platform.cfg
+++ b/etc/cloud/cloud.cfg.d/91-delphix-platform.cfg
@@ -15,6 +15,12 @@
 #
 
 #
+# The configuration in this file has to be applied after 90_dpkg.cfg,
+# which is created by the cloud-init package's postinstall script.
+# Otherwise the value of 'datasource_list' set here will be reset.
+#
+
+#
 # While we create the user using Ansible, we need to inform cloud-init
 # of this user so that SSH keys are properly inserted into this user's
 # authorized_keys file (e.g. when running in EC2). Additionally, this


jgallag88:
**Testing**

Built and deployed the `external-standard` variant, and confirmed that cloud-init was no longer reading userdata from EC2. From `/var/log/cloud-init.log`
```
__init__.py[DEBUG]: Looking for data source in: ['None']
```
Before, this read
```
__init__.py[DEBUG]: Looking for data source in: ['Ec2', 'None']
```
I also confirmed that userdata was still being read from EC2 on the internal variants.

git ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/540/

prakashsurya:
bors delegate+

prakashsurya:
Actually, 99 wouldn't work since we use that in the "internal" role to override this for internal systems. 91 is good.

prakashsurya:
Alternatively, did you consider setting the dpkg configuration that is used to populate `90_dpkg`? I think it can be configured if we set the dpkg configuration `cloud-init/datasources`:
```
   if db_get cloud-init/datasources; then
      values="$RET"
...
      cat > /etc/cloud/cloud.cfg.d/90_dpkg.cfg <<EOF
# to update this file, run dpkg-reconfigure cloud-init
datasource_list: [ $values ]
EOF
   fi
```

prakashsurya:
We could probably easily tweak the cloud-init package itself, since we already rebuild it via linux-pkg. We could either remove the code that creates `90_dpkg.cfg`, or modify the default value.

Honestly, I'm OK with this; just want to understand if any alternatives were considered, and why we opted for this approach.

jgallag88:
My feeling was that in general it's good to keep forked packages as close as possible to upstream, so if we had a choice between modifying a forked package and one of our own packages, we should lean towards doing the latter. I'm open to other way of handling this though.

prakashsurya:
That makes sense. I think we could potentially improve the situation by running something like:
```
echo 'cloud-init cloud-init/datasources select None' | debconf-set-selections
```
in the `postinst` packaging hook (or maybe a new `preinst` hook); but I guess that's not necessary with this approach. My only concern is that if the file is moved, e.g. to `92_dpkg.cfg`, things will silently break again.

Since this configuration is pretty important to our appliance working correctly, perhaps the best solution to avoid silent breakage is to verify it with blackbox?

jgallag88:
@prakashsurya I agree it would be good to test for this going forward. I've filed https://jira.delphix.com/browse/LX-1777

PR #58, opened 2019-02-26
Create a delphix group

sebroy:
This commit adds a "delphix" group to the system so that the masking
user can get permissions on root-owned files without having to relax the
ownership of those files for other users. The "delphix" user will also
be added to this group for convenience.

### Testing

I added the ansible role changes to a VM and re-applied the role. The result:
```
delphix@seb-linux:~$ groups
staff root delphix
delphix@seb-linux:~$ grep ^delphix /etc/group
delphix:x:1000:delphix
```

### DIFF
--- a/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -26,11 +26,17 @@
     state: directory
     mode: 0755
 
+- group:
+    name: delphix
+    state: present
+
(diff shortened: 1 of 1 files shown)


jgallag88:
@sebroy Just out of curiousity, can you comment on how this new group will be used? Is its purpose different than that of the `staff` group?

sebroy:
@jgallag88 the virtualization and masking Java processes both need access to the same Java trust stores. I hadn't thought of using the `staff` group for that, I suppose we could do that instead of creating a new group...

sebroy:
We've decided to use the existing `staff` group for this, so I've closed this PR.

PR #60, opened 2019-03-05
migration: rpool/crashdump must be mounted early on

pzakha:
Description is in the code'd comments.

## TESTING
* Ran migration and made sure delphix-platform service didn't fail because `/var/crash` wasn't empty.
* ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/634/

### DIFF
--- a/var/lib/delphix-platform/os-migration
+++ b/var/lib/delphix-platform/os-migration
@@ -28,6 +28,19 @@ function die() {
 }
 
 function perform_migration() {
+	#
+	# On Illumos rpool/crashdump has its mountpoint set to "legacy",
+	# and we mount it manually. On Linux we want it to be mounted
+	# automatically by setting its mountpoint property.
+	# During migration this would be done by the delphix-platform service
+	# on first boot however that fails since the kdump-tools service
+	# creates a file under /var/crash before the delphix-platform can be
+	# run. To mitigate this problem we set the mountpoint here, as we
+	# run before kdump-tools.
+	#
+	zfs set mountpoint=/var/crash rpool/crashdump ||
+		die "Failed to set mountpoint for rpool/crashdump"
+
 	zpool import -f domain0 || die "Failed to import domain0"
 	# domain0/mds is owned by delphix:staff on Illumos, and on Linux
 	# postgres can't access this dataset anymore. We change the permissions


pzakha:
cc: @sebroy, @prakashsurya

pzakha:
> Why wouldn't we create/configure this dataset in the migration scripts.

This is a migration script.

> i.e. have the datasets created via the migration scripts, mimic what we do in 90-raw-disk-image.binary ?

So in this particular case we already have a `rpool/crashdump` dataset in Illumos, so we are not creating it from scratch. 

>  Similarly, are we creating the rpool/grub dataset? If not, where should that be created?

That's a good point, we are not currently creating `rpool/grub`. The work for creating `rpool/grub` was done after Serapheim's work in dx_apply. I'll open a new bug for this.

pzakha:
I've created LX-1792 for the `rpool/grub` work

var/lib/delphix-platform/os-migration
```diff
@@ -28,6 +28,19 @@ function die() {
 }
 
 function perform_migration() {
+	#
+	# On Illumos rpool/crashdump has its mountpoint set to "legacy",
+	# and we mount it manually. On Linux we want it to be mounted
```
sebroy: Do you know why we were doing that? My hunch is that the service that runs the savecore code was running before zfs filesystems got mounted, so we needed to mount this manually first (?).

prakashsurya:
> This is a migration script.

Please define "migration script"; to me, it means one the scripts contained in the "migration-scripts" directory of appliance-build, which this is not. I'd define this as the "migration service", which is different.

> So in this particular case we already have a rpool/crashdump dataset in Illumos, so we are not creating it from scratch.

Sure, but this still doesn't answer my question as to why we do this here, rather than in one of the migration scripts.

var/lib/delphix-platform/os-migration
```diff
@@ -28,6 +28,19 @@ function die() {
...
+	# automatically by setting its mountpoint property.
+	# During migration this would be done by the delphix-platform service
+	# on first boot however that fails since the kdump-tools service
+	# creates a file under /var/crash before the delphix-platform can be
+	# run. To mitigate this problem we set the mountpoint here, as we
+	# run before kdump-tools.
+	#
+	zfs set mountpoint=/var/crash rpool/crashdump ||
```
sebroy: With this scheme, when will `/var/crash` be mounted in the boot process in relation to when the root filesystem itself gets mounted? I'm wondering if we have risk of losing crash dumps if the system crashes before it gets mounted... Would an alternative be to not have a separate dataset for `/var/crash` at all, and make it be part of the root filesystem (only trying to understand the constraints)?

prakashsurya: IIRC, the reason we have a separate dataset is so it can have a unique quota. So, if we don't use a dataset, we'll need to determine if we need a quota, and if so how to do that without ZFS.

var/lib/delphix-platform/os-migration
pzakha: We are doing this in `dxinit.sh::do_mount()`. We probably have this set to legacy because `/var` is set to legacy as well and is tied to our Illumos upgrade logic. (i.e. on first boot after upgrade we do a clone of the previous /var dataset in do_datasets() and then mount it in do_mount())

var/lib/delphix-platform/os-migration
pzakha: The reason why we have a separate dataset in `/var/crash` is because we want to set a quota there. I don't know well enough how the crash kernel works to figure out what would happen if we panic before this dataset is mounted, I think this would need more investigations.

pzakha:
@prakashsurya We have several places where there are migration scripts:
* In appliance-build, in the `migration-scripts` directory.
* In delphix-platform, called from the `delphix-migration` service
* In app-gate, in `/opt/delphix/migration`

I didn't know you specifically meant the appliance-build dx_ scripts.

We could maybe do this as part of `dx_execute`, however I do not know if it would be better. I.e. if we fail to reboot into Linux, then that would potentially mess up with the logic in `dxinit` on reboot.

prakashsurya:
> I didn't know you specifically meant the appliance-build dx_ scripts.

Cool, now I know to use "dx scripts" :) I guess my use of "migration scripts" is already outdated, since that comes from when we only had the scripts in appliance-build.

> if we fail to reboot into Linux, then that would potentially mess up with the logic in dxinit on reboot.

:+1: perfect.. so we do it here, to account for the possibility that we'll have to revert back to illumos. With that said, if we fail during a step after this runs we're in a similar situation (rollback to illumos will have to deal with this).

I'm not trying to be difficult, just trying to better understand what the conceptual responsibility is for each of the parts involved with migration.. since, at first glance, the "dx scripts" appear to handle the responsibility of doing what "90-raw-disk-image.binary" does for non-migrated systems, so I was curious why we'd break that trend for this dataset.

prakashsurya:
bors delegate+

Looks reasonable to me, pending Seb's comments.

var/lib/delphix-platform/os-migration
sebroy: I don't know enough about this to raise an issue with this change, this was just a thought that came to mind. It makes me wonder if, for example, we panic in early boot in ZFS, how would we diagnose this?

PR #61, opened 2019-03-09
Remove auditd output from journal

jgallag88:


jgallag88:
Testing:

Made sure that auditd output was still being written to `/var/log/audit/audit.log`, and that it was not being written to the systemd journal.

git ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/660/console

jgallag88:
cc @don-brady

PR #62, opened 2019-03-10
LX-1595 customize gcp guest environment

grwilson:
This adds the Linux Guest Environment so GCP variant of the delphix platform and configures the override file to ensure that only the appropriate functionality is enabled. We disable the account daemon to prevent users from being added and disable the metadata scripts to prevent running random scripts on the delphix engine.

### DIFF
--- a/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -302,3 +302,39 @@
     #
     - { key: 'AutoUpdate.Enabled', value: 'n' }
   when: platform == "azure"
+
+#
+# Customize the GCP linux environment.
+#
+# Update the override file for the GCP instance. This file gets
+# applied dynamically by running google_instance_setup script.
+#
+- blockinfile:
+    path: /etc/default/instance_configs.cfg.template
+    create: yes
+    block: |
+      #
+      # Disable the accounts daemon to prevent adding/removing
+      # users on the engine.
+      #
+      [Daemons]
+      accounts_daemon = false
+
+      #
+      # Disable user supplied startup/shutdown scripts from running on
+      # the engine.
+      #
+      [MetadataScripts]
+      shutdown = false
+      startup = false
+  when:
+    - platform == "gcp"
+  notify: "gcp config changed"
+
+#
+# Make sure that the account daemon is always disabled. The override file
+# above should prevent this and this is designed to catch any corner cases.
+#
+- command: systemctl disable google-accounts-daemon.service
+  when:
+    - platform == "gcp"
--- a/debian/rules
+++ b/debian/rules
@@ -71,7 +71,8 @@ DEPENDS += delphix-extra,
 DEPENDS.aws =
 DEPENDS.azure = walinuxagent,
 DEPENDS.esx = open-vm-tools,
-DEPENDS.gcp =
+DEPENDS.gcp = gce-compute-image-packages, python-google-compute-engine, \
+	   python3-google-compute-engine,
 DEPENDS.kvm =
 DEPENDS += $(DEPENDS.$(TARGET_PLATFORM))
 
--- a/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/handlers/main.yaml
+++ b/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/handlers/main.yaml
@@ -1,0 +1,25 @@
+#
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+#
+# Apply any changes made to the override instance config file. This
+# can only be done when we're on a running gcp instance.
+#
+- command: /usr/bin/google_instance_setup
+  listen: "gcp config changed"
+  when: not ansible_is_chroot


var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -294,3 +294,41 @@
...
+  notify: "gcp config changed"
+
+#
+# We don't want to run the account daemon when we first boot up so
+# disable here.
+#
+- command: systemctl disable google-accounts-daemon.service
+  when:
+    - platform == "gcp"
+    - ansible_is_chroot
```
sebroy: is this supposed to be `not ansible_is_chroot`?

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -294,3 +294,41 @@
...
+      #
+      # Disable user supplied startup/shutdown scripts from running on
+      # the engine.
+      #
+      [MetadataScripts]
+      shutdown = false
+      startup = false
+  when:
+    - platform == "gcp"
+    - not ansible_is_chroot
```
prakashsurya: why is this needed? we should be able to customize the file in the chroot just fine, right? and we still won't run the gcp command, since the handler has the "when" clause, right?

I'd rather use "ansible_is_chroot" as little as possible, and I don't think we need it here.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
prakashsurya: you're doing this, so we only run this command during the build, and not during the first boot? even though this is possible, I'm not sure I think we should be relying on "ansible_is_chroot" like this..

what are the requirements here? I wonder if we should disable this via the post-install package hook instead? why don't we want to run this disable anytime the platform service runs?

I just worry that if we start to do this, we'll end up with lots of conditional logic (e.g. "do this during the build, but not during runtime", "do this during runtime, but not during the build", etc.), making things more complicated for little benefit (or when a better approach exists). The original idea is we wouldn't have any build/runtime conditionas; the only reason I think using "ansible_is_chroot" makes sense, is when logic we'd run won't work during the build (e.g. creating zfs datasets).

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
grwilson: Yes, we can remove it here and just prevent the handler from running.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
grwilson: We could do this all the time so I can remove the `ansible_is_chroot`. Is the post-install hook a preferred way? I'm not that familiar with that but am willing to change it.

grwilson: The intent was to only run this at when appliance-build ran and I was trying to be more explicit about when it ran. As Prakash points out, we can just remove this and let it run all the time.

prakashsurya: if we're OK always disabling the service any time the delphix-platform service runs, then keeping it here and removing the "ansible_is_chroot" flag is best, I think.

Ultimately, I think it comes down to deciding if we want to:

1. Disable the service once and only once. This way, if the service is later enabled, and the platform service re-run, the service stays enabled.

2. Or, ensure the service is always disabled. So, if the service is later enabled, the platform service will "fix" that by disabling the service the next time it runs.

If we remove the "ansible_is_chroot" part, I feel like we want (2), so I think what you have here is best.

prakashsurya: I view the ansible code as a declaration of how we want the system to configured, rather than a way of executing commands.. and if we could use the "service" module, this difference would be more clear.

So, in this case, we're stating we always want this service to be disabled, rather than a way of running "systemctl disable".

grwilson:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/672/flowGraphTable/

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -302,3 +302,39 @@
...
+      #
+      # Disable user supplied startup/shutdown scripts from running on
+      # the engine.
+      #
+      [MetadataScripts]
+      shutdown = false
+      startup = false
+  when:
+    - platform == "gcp"
+  notify: "gcp config changed"
```
prakashsurya: will this run on the first boot of the system in GCP?

I worry the following will occur:
1. we'll configure this file during the build
2. we boot up the first time in GCP
3. this task runs, but doesn't modify the file (since it's the same as during the build)
4. since the file isn't changed, the handler doesn't run, so we don't run `google_instance_setup`

grwilson: If the file is the same then we don't need to run `google_instance_setup`.  We only need to run that if the file changes. I verified that after install the running file is updated:

```# cat /etc/default/instance_configs.cfg# This file is automatically created at boot time by the /usr/lib/python
# 3/dist-packages/google_compute_engine/instance_setup/instance_config.p
# y script. Do not edit this file directly. If you need to add items to
# this file, create or edit /etc/default/instance_configs.cfg.template
# instead and then run /usr/bin/google_instance_setup.

[Daemons]
accounts_daemon = false
clock_skew_daemon = true
ip_forwarding_daemon = true
network_daemon = true

[MetadataScripts]
shutdown = false
startup = false
```

grwilson: The idea of the notifier was to handle the upgrade case in the event that we do change this.

prakashsurya: :+1: OK, sounds good.

PR #64, opened 2019-03-19
TOOL-7390 Remove delphix-extra package

pzakha:
Remove the delphix-extra package as it will no longer be generated after TOOL-7303 lands.
This package only contains build metadata so the product has no dependency on it.

## TESTING

ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/723/


sebroy:
bors delegate+

PR #66, opened 2019-03-26
Fixes for LX-1806 and LX-1813

pzakha:
LX-1806 migration: delphix-migration service should depend on var-delphix.mount
LX-1813 migration: /var/delphix/server should be owned by root

Summary: 
- [LX-1806](https://jira.delphix.com/browse/LX-1806): delphix-migration accesses files on `/var/delphix`, which is mounted asynchronously (it is present in /etc/fstab). Adding a dependency on the `var-delphix.mount` unit waits for /var/delphix to be mounted before starting the service.
- [LX-1813](https://jira.delphix.com/browse/LX-1813): the log rotator in delphix-platform.service failed because some logs (nginx) weren't owned by root. In general on Linux files in `/var/delphix/server` should be owned by root, so we do a recursive chown. While we are there we do the same for mds, to avoid relying on the logic in svc-postgres.sh which also does some chowning.

## TESTING
- migration testing: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/36/
   note that job failure is expected. The job is used to automate the migration
- ab-pre-commit: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/765/
- For LX-1806, checked that there are no dependency cycles between the different services involved by calling `systemctl show -p Requires,Wants,Requisite,BindsTo,PartOf,Before,After <service>` and looking for effective dependencies.

### DIFF
--- a/lib/systemd/system/delphix-migration.service
+++ b/lib/systemd/system/delphix-migration.service
@@ -20,17 +20,22 @@
 # also be importing domain0, which is normally imported by the
 # zfs-import-cache service, we add similar dependencies to it
 # (i.e. devices must be ready, so run after systemd-udev-settle.service, and
-# service is required by zfs-import.target). Finally, we must make sure to
-# remove default service dependencies, as those would create a dependency
-# cycle (this service must run before local-fs.target, but default
-# dependencies require local-fs.target). Note that the zfs-import.target
-# dependency means that this service will run before local-fs.target, and so
-# before any other Delphix service.
+# service is required by zfs-import.target). The service needs to access
+# files in /var/delphix so we add a dependency on the /var/delphix mountpoint.
+# Note that /var/delphix comes from the root pool, which is imported on boot,
+# so this dependency doesn't conflict with the zfs-import.target.
+#
+# Finally, we must make sure to remove default service dependencies, as
+# those would create a dependency cycle (this service must run before
+# local-fs.target, but default dependencies require local-fs.target). Note
+# that the zfs-import.target dependency means that this service will run
+# before local-fs.target, and so before any other Delphix service.
 #
 [Unit]
 Description=Delphix OS Migration Service
 DefaultDependencies=no
 After=systemd-udev-settle.service
+After=var-delphix.mount
 Before=network-pre.target
 Before=zfs-import.target
 
--- a/var/lib/delphix-platform/os-migration
+++ b/var/lib/delphix-platform/os-migration
@@ -41,11 +41,14 @@ function perform_migration() {
 	zfs set mountpoint=/var/crash rpool/crashdump ||
 		die "Failed to set mountpoint for rpool/crashdump"
 
+	chown -R root:root /var/delphix/server ||
+		die "Failed to chown /var/delphix/server"
+
 	zpool import -f domain0 || die "Failed to import domain0"
 	# domain0/mds is owned by delphix:staff on Illumos, and on Linux
 	# postgres can't access this dataset anymore. We change the permissions
 	# to the Linux defaults.
-	chown postgres:postgres /mds ||
+	chown -R postgres:postgres /mds ||
 		die "Failed chowning /mds"
 	echo "$(date): Running migrate_config post-upgrade" \
 		>>"$MIGRATE_CONFIG_LOG"


lib/systemd/system/delphix-migration.service
```diff
@@ -20,17 +20,20 @@
...
+# those would create a dependency cycle (this service must run before
+# local-fs.target, but default dependencies require local-fs.target). Note
+# that the zfs-import.target dependency means that this service will run
+# before local-fs.target, and so before any other Delphix service.
 #
 [Unit]
 Description=Delphix OS Migration Service
 DefaultDependencies=no
 After=systemd-udev-settle.service
+After=var-delphix.mount
```
prakashsurya: Can you clarify how we can have a dependency _after_ the `/var/delphix` mount, but _before_ the `zfs-import` target?

At first glance I would have expected the `zfs-import` target to be the target that should be used to wait for that mountpoint, since that mountpoint is a ZFS dataset. There's probably some implementation details about how these two dependencies are implemented, allowing us to do this, but it's not entirely clear without more explanation.

pzakha: I've updated the comment and split the work into 2 separate commit.

The failed ab-pre-push job is due to https://jira.delphix.com/browse/INC-482, I'll restart it once the incident is resolved.
The failure in the blackbox-chained is expected, I've added a comment.

prakashsurya: OK.. so I guess my next question is how can we guarantee that "/var/delphix" will get mounted before "network-pre"? And how is this different than waiting until after "local-fs"? It's still not clear to me _why_ and/or _how_ this works.. part of my concern is if we change how we mount "/var/delphix" (e.g. swap the use of "/etc/fstab" for something else, like doing the mount in delphix-platform), will that break the dependencies here?

pzakha: I've done a bit more research on this, so here's a summary on how those dependencies work:
- `/etc/fstab` is processed by the [systemd-fstab-generator](https://www.freedesktop.org/software/systemd/man/systemd-fstab-generator.html), which generates appropriate `.mount` units. The mount unit will have some basic dependencies added to it (s.a. `Before=local-fs.target`). Here's an example for `var-delphix.mount`:

	```
	$ systemctl cat var-delphix.mount
	# /run/systemd/generator/var-delphix.mount
	# Automatically generated by systemd-fstab-generator

	[Unit]
	SourcePath=/etc/fstab
	Documentation=man:fstab(5) man:systemd-fstab-generator(8)
	Before=local-fs.target

	[Mount]
	Where=/var/delphix
	What=rpool/ROOT/delphix.BmCm6rb/data
	```
- If there are dependencies between mount points, `Before=` and `After=` dependencies will be added to the `.mount` units accordingly. If there are `RequiresMountsFor=...` entries in units, dependencies on the appropriate mount points will also be added.

- systemd will call `mount(8)` depending on the configuration of the mount unit when it is time to execute the unit. See [systemd.mount](https://www.freedesktop.org/software/systemd/man/systemd.mount.html).

So now to answer your questions:
> how can we guarantee that "/var/delphix" will get mounted before "network-pre"

There are no dependencies between `network-pre.target` and `var-delphix.mount`. `var-delphix.mount` only depends on `local-fs-pre.target`, which is run very early on and has relationship with `network-pre.target`. (See [systemd.special](https://www.freedesktop.org/software/systemd/man/systemd.special.html) for more info on `network-pre.target` and `local-fs-pre.target`)
```suggestion
After=var-delphix.mount
```

> part of my concern is if we change how we mount "/var/delphix" (e.g. swap the use of "/etc/fstab" for something else, like doing the mount in delphix-platform), will that break the dependencies here?

If you do this kind of change, then that will break `delphix-migration`, as it is intended to be run before `fs-local.target` and `delphix-platform` is intended to be run after `fs-local.target`. In general, `fs-local.target` automatically adds dependencies on each mount of the local file system, so if we are to mount /var/delphix after `fs-local.target` has run, we'd break this model.

Let me know if you have other questions.

prakashsurya: Thanks for all the info. I think this makes sense.

Out of curiosity, why do we need migration to run before `fs-local.target`?

pzakha: I guess it doesn't actually **have** to be before `fs-local.target`. The idea was that `domain0` would be imported before `zfs-import.target` (which would be the case on a normal boot, and on which `fs-local.taget` depends), but that would not be a hard requirement provided that we make sure that postgres and other services that depend on domain0 have the proper dependencies set in place. I don't think `network-pre.target` has any explicit dependencies on `fs-local.target` .

prakashsurya: OK. Thanks for your patience and explanations. I just wanted to make sure I was understanding things.

pzakha: No problems, I think it was a good idea to do this research now as it will allow us to understand the system better and have less bad surprises in the future.

PR #67, opened 2019-04-01
update JAVA_HOME to adoptopenjdk

sravyamks:
update JAVA_HOME to adoptopenJDK

**Testing:**
Linux-pkg run using forked branch of `delphix-platform` - http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/15/

appliance-build-orchestrator-pre-push : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/792/ (still running)

Other related changes for appliance-build-orchestrator run - 

app-gate -
changes : https://gitlab.delphix.com/sravya.meda/dlpx-app-gate/commit/d382bb892ff50a936a44d824941099b0143ed037
build : http://selfservice.jenkins.delphix.com/job/dlpx-app-gate/job/master/job/build-package/job/pre-push/442/

masking -
changes : https://gitlab.delphix.com/sravya.meda/dms-core-gate/commit/eddc19da356052cecd5a14ead45f838e07fae321
build - http://selfservice.jenkins.delphix.com/job/dms-core-gate/job/master/job/build-package/job/pre-push/71/

saml-app -
changes : https://gitlab.delphix.com/sravya.meda/saml-app/commit/66aabbcee3b47171f625a8a5da9f3a9b0ab5f54e

Verify JAVA_HOME on a VM from appliance build image -
```
delphix@ip-10-110-236-58:~$ echo $JAVA_HOME
/usr/lib/jvm/adoptopenjdk-java8-jdk-amd64
```

### DIFF
--- a/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -258,7 +258,7 @@
     regexp: '^{{ item.key }}='
     line: '{{ item.key }}="{{ item.value }}"'
   with_items:
-    - { key: 'JAVA_HOME', value: '/usr/lib/jvm/oracle-java8-jdk-amd64' }
+    - { key: 'JAVA_HOME', value: '/usr/lib/jvm/adoptopenjdk-java8-jdk-amd64' }
 
 #
 # Configure the Azure agent. Only run this on Azure, since that is the


prakashsurya:
Instead of making this change, can we just remove the setting of `JAVA_HOME` from this repository? It's awkward to do this here, for a couple reasons:

1. This packages doesn't depend on Java at all. So if one was to only install "delphix-platform" (i.e. without "delphix-virutualization"), they'd get a system with `JAVA_HOME` configured, but without Java installed.

2. Since linux has the "alternatives" command/system, I think we can more simply use `/usr/bin/java` which will be a symlink to the correct path (e.g. `/usr/lib/jvm/...`) for the `java` command. Thus, consumers of a "default" java can simply use `/usr/bin/java` instead of `$JAVA_HOME/bin/java`.

So, I'd suggest we change consumers to stop relying on `JAVA_HOME` (i.e. change the app-gate) if possible.

If this isn't possible, then I'd rather us move where `JAVA_HOME` into the "delphix-virtualization" package, since that's where the dependency on Java is set, and how java gets installed. Or better yet, into the java package itself, since this is a package we currently build (via the "linux-pkg" repository).

If we don't want to do either of these, then I'm OK with this change. This makes things no worse than they are today, but if we're already making changes to this code, I think it'd be a good time to reduce the technical debt here.

Sorry for the long reply; please let me know if I should clarify on anything.

basil:
I agree with @prakashsurya's suggestion to get rid of our reliance on `JAVA_HOME`. However, I think now is not the best time to break this dependency.

We still need to support illumos, and the version of `java(1)` on the `PATH` in our existing illumos-based appliance is not guaranteed to be the version that we use to build our code. On illumos-based appliances, we have the correct version installed in `/opt/jdk`. So to eliminate our reliance on `JAVA_HOME` today, we'd need to find all callers of Java (including Virtualization and Masking launch scripts, utility launch scripts, Jenkins jobs, and the Ant/Gradle builds) and update them to use `/opt/jdk/bin/java` on illumos and `[/usr/bin/]java` otherwise. Then once we drop support for illumos, we'd _still_ need to go back through that code and remove the references to `/opt/jdk`, which wouldn't be necessary once illumos support is dropped. So we'd have to do two refactorings: once to transform `JAVA_HOME` usages to `[/usr/bin/]java` usages, and once to remove illumos-specific `/opt/jdk` references.

In contrast, if we wait until after we drop support for illumos, then we can just remove all references to `/opt/jdk/bin/java` _and_ `JAVA_HOME` in one fell swoop. This would kill two birds with one stone.

Based on this reasoning, I think we should postpone the refactoring proposed here until after we drop support for illumos.

prakashsurya:
Thanks @basil. That sounds reasonable to me.

sravyamks:
Thank you everyone for reviewing -  below are the updated review requests and builds

**Testing:**

saml-app -
changes : http://reviews.delphix.com/r/48313/

app-gate - 
changes : http://reviews.delphix.com/r/48310/
build : http://selfservice.jenkins.delphix.com/job/dlpx-app-gate/job/master/job/build-package/job/pre-push/447/

masking -
changes : http://reviews.delphix.com/r/48311/
build - http://selfservice.jenkins.delphix.com/job/dms-core-gate/job/master/job/build-package/job/pre-push/73/

linux-pkg -
changes : https://github.com/delphix/linux-pkg/pull/30
build : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/15/

appliance-build-orchestrator-pre-push : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/806/

Test java version on a VM from image created (`dlpx-sravya.meda-master`) - 
```
delphix@ip-10-110-226-192:~$ java -version
openjdk version "1.8.0_202"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_202-b08)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.202-b08, mixed mode)
delphix@ip-10-110-226-192:~$ javac -version
javac 1.8.0_202
delphix@ip-10-110-226-192:~$ echo $JAVA_HOME
/usr/lib/jvm/adoptopenjdk-java8-jdk-amd64
delphix@ip-10-110-226-192:~$ cd /usr/lib/jvm/
delphix@ip-10-110-226-192:/usr/lib/jvm$ ls
adoptopenjdk-java8-jdk-amd64
```

PR #68, opened 2019-04-15
DLPX-63604 ssh on linux needs inactivity timeout

grwilson:


var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -205,6 +205,11 @@
     - { key: "AllowStreamLocalForwarding", value: "no" }
     - { key: "AllowTcpForwarding", value: "no" }
     - { key: "X11Forwarding", value: "no" }
+    #
+    # Enable an inactivity timeout so that idle connections are closed.
```
prakashsurya: IMO, a better comment would explain why we're making these settings; i.e. why do we want idle connections to be closed? Currently, this just says what we're doing, which we probably could gather from the documentation for these tunables.

grwilson: I updated the comment.

PR #71, opened 2019-04-19
DLPX-63634 Linux upgrade verify causes MSSQL VDBs to go to INACTIVE status

prakashsurya:


prakashsurya:
cc: @pzakha

prakashsurya:
I verified that when MSSQL VDBs running on the host, if one runs a "dry run" upgrade (i.e. `upgrade -nv in-place`) that the VDBs will become `INACTIVE` and will unable to be started. Additionally, the contents of `/etc/rtslib-fb-target/saveconfig.json` on the host will be empty, and `targetcli ls` will be empty. After making this change, `upgrade -nv in-place` no longer causes these problems.

I discussed this issue with Pavel offline, and there might still be some additional things to pursue to harden this upgrade path, to try and prevent this corruption by the container (e.g. even if it maliciously tried to do so).. but until we investigate those ideas, this should at least address the immediate problem of the MSSQL VDBs going down.

PR #73, opened 2019-04-23
Turn off device initialization for Azure

shartse:
Device initialization seems to either block or slow management stack startup (despite the fact that it should be happening asynchronously), causing timeouts in blacbox tests for Azure. After some discussion, it seems that this feature does provide a significant gain for some platforms and potentially causes harm in others. At the very least, the benefits of device initialization have not been established for Azure. 

While we investigate further, I wanted to supply a mechanism to adjust the default on a platform by platform basis. An app gate change clears the default features: http://reviews.delphix.com/r/48857/diff/1/ and this change will repopulate `STORAGEDEVICEINITIALIZE` for all but `azure`. 

I've lost track of what exactly `ab-pre-push` does, but I've kicked off this job with the hope of creating and testing both `aws` and `azure` VMs: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/931/ (`git ab-pre-push -p aws,azure`)

### DIFF
--- a/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -304,6 +304,20 @@
   when: platform == "azure"
 
 #
+# Supply the default engine parameters
+#
+- lineinfile:
+    path: /opt/delphix/server/etc/delphix_config_override.properties.template
+    regexp: '^{{ item.key }}='
+    line: '{{ item.key }}={{ item.value }}'
+  with_items:
+    #
+    # Enable device initialization by default on all platforms except for Azure
+    #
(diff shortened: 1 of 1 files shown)


jgallag88:
I think we should try to avoid modifying files that are provided by packages (apart from the ones in `/etc` which `dpkg` considers conffiles and handles specially). In fact, I think this might break the 'first boot verification' logic in our jenkins jobs which runs `debsums` to check that these files haven't been modified.

Another approach might be to modify the `delphix_config_override.properties` file in `/var/delphix` that gets created from the `delphix_config_override.properties.template` in `/opt/delphix` when the mgmt stack first starts.

shartse:
I talked to John a bit about this and it seems like the simplest approach would be to add this logic to where we populate these override files in the first place, in `start_mgmt_server_jvm.bash`, to insure that the `/var/delphix` has been created before we try to modify it.

PR #80, opened 2019-05-16
Prevent systemd from creating .device units for zvols

jgallag88:
By default, systemd creates multiple .device units for each block device on
the system: one for the device file in /dev/, one for sysfs representation of
the device, and one for each of the udev-created symlinks to the device. If
the device is partitioned, each of these units is also created for each of the
partitions. For us, creating a zvol typically triggers the creation of 12
.device units. On larger engines, we can easily hit systemd's hard limit of
128K units on a system, at which point systemd stops functioning.

We can work around this by preventing udev from notifying systemd about the
existence of zvols, so that systemd doesn't generate .device units for them.
We don't use these auto-generated units anyway.

### DIFF
--- a/etc/udev/rules.d/999-delphix-systemd.rules
+++ b/etc/udev/rules.d/999-delphix-systemd.rules
@@ -1,0 +1,30 @@
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


jgallag88:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1086/console

prakashsurya:
What are the downsides of this? Does it mean we won't get devices in /dev for zvols?

jgallag88:
We still get devices in `/dev/` and symlinks to the devices in `/dev/zvol`. With this change, we won't get corresponding .device units in systemd though. The man page says that these units "may be used to define dependencies between devices and other units". We don't use .device units for zvols in this way, and I don't think we have plans to, so I can't see any downsides to preventing them from being generated.

prakashsurya:
bors delegate+

jgallag88:
blackbox mssql: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-smoke/79030/console

PR #83, opened 2019-05-20
After migration, switch to the Linux bootloader

prakashsurya:
This change updates the "os-migration" script to switch to the Linux
bootloader after a successful migration; addressing the following Jira
tasks in the process:

 * LX-1554 migration needs boot loader conversion
 * LX-1792 migration: Create dataset for rpool/grub

### DIFF
--- a/var/lib/delphix-platform/os-migration
+++ b/var/lib/delphix-platform/os-migration
@@ -30,6 +30,54 @@ function die() {
 	exit 1
 }
 
+function warn() {
+	echo "$(basename "$0"): $*" >&2
+}
+
+function set_bootfs_mounted_cleanup() {
+	umount "/mnt" || warn "'umount' of '/mnt' failed" >&2
+}
+
+#
+# This function has been copied from the delphix/appliance-build repo's
+# "rootfs-container" script. When doing a migration type of upgrade, we
+# will not have that script available, so we've opted to copy this
+# function from that script, so it can still be used to update the
+# Delphix appliance's bootloader after a migration has completed.
+#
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1115/)

prakashsurya:
cc: @pzakha @sdimitro

prakashsurya:
Cool, thanks. I'll look into moving it.

pzakha:
As a side note, I think it's useful to keep this as a callable script in case it needs to be called manually by support when doing a migration repair.

prakashsurya:
> As a side note, I think it's useful to keep this as a callable script in case it needs to be called manually by support when doing a migration repair.

So, then.. the question is.. do we move the script into delphix-platform, which comes with it's own set of benefits and drawbacks (e.g. always available, but can't use the "new version" during an upgrade)? or do we include the linux upgrade scripts in the migration image (which is consistent with how we do it for linux upgrade)?

pzakha:
I'm fine either way.

Having the script in the upgrade image might reduce the amount of code duplication (or at least confine the code duplication to the same repository), at the expense of not having a fixed path to call the script from. Also until when will those scripts be available (i.e. are the upgrade scripts for a the version we upgraded to cleaned up at some point)?

Having the upgrade scripts in delphix-platform makes it easier to find/invoke at the expense of having some code copied across repositories.

PR #84, opened 2019-05-21
Add "gdb" package dependency

prakashsurya:
The "gdb-python" package has been modified to install its files into
"/opt/gdb-python" instead of "/" to prevent conflicts with the "gdb"
packages supplied by Ubuntu. Thus, we can now have both the Ubuntu "gdb"
package, and our custom "gdb-python" package, installed simultaneously.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -92,6 +92,7 @@ DEPENDS += bcc-tools, \
 	   dstat, \
 	   emacs-nox, \
 	   ethtool, \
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1120/)

PR #86, opened 2019-05-23
Don't use default dependencies with SSH service

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1155/)

prakashsurya:
With this change, `ssh.service` no longer depends on `basic.target`, which means it also doesn't depend on `local-fs.target` anymore (which was the goal, set out in [DLPX-64112](https://jira.delphix.com/browse/DLPX-64112)). See [here](https://www.freedesktop.org/software/systemd/man/bootup.html) for more details on the bootup dependency tree.

Before this change:
```
$ sudo systemctl show -p After ssh.service
After=network.target basic.target system.slice sysinit.target auditd.service -.mount systemd-journald.socket
$ sudo systemctl show -p DefaultDependencies ssh.service
DefaultDependencies=yes

$ sudo systemd-analyze critical-chain ssh.service
The time after the unit is active or started is printed after the "@" character.
The time the unit takes to start is printed after the "+" character.

ssh.service +5.318s
└─basic.target @31.938s
  └─sockets.target @31.937s
    └─dbus.socket @31.935s
      └─sysinit.target @31.931s
        └─cloud-init.service @7.090s +24.837s
          └─systemd-networkd-wait-online.service @7.060s +27ms
            └─systemd-networkd.service @4.799s +2.253s
              └─network-pre.target @4.798s
                └─cloud-init-local.service @2.760s +2.036s
                  └─open-vm-tools.service @2.755s
                    └─vgauth.service @2.751s
                      └─systemd-tmpfiles-setup.service @2.673s +72ms
                        └─systemd-journal-flush.service @2.258s +413ms
                          └─var-log.mount @2.131s +117ms
                            └─local-fs-pre.target @1.587s
                              └─keyboard-setup.service @1.407s +178ms
                                └─systemd-journald.socket @1.368s
                                  └─system.slice @1.366s
                                    └─-.slice @1.134s
```

After this change:
```
$ sudo systemctl show -p After ssh.service
After=network.target -.mount systemd-journald.socket auditd.service system.slice
$ sudo systemctl show -p DefaultDependencies ssh.service
DefaultDependencies=no
$ sudo systemd-analyze critical-chain ssh.service
The time after the unit is active or started is printed after the "@" character.
The time the unit takes to start is printed after the "+" character.

ssh.service +118ms
└─network.target @9.358s
  └─systemd-resolved.service @7.223s +2.132s
    └─systemd-networkd.service @4.948s +2.270s
      └─network-pre.target @4.946s
        └─cloud-init-local.service @2.637s +2.307s
          └─open-vm-tools.service @2.630s
            └─vgauth.service @2.624s
              └─systemd-tmpfiles-setup.service @2.542s +70ms
                └─local-fs.target @2.536s
                  └─run-user-65433.mount @58.491s
                    └─local-fs-pre.target @1.554s
                      └─keyboard-setup.service @1.326s +226ms
                        └─systemd-journald.socket @1.311s
                          └─system.slice @1.306s
                            └─-.slice @646ms
```

prakashsurya:
> which means it also doesn't depend on local-fs.target anymore
Actually, I think this might not be the case. Looking at the "critical-chain" after this change, it appears `local-fs.target` is still a dependency; likely due to the dependency on `network.target` (and perhaps that target's interaction with `network-pre.target` and `cloud-init-local.service`). I'll have to look into how best to override the `After` settings for `ssh.service` (it appears this isn't possible with the normal `override.conf` file).

prakashsurya:
With this second commit, I see the following after a reboot:
```
$ sudo systemctl show -p After ssh.service
After=-.mount systemd-journald.socket system.slice
$ sudo systemctl show -p DefaultDependencies ssh.service
DefaultDependencies=no
$ sudo systemd-analyze critical-chain ssh.service
The time after the unit is active or started is printed after the "@" character.
The time the unit takes to start is printed after the "+" character.

ssh.service +321ms
└─systemd-journald.socket @1.744s
  └─system.slice @1.739s
    └─-.slice @1.390s
```

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -347,3 +347,38 @@
...
     - platform == "gcp"
+
+#
+# We want the ssh service to start as early as possible during boot up,
+# so we explicitly remove all dependencies of the service here. We've
+# configured the service not to use the "default" dependencies, via the
+# "override.conf" file in the delphix-platform package. Unfortuantely
+# though, we can't remove any dependencies specified using "After=" in
+# the original SSH service file, by using that same "override.conf"
+# file. Thus, to remove the "After=" dependencies, we have to create a
```
jgallag88: > we can't remove any dependencies specified using "After=" in the original SSH service file, by using that same "override.conf" file.

Can you elaborate on this? I thought you could typically clear parameters that takes lists as values by specifying the parameter again but without any value.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -347,3 +347,38 @@
...
+# completely new unit definitation for the SSH service, that will
+# override the original unit definitation; Systemd allows us to use
+# files in "/etc/systemd" to replace files in "/lib/systemd".
+#
+# So, the strategy here is to copy the original unit definition from
+# "/lib/systemd" (the file is provided by the SSH package) to
+# "/etc/systemd", and then modify the copy to remove the "After=" line.
+# This way, if the original unit definitation is changed on upgrade,
```
pzakha: typo: definitation

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
prakashsurya: I tried that. It didn’t work. The behavior I saw was that it concatenates all the “After” lines together. So adding “After=“ effectively does nothing. I couldn’t find docs to clearly explain how to clear it, either.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -347,3 +347,38 @@
...
+    remote_src: yes
+    src: /lib/systemd/system/ssh.service
+    dest: /etc/systemd/system/ssh.service
+    owner: root
+    group: root
+    mode: 0644
+
+- lineinfile:
+    path: /etc/systemd/system/ssh.service
+    regexp: "^After="
```
pzakha: In the case the set of `After` dependencies change, we would probably want to be notified of that so that we can review the changes before removing them. I'd propose that we make a strict check for the current `After` dependencies and fail if those are not there.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -347,3 +347,38 @@
 - command: systemctl disable google-accounts-daemon.service
   when:
     - platform == "gcp"
+
+#
+# We want the ssh service to start as early as possible during boot up,
+# so we explicitly remove all dependencies of the service here. We've
+# configured the service not to use the "default" dependencies, via the
+# "override.conf" file in the delphix-platform package. Unfortuantely
```
pzakha: Do you think it would be cleaner to pull all the changes in here instead of relying on both override.conf and an /etc/ file?

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
prakashsurya: If we want to do that, I’d suggest we do this by creating a new package like we discussed. 

IMO, if we do like you describe, we can end up causing upgrade test failures anytime the ssh package changes; and that’s the wrong place to put that sort of notification. 

I think it’d be better to be notified of this sort of thing in the linux-pkg build/update job.

I’m open to seeing what others think, though.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
jgallag88: OK, I guess that only works for some parameters. I see that **systemd.unit(5)** mentions
> Dependencies (After=, etc.) cannot be reset to an empty list, so dependencies can only be added in drop-ins. If you want to remove dependencies, you have to override the entire unit.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
prakashsurya: I don’t. I considered that, but I prefer this two pronged approach so we can still use the override file for all things that can be solved with it. I think the override file is much easier to use/test/maintain, and more reliable, so I’d like to use it over ansible wherever possible.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
pzakha: Okay, I'm fine with leaving it as is in that case. 

I'm still a bit torn whether this is better than having our own package since ssh is such a critical component and we'd want to be aware of what's going on with it. I did a quick test and it takes 4.5 minutes to build `openssh-server` (+1 minute to install build dependencies).

prakashsurya: If you want to get `openssh-server` integrated with linux-pkg, I'm OK with that approach too.

IMO, adding latency to the linux-pkg build is a tangential concern, and we shouldn't let that sway our decision about when to add new downstream packages. If that's a problem, we need to fix the architecture of that project, rather than avoid adding downstream packages.

pzakha: I see both approaches as having their pros and cons, but after thinking a bit more I think it is fine leaving this in `delphix-platform`.

The reason I'm mentioning the build time of the package is because as of right now there are 2 valid approaches and no clear-cut benefit of one over the other. If adding the package was a product requirement then we'd of course add it, even if it increased the build time by a lot.

Regarding linux-pkg, it was designed based on some assumptions (e.g. number of custom packages we will have and upper limit on the total build time of those packages). If we have a need to rethink the architecture of linux-pkg we will do it, but that would incur engineering costs and could add more complexity to the build.

prakashsurya:
@sebroy @basil Do we have a plan on how to verify the Jenkins job concerns laid out in [DLPX-64112](https://jira.delphix.com/browse/DLPX-64112) are present with this change in place? And assuming the problem exists, who would be responsible for fixing the jobs to work after this lands?

basil:
As I noted in [the bug](https://jira.delphix.com/browse/DLPX-64112), I think the Jenkins job changes would need to be implemented as a prerequisite to this product change, since otherwise the Jenkins jobs would be racing with the mounting of the Delphix user's home directory and might fail. As far as who could implement the Jenkins job changes, I could ask Garrett or Jeff. However, this could take a couple of days since I know both of them are currently working on other projects as well.

prakashsurya:
There's no rush to get this in, so I'm OK waiting a couple of days for the Jenkins work to land first.

var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -397,3 +397,9 @@
   with_items:
     - login
     - sshd
+
+- shell: |
```
pzakha: What's the motivation for this code?

prakashsurya: Testing. This won't get integrated.

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1400/) (`internal-qa` only is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1399/))

PR #88, opened 2019-05-30
disable spectre/meltdown/mds mitigations

grwilson:
##### Disabling mitigations
    
Since the platform runs only trusted code, then we rely on the hypervisor
to mitigate any vulnerabilities. Cloud vendors have made the following
statements:
    
Azure:
> Microsoft has deployed mitigations across all our cloud services. The
> infrastructure that runs Azure and isolates customer workloads from each
> other is protected. This means that a potential attacker using the same
> infrastructure can’t attack your application using these vulnerabilities.
>     
> This Azure infrastructure update addresses the disclosed vulnerability at
> the hypervisor level and does not require an update to your Windows or
> Linux VM images. However, as always, you should continue to apply security
> best practices for your VM images.

    
AWS:
> AWS has designed and implemented its infrastructure with protections
> against these types of bugs, and has also deployed additional protections
> for MDS. All EC2 host infrastructure has been updated with these new
> protections, and no customer action is required at the infrastructure level.
>     
> All instances across the Amazon EC2 fleet are protected from all known
> instance-to-instance concerns of the CVEs previously listed. Instance-to-
> instance concerns assume an untrusted neighbor instance could read the
> memory of another instance or the AWS hypervisor. This issue has been
> addressed for AWS hypervisors, and no instance can read the memory of
> another instance, nor can any instance read AWS hypervisor memory. As
> we’ve stated, we haven’t observed meaningful performance impact for the
> overwhelming majority of EC2 workloads.

    
GCP:
> The host infrastructure that runs Compute Engine isolates customer
> workloads from each other. Unless you are running untrusted code inside
> your VMs, no further action is required.

### DIFF
--- a/files/common/etc/default/grub.d/override.cfg
+++ b/files/common/etc/default/grub.d/override.cfg
@@ -22,6 +22,12 @@ GRUB_CMDLINE_LINUX_DEFAULT="console=tty0 console=ttyS0,38400n8"
 GRUB_SERIAL_COMMAND="serial --speed=38400 --unit=0 --word=8 --parity=no --stop=1"
 
 #
+# Disable Spectre/Meltdown/Foreshadow/MDS (Microarchitectural Data Sampling)
+# mitigations to increase performance.
+#
+GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT mitigations=off"
+
+#
 # Disable IPv6 because it is not supported by the Delphix appliance. We
 # do so here, and not from systemd-sysctl, because while configuring
 # net.ipv6.conf.{default,all}.disable_ipv6 works for the currently


grwilson:
@pzakha @sebroy: can you guys take a look?

pzakha:
This looks fine to me, except that from a security stand point, are we sure that there couldn't be cross-instance vulnerabilities in the cloud? I.e. if there is a malicious instance on the same host, could it use side-channels to attack our VM with those mitigations disabled?

Also, note that none of those configurations will make it to the cmd_line at migration time, meaning that the user will have to reboot after migration for those parameters to take effect. You can also add those parameters for migration, but it is non trivial because of some length restrictions in the freebsd bootloader command parser. You can see https://github.com/delphix/appliance-build/pull/242 for more details.

grwilson:
Cross-instance attacks are mitigated by the hypervisor. Here are some comments:

Azure:
> Microsoft has deployed mitigations across all our cloud services. The infrastructure that runs Azure and isolates customer workloads from each other is protected. This means that a potential attacker using the same infrastructure can’t attack your application using these vulnerabilities.

> This Azure infrastructure update addresses the disclosed vulnerability at the hypervisor level and does not require an update to your Windows or Linux VM images. However, as always, you should continue to apply security best practices for your VM images.

AWS:
> AWS has designed and implemented its infrastructure with protections against these types of bugs, and has also deployed additional protections for MDS. All EC2 host infrastructure has been updated with these new protections, and no customer action is required at the infrastructure level.  

> All instances across the Amazon EC2 fleet are protected from all known instance-to-instance concerns of the CVEs previously listed. Instance-to-instance concerns assume an untrusted neighbor instance could read the memory of another instance or the AWS hypervisor. This issue has been addressed for AWS hypervisors, and no instance can read the memory of another instance, nor can any instance read AWS hypervisor memory. As we’ve stated, we haven’t observed meaningful performance impact for the overwhelming majority of EC2 workloads.

GCP:

> The host infrastructure that runs Compute Engine isolates customer workloads from each other. Unless you are running untrusted code inside your VMs, no further action is required.

Since we run only trusted code, then we should be fine disabling the mitigations.

pzakha:
I see, thanks for the details. I think we should add those details as comments, including references to the cloud vendor statements.

What's your plan for migration?

PR #91, opened 2019-05-31
Migration: use the by-id links when importing domain0 (except for aws)

shartse:
When available, we should use the `by-id` links for device names when importing `domain0` after migration. This has to do with persistent `vdev` names allow for device reordering, etc. See DLPX-64115 for further motivation. This is part of a corresponding change in the `app-gate` which creates `domain0` and modifies the `vdevs` in `domain0` using these names. 

AWS does not provide a `by-dev` link, but I've confirmed that `esx`, `azure`, and `gcp` all do.  

Manually tested this on a platform that does have `by-id` links (ESX) and one that does not (AWS). Confirmed that migration works, `domain0` is correctly imported and the `vdevs` and be removed and readded as expected.

### DIFF
--- a/files/common/var/lib/delphix-platform/os-migration
+++ b/files/common/var/lib/delphix-platform/os-migration
@@ -47,7 +47,20 @@ function perform_migration() {
 	chown -R root:root /var/delphix/server ||
 		die "Failed to chown /var/delphix/server"
 
-	zpool import -f domain0 || die "Failed to import domain0"
+	#
+	# In most cases, use the devices' by-id links to import domain0. This ensures
+	# persistent vdev names which can tolerate device reordering. AWS does not
+	# provide a by-id link, so we use simple device names (i.e. /dev/xvda)
+	#
+
+	BY_ID_LINK="/dev/disk/by-id"
+	if [[ -d "$BY_ID_LINK" ]]; then
+		zpool import -f -d $BY_ID_LINK domain0 ||
+			die "Failed to import domain0 by-id"
+	else
+		zpool import -f domain0 || die "Failed to import domain0"
+	fi
+
 	#
 	# domain0/mds is owned by delphix:staff on Illumos, and on Linux
 	# postgres can't access this dataset anymore. We change the permissions


var/lib/delphix-platform/os-migration
```diff
@@ -47,7 +48,13 @@ function perform_migration() {
 	chown -R root:root /var/delphix/server ||
 		die "Failed to chown /var/delphix/server"
 
-	zpool import -f domain0 || die "Failed to import domain0"
+	PLATFORM=$(cat "$PLATFORM_FILE")
+	if [[ "$PLATFORM" = "aws" ]]; then
```
grwilson: I'm wondering if it wouldn't be safer to only use `/dev/disk/by-id` if that directory exists. Or are we assuming that unless we're on AWS, that we better have devids. This also got me thinking about a couple customers where we had to zero out their devids on vmware. I wonder if we need a pre-migration check to verify that all disks in the pool have devids. Or should we try to reproduce that config and see what happens when we migrate them and attempt to address it here.

prakashsurya: Alternatively, would it be reasonable to change the ordering of our "zfs import" logic within ZFS, to prefer using "/dev/disk/by-id" over other method?

One thing I'm concerned with, is after this initial import, if we ever decide not to use the cache file for pool imports, then we'd revert back to not using "/dev/disk/by-id", right? due to the "zfs-import-scan" service just using "zpool import" without "-d".. similarly, if support ever has to manually import, I'm sure they might forget to use "-d".

If we change the ZFS import logic to always prefer "/dev/disk/by-id", this might help us maintain this preference across all imports (outside of this script).

shartse: @prakashsurya I tested out that theory and saw that it's not required to continue passing the `-d` for future imports to keep the `by-id` names. Does this address your concern or do you think there still might be issues? 
```
delphix@sh-import-test:~$ sudo zpool create mypool sdb
delphix@sh-import-test:~$ zpool list -v
NAME        SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
mypool     22.5G   135K  22.5G        -         -     0%     0%  1.00x    ONLINE  -
  sdb      7.50G    49K  7.50G        -         -     0%  0.00%      -  ONLINE  
delphix@sh-import-test:~$ sudo zpool export mypool
delphix@sh-import-test:~$ sudo zpool import -d /dev/disk/by-id mypool
delphix@sh-import-test:~$ zpool list -v
NAME                                       SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
mypool                                    22.5G   184K  22.5G        -         -     0%     0%  1.00x    ONLINE  -
  wwn-0x6000c295924f062c6b531f43bbc8674a  7.50G  65.5K  7.50G        -         -     0%  0.00%      -  ONLINE  
delphix@sh-import-test:~$ sudo zpool export mypool
delphix@sh-import-test:~$ sudo zpool import mypool
delphix@sh-import-test:~$ zpool list -v
NAME                                       SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
mypool                                    22.5G   225K  22.5G        -         -     0%     0%  1.00x    ONLINE  -
  wwn-0x6000c295924f062c6b531f43bbc8674a  7.50G    79K  7.50G        -         -     0%  0.00%      -  ONLINE  
```

shartse: @grwilson I could see either way. The app-stack code is designed to be flexible if there are not `by-id` links available so maybe this should just do best effort and otherwise fall back to the other method.

And yeah - I forgot about the zero'd devids. I think I'd like to do some experiments to see how that's handled when imported into linux. I suspect a `by-id` would repopulate, but I'm not sure.

prakashsurya: > I tested out that theory and saw that it's not required to continue passing the -d for future imports to keep the by-id names.

Partially. If the cachefile exists, I believe it'll use that, which will have the `by-id` names.. so that should help when manually running `zpool import`.

But, in the case where we opt not to use the cache file, I think it won't work correctly; e.g. `zpool import -o cachefile=none` .. or removing the cache file, and then rebooting (which will trigger the `zfs-import-scan` service to import the pool).

If we always rely on the cache file, then this approach is probably fine for our use case (the cache file is properly generated here, and we always use it when importing in the future); but if we sometimes import without the cache file, this approach might lead us to import incorrectly in those cases where we don't use the cache file.

shartse: I tested `zpool import -o cachefile=none` and found that the `by-id` vdev names/ids stayed the same. I also tried deleting the cachefile and rebooting - the `zfs-import-scan` didn't seem to activate but I was able to import the pool normally with both `zpool import -o cachefile=none` and a regular `zpool import`. 

Maybe they're something I haven't tried yet, but it so far it seems pretty hard to get the pool to revert to the old `vdev` names, even without a the cachefile.

prakashsurya: :+1: that all looks good to me.. clearly my understanding of the import process is lacking.. thanks for the verification.

shartse: Definitely worth checking - thanks for the suggestions!

pzakha: Can you add a quick comment as to why we have a special path for AWS?

PR #92, opened 2019-05-31
Directory "/var/log/delphix-upgrade" isn't used anymore

prakashsurya:
The "/var/log/delphix-upgrade" directory is no longer used during
upgrade, so we don't need it to be created via delphix-platform anymore.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1235/)

PR #94, opened 2019-06-03
ESX package needs special files to enable cloud-init

pzakha:
## Description

This allows cloud-init to detect our ESX deployments as using the "NoCloud" datasource. We use the special value "iid-system-uuid" as instance ID to leverage the logic added here: https://github.com/delphix/cloud-init/pull/7.

Most of the logic in this PR is intended to make it easy to add platform-specific files.

## Next Steps

Appliance-build change to enable all cloud-init datasources.

## Testing
- See testing from https://github.com/delphix/appliance-build/pull/304


pzakha:
cc: @jgallag88 @prakashsurya @grwilson

Makefile
```diff
@@ -34,16 +36,25 @@ packages: $(addprefix package-,$(ALL_PLATFORMS))
 
 package-%:
 	@rm -f debian/changelog
+	@rm -rf $(PKG_FILES)
 
 	dch --create --package delphix-platform -v $(VERSION) \
 			"Automatically generated changelog entry."
 
+	# Copy the files that should be in the package to the staging directory.
```
prakashsurya: Why?

Makefile
```diff
@@ -55,16 +66,8 @@ package-%:
...
-	debian/postinst \
-	debian/prerm \
-	usr/bin/download-latest-image \
-	usr/bin/unpack-image \
-	var/lib/delphix-platform/ansible/apply \
-	var/lib/delphix-platform/os-migration
-
 shellcheck:
-	shellcheck $(SHELL_SCRIPTS)
+	shellcheck $$(shfmt -f .)
```
prakashsurya: After a build, will this still work? It won't traverse down into the "stage" directory?

Makefile
```diff
@@ -55,16 +66,8 @@ package-%:
...
 
 shfmtcheck:
-	! shfmt -d $(SHELL_SCRIPTS) | grep .
+	shfmt -d .
```
prakashsurya: Did you verify this will fail when it detects a problem? i.e. the TravisCI checks will properly report the failure?

I'd much rather to see this split into a separate commit, with a test commit to verify the failure modes still work correctly.

pzakha: I've just tested it and it works with travis

Makefile
pzakha: I've tested it, and yeah, it will traverse into the staging directories. Is this an issue?

pzakha:
Almost all this change is related to the directory structure. The only change that is related strictly to ESX are the 2 new files in `files/esx/var/lib/cloud/seed/nocloud/`. Since part of the directory tree change is intended to combine common and platform-specific directories, to be able to fully test this change, we need both of those changes.

That said, I can split all of those changes in separate commits for clarity, however I do not see the point of testing them separately.

Makefile
pzakha: I've updated the comment.

pzakha:
I've split up the work into 3 commits.
- I've tested the make check simplifications both manually in the container and through travis. I've also verified that the files returned by `shfmt -f .` are a superset of the scripts we had before (we missed docker-run and the signature-key one).
- I've added comments to make the new layout more clear

While I could have added those 2 new ESX in a different way, I've decided that having per-platform directories to put files was a cleaner method, which could also be useful for future additions.

@prakashsurya please let me know if you have any other concerns, otherwise I'm ready to integrate.

debian/rules
```diff
@@ -128,6 +128,22 @@ DEPENDS += bcc-tools, \
...
+override_dh_install:
+	#
+	# All the files that we want to include in this package have to
+	# be present in the "stage" directory. Here we move them into 
+	# debian/tmp since it is a special directory and dh_install will
+	# strip "debian/tmp" from the destination path when --auto-dest
+	# is set. 
+	#
+	# Note that the debian/tmp/ directory is cleaned up when
+	# dpkg-buildpkg is ran so we cannot use debian/tmp/ instead of
```
prakashsurya: Can you explain this restriction?

I think it might be cleaner to just use "debian/tmp" instead of "stage", moving the complexity that's currently in `Makefile` to here. E.g.
```
-   mv stage/* debian/tmp/
+
+   cp -r files/common/* debian/tmp/
+   if [ -d files/$(TARGET_PLATFORM)/ ]; then \
+       cp -r files/$(TARGET_PLATFORM)/* debian/tmp/; \
+   fi
+
+   sed "s/@@TARGET_PLATFORM@@/$(TARGET_PLATFORM)/" \
+       debian/tmp/var/lib/delphix-appliance/platform.in \
+       >debian/tmp/var/lib/delphix-appliance/platform
+   rm debian/tmp/var/lib/delphix-appliance/platform.in
+
+   ./scripts/download-signature-key.sh upgrade 5.3 \
+       >debian/tmp/var/lib/delphix-appliance/key-public.pem.upgrade.5.3
+
    dh_install --autodest "debian/tmp/*"
```

In general, I don't like polluting the filesystem with temporary files and directories, like we'd now be doing with the "stage" directory, when it can be avoided. Further, when running `make packages`, we'd clobber the "stage" directory in between each different platform packages build (right?), so I don't see any benefit to using a new "stage" directory over "debian/tmp" directly.

.gitignore
```diff
@@ -21,3 +21,4 @@ debian/delphix-platform.debhelper.log
 debian/delphix-platform.substvars
 debian/delphix-platform
 debian/files
```
prakashsurya: we should have `debian/tmp` listed here too.

this is unrelated to this change, but the 3 `delphix-platform` files listed here should actually be the 3 files specific to each platform's package.. e.g.
```
debian/delphix-platform-kvm.debhelper.log
debian/delphix-platform-kvm.substvars
debian/delphix-platform-kvm
```
we must have left off this change when we moved to the platform specific packages.

prakashsurya:
The link in the testing section doesn't appear to contain the linux-pkg build, that would have built this package. Is that the correct link?

debian/rules
pzakha: Yeah I think moving the Makefile logic into `debian/rules` is cleaner. I'll give it a try.

The restriction was that we can't move files into debian/tmp directly from the Makefile, since debian/tmp would be clobbered when dpkg-buildpacakge is invoked.

prakashsurya:
> While I could have added those 2 new ESX in a different way, I've decided that having per-platform directories to put files was a cleaner method, which could also be useful for future additions.

The implementation/design is good, it just wasn't clear to me when I first looked at the change. First, I didn't notice the new files (partially due to github's "diff" UI).. then I noticed the files, but didn't notice they were in a "esx" directory, which was different than the other files.. then I had to decipher why we needed a "common" directory and an "esx" directory..

So, in this case, it would have been helpful if the PR description could have explained these changes, and explained why they all were needed (having separate commits can help with this too). E.g. "we need a special file when running in ESX, but this file can't exist for other platforms, so we need a way for the package to deliver platform specific files.. this isn't supported, so we need to add support, then use this to add the ESX specific file(s)".

PR #99, opened 2019-06-17
Migrate masking data directory

jgallag88:
The masking home directory has moved between 5.3 and 6.0 and this adds the necessary migration logic for that change. Also, since the id of the `staff` group has changed, it `chowns` the directory so that it is owned by `masking`/`staff`, as was on Illumos.

### DIFF
--- a/files/common/var/lib/delphix-platform/os-migration
+++ b/files/common/var/lib/delphix-platform/os-migration
@@ -47,6 +47,13 @@ function perform_migration() {
 	chown -R root:root /var/delphix/server ||
 		die "Failed to chown /var/delphix/server"
 
+	if [[ -d /var/delphix/dmsuite ]]; then
+		mv -T /var/delphix/dmsuite /var/delphix/masking ||
+			die "Failed to move masking data directory"
+		chown -R masking:staff /var/delphix/masking ||
+			die "Failed to chown masking data directory"
+	fi
+
 	#
 	# In most cases, use the devices' by-id links to import domain0. This ensures
 	# persistent vdev names which can tolerate device reordering. AWS does not


jgallag88:
**Testing**

Set up a masking engine, created and ran a masking job, and then upgraded the engine using a migration image built from this change and additional dms-core-gate changes. After the upgrade succeeded, re-ran the masking job and confirmed that the output was the same as it was before the upgrade.

migration blackbox: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/357/console

jgallag88:
cc @shartse @pzakha

jgallag88:
I have a fix for that issue (the `dms-core-gate` change I mentioned), and I tested this change in conjunction with that one. I should have a review for that one open shortly.

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -47,6 +47,13 @@ function perform_migration() {
 	chown -R root:root /var/delphix/server ||
 		die "Failed to chown /var/delphix/server"
 
+	if [[ -d /var/delphix/dmsuite ]]; then
+		mv /var/delphix/dmsuite /var/delphix/masking ||
```
prakashsurya: Idk if this matters, but I think this will put “dmsuite” under “/var/delphix/masking”, if the “masking” directory already exists...

pzakha: Hmm, that's a good point. We should probably check for that.

jgallag88: Good point. It shouldn't matter, but I'll change it to be `mv -T` to be safe.

PR #100, opened 2019-06-18
DLPX-63558 update unpack-image script in Lumen to accept future wrapper upgrade image and hand-off to complete-unpack script

sanbhara:
This change is part of SUV project. We need unpack logic in Lumen to be able to handle a future modified upgrade image that includes two separate artifacts - the delphix appliance upgrade image, and the verification package. The design of the unpack logic is in doc https://docs.google.com/document/d/1sWXlKR4IvFfwrNAM7bMILT0l9MQby_kFNxSIunRpXDU .

Testing : manually tested with a `meta-tar` based upgrade image with following structure :
```
$ tar -tvf delphix_6.0.0.1_2019-05-30-02-00.upgrade.tar.gz
-rw-r--r-- root/root 5634876696 2019-06-10 22:17 delphix_6.0.0.1_2019-05-30-02-00.upgrade.tar.gz
-rw-r--r-- delphix/staff   2999 2019-06-18 18:26 verify-6.0.0.1_2019-05-30-02-00.tar.gz
-rwxr-xr-x delphix/staff   4979 2019-06-18 18:31 complete-unpack
-rw-r--r-- delphix/staff    386 2019-06-18 18:38 SHA256SUMS
-rw-r--r-- delphix/staff    415 2019-06-17 22:27 file-manifest.json
```

Above image along with a sample complete-unpack script (this is written in bash now but we are considering if using python is better. The changes to unpack-image should be agnostic to what language the complete-unpack script will use) can be accessed at /net/nas/nas/engineering/sanbhara/

automated tests : No new automated tests are planned at the moment since we do not have a meta-tar based upgrade image to test against. Will run existing upgrade tests to ensure change does not affect existing upgrades.

git-ab-pre-push : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1434/

### DIFF
--- a/files/common/var/lib/delphix-platform/os-migration
+++ b/files/common/var/lib/delphix-platform/os-migration
@@ -34,6 +34,13 @@ function perform_migration() {
 	chown -R root:root /var/delphix/server ||
 		die "Failed to chown /var/delphix/server"
 
+	if [[ -d /var/delphix/dmsuite ]]; then
+		mv -T /var/delphix/dmsuite /var/delphix/masking ||
+			die "Failed to move masking data directory"
+		chown -R masking:staff /var/delphix/masking ||
+			die "Failed to chown masking data directory"
+	fi
+
 	#
 	# In most cases, use the devices' by-id links to import domain0. This ensures
 	# persistent vdev names which can tolerate device reordering. AWS does not


files/common/usr/bin/unpack-image
```diff
@@ -96,22 +117,35 @@ tar -xzf "$UPGRADE_IMAGE_PATH" ||
...
+	# verify signatures and hashes before hand-off
+	verify_sig $opt_s
+
+	#
+	# This is newstyle upgrade image that brings with it logic to complete
+	# the unpack process. Hand-off the rest of unpack steps to the script within.
+	#
+	report_progress_inc 50 "Handoff unpack to new script."
+
+	"$(pwd)"/complete-unpack ${f_str:+"-f"} ${s_str:+"-s"} ${x_str:+"-x"} ||
```
prakashsurya: I don't really understand why we use these "_str" variables, vs. just passing in the original value of "$@" passed into this script (we might need to copy it first, since I don't know if `getopts` will mangle it above).

Futher, if we want to use the `${:+}` construct, why not use the "opt_" variables? Why do we need these "_str" variables?

E.g. I'd most prefer something like:
```
"$(pwd)"/complete-unpack "$@"
```
But if we can't do that, perhaps we could do:
```
"$(pwd)"/complete-unpack ${opt_f:+"-f"} ...
```
?

Also, can we drop the use of `"$(pwd)"` here.. I don't see why it's required.. can't we more simply do: `./complete-unpack` ?

files/common/usr/bin/unpack-image
```diff
@@ -96,22 +117,35 @@ tar -xzf "$UPGRADE_IMAGE_PATH" ||
...
+		die "'complete-unpack' hand-off failed"
+
+	# cleanup the temporary directory and exit
+	popd &>/dev/null || die "'popd' failed"
+	cleanup
```
prakashsurya: We use `trap` to call the `cleanup` function, so that we can cleanup even in the event of an error. Unless I'm missing something, there's no need to explicitly call `cleanup` here.

files/common/usr/bin/unpack-image
```diff
@@ -96,22 +117,35 @@ tar -xzf "$UPGRADE_IMAGE_PATH" ||
...
+	# This is newstyle upgrade image that brings with it logic to complete
+	# the unpack process. Hand-off the rest of unpack steps to the script within.
+	#
+	report_progress_inc 50 "Handoff unpack to new script."
+
+	"$(pwd)"/complete-unpack ${f_str:+"-f"} ${s_str:+"-s"} ${x_str:+"-x"} ||
+		die "'complete-unpack' hand-off failed"
+
+	# cleanup the temporary directory and exit
+	popd &>/dev/null || die "'popd' failed"
```
prakashsurya: The flow you're introducing here is awkward to me.. i.e. how we duplicate some of the "post unpack" logic here, and also below, in the case that there isn't a "complete unpack" script available..

I'd rather us do something like this:
```
if [[ -f "complete-unpack" ]]; do
    ... do stuff ...
else
    ... do other stuff ...
fi

popd &>/dev/null || die "'popd' failed"
report_progress_inc 100 "Unpacking successful."
exit 0
```

This way, it's more clear exactly which steps are common to both code paths, and which steps are not common; and we also avoid duplicating the small amount of logic that's common between them.

files/common/usr/bin/unpack-image
```diff
@@ -96,22 +117,35 @@ tar -xzf "$UPGRADE_IMAGE_PATH" ||
 
 report_progress_inc 40 "Verifying format."
 
+# check if this is new style upgrade image
+if [[ -f "complete-unpack" ]]; then
+
+	# verify signatures and hashes before hand-off
+	verify_sig $opt_s
```
prakashsurya: Can we restructure this better, so that both code paths (with and without "complete-unpack") can call this logic at the same place (in which case, a function probably won't been required)? To me, having both code paths do the same thing, but in two different places in the script, makes things more difficult to read and understand (even if the actual logic is in a function)..

files/common/usr/bin/unpack-image
```diff
@@ -96,22 +117,35 @@ tar -xzf "$UPGRADE_IMAGE_PATH" ||
 
 report_progress_inc 40 "Verifying format."
 
+# check if this is new style upgrade image
+if [[ -f "complete-unpack" ]]; then
```
prakashsurya: I'm fine with this name, but if we wanted to be more consistent with the nomenclature used on illumos, we could name this script `prepare`, which matches with `dx_prepare` on illumos that effectively does the same thing. That's largely where the terms `unpack` `upgrade`, `execute`, `verify`, etc. come from; i.e. `dx_unpack`, `dx_upgrade`, `dx_execute`, `dx_verify`, etc.

Ultimately, I'm good either way; this is fine with me.

files/common/usr/bin/unpack-image
sanbhara: So the behavior of the `${:+}` construct seems to require a null value in order for the replacement to not occur. In the script as is, opt_f are all not-null already - with a value of false. My original implementation had me recording the input parameters in a string and passing that in. But this failed the shell formatting check which asked the string variable to be enclosed with double quotes. And enclosing in double quotes makes it a single long input rather than multiple input options.

using `$@` I thought was not available since we use shift. Let me take a look to see if it works.

files/common/usr/bin/unpack-image
sanbhara: Will remove this.

sanbhara:
> This seems reasonable to me.. With that said, I'm hesitant to land this as-is, since it's effectively adding dead code, since I don't see us adding the new `complete-unpack` script in this change.
> 
> If we want to move in this direction, I'd much rather create the new `complete-unpack` script containing the existing logic required to unpack the image.. deal with the flag day that might result.. while avoiding the introduction of code that we can't execute/test/etc.

The new complete-unpack script will not exist in any of our present repos since, along with the verification package itself, this script is going to have to support multiple delphix version upgrade images. We are not yet at a stage where sufficient work towards getting this infrastructure in place can be guaranteed for Lumen timeframe. So any new script we might create will become obsolete once SUV phase 2 ships. At the same time, we are not certain exactly which release will contain SUV phase 2. This is the reason to go this route.

files/common/usr/bin/unpack-image
prakashsurya: Yea, I see.. so maybe your implementation is actually the best we can do, given the constraints.. I think Bash is lacking here, and making this a little more convoluted than it might be in another language..

If we keep your implementation, do we still need to initialize the `_str` variable to the empty string? I'd be OK dropping the initialization if we can avoid it.

files/common/usr/bin/unpack-image
sanbhara: yes I agree. I will move this above the if block.

files/common/usr/bin/unpack-image
sanbhara: ok I'll update this.

prakashsurya:
> So any new script we might create will become obsolete once SUV phase 2 ships.

You mean, the implementation of the script will change, of how the script will be called will change?

IMO, it's perfectly fine to create a `complete-unpack` script that works today, with the assumption that the implementation of it will significantly change later. I think that's the point of having that script delivered with the upgrade image, rather than baked into this script.

> At the same time, we are not certain exactly which release will contain SUV phase 2.

Which is part of my concern... as-is, this code path is not testable, so we don't have any guarantees that'll it'll continue working release after release.. and no guarantee that it'll work by the time we need it for SUV.

Perhaps I don't fully understand the motivations here (please let me know if I'm missing something).. but I think moving the current unpack logic into a script delivered by the upgrade image would be best (e.g. add a new `complete-unpack` script into the appliance-build repository, with the other upgrade scripts).. even with the assumption that, that script will be drastically changed once we implement the rest of the SUV project.

sanbhara:
> > So any new script we might create will become obsolete once SUV phase 2 ships.
> 
> You mean, the implementation of the script will change, of how the script will be called will change?
> 
> IMO, it's perfectly fine to create a `complete-unpack` script that works today, with the assumption that the implementation of it will significantly change later. I think that's the point of having that script delivered with the upgrade image, rather than baked into this script.
> 
> > At the same time, we are not certain exactly which release will contain SUV phase 2.
> 
> Which is part of my concern... as-is, this code path is not testable, so we don't have any guarantees that'll it'll continue working release after release.. and no guarantee that it'll work by the time we need it for SUV.
> 
> Perhaps I don't fully understand the motivations here (please let me know if I'm missing something).. but I think moving the current unpack logic into a script delivered by the upgrade image would be best (e.g. add a new `complete-unpack` script into the appliance-build repository, with the other upgrade scripts).. even with the assumption that, that script will be drastically changed once we implement the rest of the SUV project.

Yes this is an option and I am open to it. But what I mean by this new script will become obsolete is that the new script in appliance-build that we add today, will have to be deleted from appliance-build in SUV phase 2, not just heavily modified. It will have to live in a separate repo not tied to the delphix version. But that should be fine. 

The reason to go with present approach (even though no automated tests can be written for it) is that it has minimal impact on Lumen and does not affect existing upgrade logic in any way. What do you think?

files/common/usr/bin/unpack-image
sanbhara: ya the initialization seems redundant. can remove those. Will wait for us to decide on whether to create the new unpack-image/prepare script in appliance-build repo before addressing the code flow comments.

prakashsurya:
I think we should not be optimizing for having minimal impact on Lumen; I think we should be optimizing for reliability, even if we have to impact Lumen in the process.. especially now, as it's all internal, and we don't have to worry about anything having customer-facing impact.

If we don't use a `complete-unpack` script now, my main concern is something like this..

1. `unpack-image` is tested with your `complete-unpack` script and works today
2. At some later point, `unpack-image` is broken when using `complete-unpack`
3. We don't notice the breakage introduced in (2), since we have no way of testing that code path
4. The broken version of `unpack-image` goes into some release
5. We do the parts of SUV that depend on `complete-unpack`, and only then, realize `unpack-image` is broken

To get past (5), the customer would likely have to go through two upgrades.. one to get the fixed version of `unpack-image`, and then a second one to get the upgrade image that uses `complete-unpack`.

If we start to use a `complete-unpack` script today, then I think we can guarantee that workflow will continue to work, since we have QA/tests that'll continually verify that code path. Thus, we'll never end up in state (5) above. I also think a flag day for Lumen upgrade is small price to pay (especially now, as there's no customer impact), to guarantee the reliability of this new code path.

> But what I mean by this new script will become obsolete is that the new script in appliance-build that we add today, will have to be deleted from appliance-build in SUV phase 2, not just heavily modified. It will have to live in a separate repo not tied to the delphix version. But that should be fine.

Yea, that's not a problem. The important part is to get the "interface" between the upgrade image and this `unpack-image` script solidified; once we do that (e.g. with this new `complete-upgrade` script), it doesn't really matter where the source of the script lives.

sanbhara:
> I think we should not be optimizing for having minimal impact on Lumen; I think we should be optimizing for reliability, even if we have to impact Lumen in the process.. especially now, as it's all internal, and we don't have to worry about anything having customer-facing impact.
> 
> If we don't use a `complete-unpack` script now, my main concern is something like this..
> 
> 1. `unpack-image` is tested with your `complete-unpack` script and works today
> 2. At some later point, `unpack-image` is broken when using `complete-unpack`
> 3. We don't notice the breakage introduced in (2), since we have no way of testing that code path
> 4. The broken version of `unpack-image` goes into some release
> 5. We do the parts of SUV that depend on `complete-unpack`, and only then, realize `unpack-image` is broken
> 
> To get past (5), the customer would likely have to go through two upgrades.. one to get the fixed version of `unpack-image`, and then a second one to get the upgrade image that uses `complete-unpack`.
> 
> If we start to use a `complete-unpack` script today, then I think we can guarantee that workflow will continue to work, since we have QA/tests that'll continually verify that code path. Thus, we'll never end up in state (5) above. I also think a flag day for Lumen upgrade is small price to pay (especially now, as there's no customer impact), to guarantee the reliability of this new code path.
> 
> > But what I mean by this new script will become obsolete is that the new script in appliance-build that we add today, will have to be deleted from appliance-build in SUV phase 2, not just heavily modified. It will have to live in a separate repo not tied to the delphix version. But that should be fine.
> 
> Yea, that's not a problem. The important part is to get the "interface" between the upgrade image and this `unpack-image` script solidified; once we do that (e.g. with this new `complete-upgrade` script), it doesn't really matter where the source of the script lives.

Sounds good. Ill create a new script (did you prefer `prepare`?) in appliance-build repo and open a PR. Let me DM you for workflow and testing questions for the two PRs in appliance-build and delphix-platform.

PR #104, opened 2019-06-25
Allow resuming migration after a failure

pzakha:
This review paves the path for doing migration stress testing.

The main new feature is to allow resuming migration after a failure.

This required some changes in the service dependencies, to make sure that Delphix services do not attempt to start if migration has failed. A RequiredBy dependency is there to make sure the services do not start when delphix-migration is enabled, and a check for the `perform-migration` flag file which is cleared once delphix-migration completes successfully.

We also needed the delphix-migration service to be idempotent. If the service fails, we need to be able to re-run the service after the issue is fixed. To implement this fully, we need to also modify `migrate_config.py` in the app gate (see http://reviews.delphix.com/r/50390).

Finally, we needed a way to restart the `delphix-migration` service and other Delphix services after migration issues are fixed to resume the migration. I've first experimented on how to achieve this through systemd dependencies alone. This is feasible but not trivial, and requires at least 2 commands: first a command to re-enable delphix-migration, and then a command to restart a service/target that would trigger the restart of all the required services through dependencies. Ultimately, I decided to go to a more explicit, albeit less concise approach and created a utility to do that for us. The advantages of the utility is that it is very simple to use and can be used by support to help diagnose and fix migration issues.

## Future Work

- Might want to move the Service class to a separate utility, so that we can expand on it to make helpful tools unrelated to migration.
- Might want to make some improvements to `migrationctl`, e.g. add the `restart` command to retry migration even if delphix-migration have succeeded. We will need to evaluate whether this functionality has any value first.

## Testing
- ab-pre-push (with migration): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1478/
- manually test various migration failure points, and attempt resuming migration with `migrationctl resume`
- test resuming a failed migration following a reboot
- test the `migrationctl status` command at different stages of migration
- test migrationctl on a system that was already successfully migrated & test migrationctl on a system that was Linux native

Example output for `migrationctl status`:
```
$ ./migrationctl status
SERVICE                                  STATE
delphix-migration.service                failed
rtslib-fb-targetctl.service              inactive
delphix-platform.service                 inactive
delphix-postgres@default.service         inactive
delphix-mgmt.service                     inactive

Service delphix-migration.service has failed.
Run "sudo journalctl -u delphix-migration.service" to find out more.

$ ./migrationctl status
SERVICE                                  STATE
delphix-migration.service                active
rtslib-fb-targetctl.service              active
delphix-platform.service                 active
delphix-postgres@default.service         active
delphix-mgmt.service                     active

Migration was successful.

$ ./migrationctl status
SERVICE                                  STATE
delphix-migration.service                active
rtslib-fb-targetctl.service              active
delphix-platform.service                 active
delphix-postgres@default.service         active
delphix-mgmt.service                     activating

Migration is in progress, please wait.

$ ./migrationctl status
Directory /var/delphix/migration/illumos-config missing
Error: System doesn't come from Illumos

$ ./migrationctl status
System rebooted after a successful run of delphix-migration.service.
```

Example output for `migrationctl resume`:
```
$ sudo ./migrationctl resume
Success: Service delphix-migration.service enabled
Starting delphix-migration.service ...
Error: Service delphix-migration.service failed to start
Run "sudo journalctl -u delphix-migration.service" to find out more

$ sudo ./migrationctl resume
Success: Service delphix-migration.service enabled
Starting delphix-migration.service ...
Success: Service delphix-migration.service started
Starting rtslib-fb-targetctl.service ...
Success: Service rtslib-fb-targetctl.service started
Starting delphix-platform.service ...
Success: Service delphix-platform.service started
Starting delphix-postgres@default.service ...
Success: Service delphix-postgres@default.service started
Starting delphix-mgmt.service ...
Success: Service delphix-mgmt.service started
Success: Migration completed
```


pzakha:
cc: @prakashsurya

files/common/lib/systemd/system/delphix-platform.service
```diff
@@ -19,6 +19,14 @@ Description=Delphix Appliance Platform Service
 After=local-fs.target
 Before=rsync.service
 
+#
+# This file is created before performing an OS migration. It is cleared
+# after a succesful run of the delphix-migration service. If the latter
+# has failed then it is unsafe for this service nor its dependencies to
+# run.
+#
+ConditionPathExists=!/var/delphix/migration/perform-migration
```
prakashsurya: Can you comment on under what circumstances this condition would trigger, and prevent this service from running?

Since this service requires `delphix-migration`, I don't understand the purpose of this condition.

files/common/lib/systemd/system/rtslib-fb-targetctl.service.d/override.conf
```diff
@@ -1,3 +1,6 @@
+[Unit]
+After=zfs-mount.service
```
prakashsurya: Why are you adding this dependency?

files/common/usr/bin/migrationctl
```diff
@@ -0,0 +1,282 @@
...
+    MIGRATION_CHECK_DIR = '/var/delphix/migration/illumos-config'
+
+    def __init__(self):
+        Migration.system_check()
+
+        self.svc_migration = Service('delphix-migration.service')
+        self.svc_iscsi = Service('rtslib-fb-targetctl.service')
+        self.svc_platform = Service('delphix-platform.service')
+        self.svc_postgres = Service('delphix-postgres@default.service')
+        self.svc_mgmt = Service('delphix-mgmt.service')
```
prakashsurya: I haven't really given this script a close look.. but why don't we have our other services listed here? e.g. `delphix-stat`, `delphix-fluentd`, `delphix-nginx`, etc.. how did you come to include these services, and exclude others? What about the NFS services? and/or any other system services that may be relevant (e.g. perhaps ssh)?

files/common/lib/systemd/system/rtslib-fb-targetctl.service.d/override.conf
```diff
@@ -6,5 +9,12 @@
...
-[Unit]
 ConditionVirtualization=!container
+
+#
+# This file is created before performing an OS migration. It is cleared
+# after a succesful run of the delphix-migration service. If the latter
+# has failed then it is unsafe for this service nor its dependencies to
+# run.
+#
+ConditionPathExists=!/var/delphix/migration/perform-migration
```
prakashsurya: Same comment as I left on the delphix-platform service change...

files/common/lib/systemd/system/delphix-migration.service
```diff
@@ -38,18 +38,39 @@ After=systemd-udev-settle.service
...
+# that on reboot this services won't race with zfs-import-cache.service
+# which will also try to import domain0.
+#
+Before=zfs-import-cache.service
+
+#
+# When the delphix-migration service is restarted manually following a
+# failure, we need to ensure that it is run services that depend on it.
+#
+Before=rtslib-fb-targetctl.service
```
prakashsurya: IIRC, since you list these as `RequiredBy`, `Before` is redundant.. at least in my recent testing of `delphix.target` that's how it worked..

It would be good to remove these if we can.

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -20,8 +20,8 @@ MIGRATE_CONFIG_LOG="/var/delphix/migration/log"
 PG_REINDEX=/var/delphix/server/db/force-reindex
 
 #
-# The perform-migration flag file is created by dx_execute prior rebooting into
-# Linux.
+# The perform-migration flag file is created by dx_execute prior rebooting
```
prakashsurya: This isn't really a comment on this PR.. but can you remind me why we chose the current design.. where delphix-migration is always enabled by default, but it only "does stuff" when this file is present.. vs. a design where delphix-migration is disabled by default, and dx_execute would enable it for a migrated appliance? I found it a little strange that on first boot of a non-migrated appliance, delphix-migration was running.

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -75,12 +96,17 @@ function perform_migration() {
 	"$MIGRATE_CONFIG_SCRIPT" post-upgrade >>"$MIGRATE_CONFIG_LOG" 2>&1 ||
 		die "Failed migrate_config post-upgrade, see" \
 			"$MIGRATE_CONFIG_LOG"
+	if zfs list rpool/fail-migration &>/dev/null; then
```
prakashsurya: would it be better to use the stress point script/architecture for this?

files/common/usr/bin/migrationctl
pzakha: I've selected the services that are affected by the migration code. I haven't really looked at the other services, but it would probably be a good idea to start them as well if they have failed due to missing dependencies on delphix-platform.

I think once the changes for `delphix.target` are in we can also start that service to make sure that all of the Delphix services are started successfully.

files/common/lib/systemd/system/rtslib-fb-targetctl.service.d/override.conf
pzakha: Ah, I will add a comment here. This is a temporary substitute for the work that should be done in https://jira.delphix.com/browse/DLPX-64724.

files/common/lib/systemd/system/delphix-migration.service
pzakha: The man pages for `systemd.unit` explicitly state that they are not redundant:
```
       Requires=
           ... This has to be
           configured independently with the After= or Before= options. If a unit
           foo.service requires a unit bar.service as configured with Requires=
           and no ordering is configured with After= or Before=, then both units
           will be started simultaneously and without any delay between them if
           foo.service is activated. Often, it is a better choice to use Wants=
           instead of Requires= in order to achieve a system that is more robust
           when dealing with failing services.
```

I'm not sure if your testing was being affected by other dependencies.

files/common/var/lib/delphix-platform/os-migration
pzakha: The main reason is that we cannot run `systemctl enable delphix-migration` in Illumos prior to rebooting the system.

That said, if all that enable does is create the symlinks from the `Install` section, perhaps we can leverage that instead.

files/common/var/lib/delphix-platform/os-migration
pzakha: Oh that was test code, it shouldn't have made it into the review, thanks for pointing that out. And yes indeed, I'm planning to add stress points here once they are supported for migration.

files/common/lib/systemd/system/delphix-migration.service
prakashsurya: Right.. but `RequiredBy` is different than `Requires`.. Here's what `delphix.target` looks like (on a trunk image as of today), even though we don't have any explicit `Before` or `After` directives for it (unless I'm mistaken):
```
$ systemctl show -p After delphix.target
After=delphix-nginx.service delphix-platform.service delphix-stat.service delphix-postgres@default.service delphix-rollback.service delphix-mgmt.service
```

With that said, testing this out for `delphix-migration.service` and `delphix-platform.service`, it doesn't appear to automatically add the `After` dependency, so what you have here is good.

files/common/lib/systemd/system/rtslib-fb-targetctl.service.d/override.conf
prakashsurya: I think it might be better to wait on `zfs.target` instead? Since, this makes me think we specifically care about the mounts, which isn't true.. idk if using `zfs.target` will impact anything practically, but it seems more appropriate conceptually (at least until we fix DLPX-64724).

files/common/usr/bin/migrationctl
prakashsurya: > I think once the changes for delphix.target are in we can also start that service to make sure that all of the Delphix services are started successfully.

`delphix.target` is in the latest trunk builds.. I think we'll need to restart/start this target, since otherwise, we'll fail to restart/start all the other services that comprise the appliance.. as-is, I think they'd fail due to `delphix-platform.service` being down, and then we'd never bring the back up. Have you tested this at all?

files/common/lib/systemd/system/rtslib-fb-targetctl.service.d/override.conf
pzakha: I was about to add this comment:
```
#
# This service depends on zvol dev links to be present, which should
# be implemented in DLPX-64724. In the meanwhile, we have this service
# depend on zfs-mount.service, which doesn't guarantee that the zvol
# links will be there, but gives a time buffer after pools have been
# loaded to create the zvol links. This new dependency is especially
# important since local-fs.target, which this service depends on, no
# longer depends on zfs-mount.service.
#
```

In any case I'm working on fixing DLPX-64724 right now, so maybe I'll just update this if I get it fixed first.

files/common/usr/bin/migrationctl
pzakha: Yes, I've tested this for all the other service. I've rebased my code on top of the delphix.target changes and will be able to test it soon. Let me push my new changes (not ethat I will change the commit history later).

prakashsurya:
I know we've already talked about this offline, but I still feel like this is adding a lot of code and complexity when (IMO) we could accomplish this more elegantly by leaning more heavily on systemd. For example, we can get the status of all services with a few lines of Bash:
```
$ for svc in $(systemctl show -p Wants delphix.target | sed 's/Wants=//'); do
> state=$(systemctl show -p ActiveState "$svc" | sed 's/ActiveState=//');
> printf "%-35s %10s\n" "$svc" "$state";
> done
delphix-nginx.service                   active
delphix-stat.service                    active
delphix-rollback.service              inactive
delphix-mgmt.service                    active
delphix-platform.service                active
delphix-migration.service               active
delphix-postgres@default.service        active
```

And then after restarting `delphix.target`:
```
$ sudo systemctl restart delphix.target &
$ for svc in $(systemctl show -p Wants delphix.target | sed 's/Wants=//'); do
> state=$(systemctl show -p ActiveState "$svc" | sed 's/ActiveState=//');
> printf "%-35s %10s\n" "$svc" "$state";
> done
delphix-postgres@default.service        active
delphix-stat.service                    active
delphix-mgmt.service                activating
delphix-platform.service                active
delphix-rollback.service              inactive
delphix-nginx.service                   active
```

And if I re-enable the migration service, and modify it to force it to fail, and then restart `delphix.target`:
```
$ sudo systemctl enable delphix-migration
Created symlink /etc/systemd/system/delphix.target.wants/delphix-migration.service → /lib/systemd/system/delphix-migration.service.
Created symlink /etc/systemd/system/delphix-platform.service.requires/delphix-migration.service → /lib/systemd/system/delphix-migration.service.
$ sudo systemctl restart delphix.target &
$ for svc in $(systemctl show -p Wants delphix.target | sed 's/Wants=//'); do
> state=$(systemctl show -p ActiveState "$svc" | sed 's/ActiveState=//');
> printf "%-35s %10s\n" "$svc" "$state";
> done
delphix-stat.service                    active
delphix-migration.service               failed
delphix-rollback.service              inactive
delphix-postgres@default.service        active
delphix-platform.service                active
delphix-nginx.service                   active
delphix-mgmt.service                activating
```

If I haven't convinced you of this already, I doubt I'm going to convince you now, but I at least wanted to get my opinions into a comment here.

files/common/lib/systemd/system/rtslib-fb-targetctl.service.d/override.conf
prakashsurya: If we don't fix DLPX-64724 first, I still think using `zfs.target` would be better than `zfs-mount.service`.. `zfs.target` should activate after `zfs-mount.service`, so practically I think it's the same.. But I guess I'm fine with this way too.

pzakha:
> I think we should remove the `ConditionPathExists` directive being added to the services, since I don't think that's needed (i.e. adds complexity for no benefit).

Now that you mention it, I've realized this was supposed to be:
```
ExecStartPre=/bin/bash -xc "test ! -e /var/delphix/migration/perform-migration"
```
instead of ConditionPathExists. I had this on my test system but haven't updated it in the code. ConditionPathExists is kinda bad since it still lets dependent services go through.
Regarding the new check, I _could_ remove it, since theoretically the migration service would only get disabled after it has run successfully so it shouldn't be needed. It's more of an extra precaution check to make sure we do not accidentally start the other services before migration is fully completed.
 
> How do we plan to ensure the migration service is idempotent? E.g. if we used something like Ansible, it would help us ensure it's idempotent, as we could run Ansible twice, and it'd tell us if it changed anything. Since we're not using a tool like that, what are our plans to ensure idempotency? I worry that we may introduce subtle differences between a system where migration was successful on the first run, and another system where migration had to be restarted/resumed.

I see 2 ways of accomplishing this:
1. We will have stress testing, which will introduce failure points at different stages of migration and then run some tests on the system to make sure it behaves as expected.
2. By inspection. The `os-migration` script is fairly straight forward. The `migrate_config` script is a bit more complicated, but is already designed to be idempotent (except for the issue here: http://reviews.delphix.com/r/50390)

I think Ansible would actually make things more complicated in this scenario. With a shell script we control exactly which commands are executed, which is what we want with migration. Ansible would also have to call a few shell commands for which there are no Ansible equivalent, which would reduce its value even further.

prakashsurya:
Sorry, I didn’t mean to imply that we should use Ansible here. That was only meant to highlight that the tooling (or lack thereof) were using here makes it difficult to guarantee the migration service is idempotent.

> It's more of an extra precaution check to make sure we do not accidentally start the other services before migration is fully completed.

How would we accidentally start the other services? Without a clear expectation of when this would be used, I don’t understand why we’d want to rely on it to enforce correctness. Even if we did accidentally start the services, I’d suggest we fix that bug, than rely on this condition.

files/common/lib/systemd/system/delphix-platform.service
pzakha: We talked about this offline, so I'm just putting this for the comments. This is just an additional failsafe that kicks in if delphix-migration is manually disabled before it has run successfully. Also, it should be replaced by `ExecStartPre`.

files/common/lib/systemd/system/rtslib-fb-targetctl.service.d/override.conf
pzakha: Following my tests in https://github.com/delphix/zfs/pull/56, I've changed this to `zfs.target` as it is indeed better.

pzakha:
Latest ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1534/ (failure is DLPX-64355)

PR #106, opened 2019-06-26
DLPX-63558 update unpack-image script in Lumen to accept future wrapper upgrade image and hand-off to complete-unpack script

sanbhara:
This review updates unpack-image to hand over unpack duties to a new prepare script (being added in PR https://github.com/delphix/appliance-build/pull/325) to complete the unpack. This design is necessary for SUV phase 2 since the structure of the upgrade image will change with the introduction of a new artifact - the verification package - outside of payload.tar.gz. This change in architecture at a future release (potentially 6.0.2.0) requires Lumen to have logic that can upgrade to versions with and without the updated architecture.

This hand-off design also gives us flexibility in introducing changes to the upgrade image as needed without having to roll-out such changes over multiple releases with prior releases having the necessary logic to handle the changes.

`Manual Tests:`
Manual test of unpacking a new upgrade image generated with the new prepare script using this unpack-image script was successful. 

```
tar -tvf internal-dev.upgrade.tar.gz 
-rw-r--r-- root/root       512 2019-06-25 00:05 SHA256SUMS.sig.5.3
-rw-r--r-- root/root       234 2019-06-25 00:05 SHA256SUMS
-rw-r--r-- root/root      2330 2019-06-25 00:05 version.info
-rwxrwxr-x 1000/1000      3939 2019-06-24 23:57 prepare
-rw-r--r-- root/root 3424779189 2019-06-25 00:05 payload.tar.gz

sudo ./unpack-image -f internal-dev.upgrade.tar.gz 
Progress increment: 19:06:44:697930129+0000, 10, Extracting upgrade image.
Progress increment: 19:08:00:434560383+0000, 40, Verifying format.
Progress increment: 19:08:21:353591869+0000, 50, Handoff unpack to prepare script.
Progress increment: 19:09:53:812158271+0000, 90, Prepare completed successfully.
Progress increment: 19:09:53:814707365+0000, 100, Unpacking successful.

cat /var/dlpx-update/latest/SHA256SUMS
3c36885ad57a1ec5ac3c4aa62593cf4dad565b8a16394a76f6b39b6c4b560b34  payload.tar.gz
f0517a5358422c707cd700fdafb784d3d558f9f1a089d3dcf849ab797f4bd88e  version.info
8ee737bc908ea6ae60344a34461462234f1837995fd2fa0c1100fcd653068e5a  prepare
```

`Automated Tests:`
git-ab-pre-push with updated parameter for appliance-build commit from PR https://github.com/delphix/appliance-build/pull/325 SUBMITTED : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1517/

git ab-pre-push after https://github.com/delphix/appliance-build/pull/325 merged is CLEAN : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1529/ 
vm: upgrade-testing-1739-8a5cd044.dlpxdc.co

### DIFF
--- a/files/common/usr/bin/unpack-image
+++ b/files/common/usr/bin/unpack-image
@@ -42,7 +42,7 @@ function cleanup() {
 # the same version. Otherwise, this will fail if there's already an
 # unpacked image of the same version.
 #
-opt_f=false
+opt_f=""
 
 #
 # This option will skip the signature verification portion of this
@@ -50,7 +50,7 @@ opt_f=false
 # ensure the image hasn't been modified, but it can be useful for
 # development and/or testing purposes.
 #
-opt_s=false
+opt_s=""
 
 #
 # This option will cause this script to modify the "version.info" file,
@@ -60,7 +60,7 @@ opt_s=false
 # perhaps this otherwise would not have occurred; this is mostly useful
 # for internal testing of the "not-in-place" upgrade code paths.
 #
-opt_x=false
+opt_x=""
 
 while getopts ':fsx' c; do
 	case "$c" in
@@ -85,6 +85,7 @@ UPGRADE_IMAGE_PATH="$(readlink -f "$1")"
 [[ -n "$UPGRADE_IMAGE_PATH" ]] || die "unable to determine upgrade image path"
 
 trap cleanup EXIT
+
 UNPACK_DIR=$(mktemp -d -p "$UPDATE_DIR" -t unpack.XXXXXXX)
 [[ -d "$UNPACK_DIR" ]] || die "failed to create unpack directory '$UNPACK_DIR'"
 pushd "$UNPACK_DIR" &>/dev/null || die "'pushd $UNPACK_DIR' failed"
@@ -96,11 +97,11 @@ tar -xzf "$UPGRADE_IMAGE_PATH" ||
 
 report_progress_inc 40 "Verifying format."
 
-for file in SHA256SUMS payload.tar.gz version.info; do
+for file in SHA256SUMS prepare; do
 	[[ -f "$file" ]] || die "image is corrupt; missing '$file' file"
 done
 
-if ! $opt_s; then
+if [[ -z "$opt_s" ]]; then
 	openssl dgst -sha256 \
 		-verify /var/lib/delphix-appliance/key-public.pem.upgrade.5.3 \
 		-signature SHA256SUMS.sig.5.3 \
@@ -113,42 +114,12 @@ fi
 sha256sum -c SHA256SUMS >/dev/null ||
 	die "image is corrupt; checksums don't match"
 
-tar -xzf payload.tar.gz || die "failed to extract payload.tar.gz"
-rm payload.tar.gz || die "failed to remove payload.tar.gz"
-
-#
-# We need to be careful when sourcing this file, since it can conflict
-# with (and clobber) functions and/or variables previously defined.
-#
-# shellcheck disable=SC1091
-. version.info || die "sourcing version.info file failed"
-
-[[ -n "$VERSION" ]] || die "VERSION variable is empty"
-[[ -n "$MINIMUM_VERSION" ]] || die "MINIMUM_VERSION variable is empty"
-[[ -n "$MINIMUM_REBOOT_OPTIONAL_VERSION" ]] ||
-	die "MINIMUM_REBOOT_OPTIONAL_VERSION variable is empty"
-
-if $opt_x; then
-	sed -i \
-		"s/^\\(MINIMUM_REBOOT_OPTIONAL_VERSION\\)=.*$/\\1=$VERSION/" \
-		version.info ||
-		die "'sed -i ... version.info' failed"
-
-	# shellcheck disable=SC1091
-	. version.info || die "sourcing version.info file failed"
-fi
-
 popd &>/dev/null || die "'popd' failed"
 
-$opt_f && rm -rf "${UPDATE_DIR:?}/$VERSION" >/dev/null 2>&1
+report_progress_inc 50 "Handoff unpack to prepare script."
 
-[[ -d "$UPDATE_DIR/$VERSION" ]] && die "version $VERSION already exists"
-
-mv "$UNPACK_DIR" "$UPDATE_DIR/$VERSION" ||
-	die "failed to move unpacked upgrade image to $UPDATE_DIR/$VERSION"
-
-rm -f "$UPDATE_DIR/latest" || die "failed to remove 'latest' symlink"
-ln -s "$VERSION" "$UPDATE_DIR/latest" || die "failed to create 'latest' symlink"
+"$UNPACK_DIR"/prepare ${opt_f:+"-f"} ${opt_s:+"-s"} ${opt_x:+"-x"} "$UNPACK_DIR" ||
+	die "'prepare' hand-off failed"
 
 report_progress_inc 100 "Unpacking successful."
 


files/common/usr/bin/unpack-image
```diff
@@ -42,15 +42,15 @@ function cleanup() {
 # the same version. Otherwise, this will fail if there's already an
 # unpacked image of the same version.
 #
-opt_f=false
+opt_f=""
```
prakashsurya: Let's just remove the declaration of these options, and their comments. I see you copied the comments over to the new `prepare` script, and I think that's the more appropriate place for them. And these declarations don't do anything special, so I think removing them would be best (the script will work the same).

files/common/usr/bin/unpack-image
```diff
@@ -113,43 +113,13 @@ fi
...
-if $opt_x; then
-	sed -i \
-		"s/^\(MINIMUM_REBOOT_OPTIONAL_VERSION\)=.*$/\1=$VERSION/" \
-		version.info ||
-		die "'sed -i ... version.info' failed"
-
-	# shellcheck disable=SC1091
-	. version.info || die "sourcing version.info file failed"
-fi
+./prepare ${opt_f:+"-f"} ${opt_s:+"-s"} ${opt_x:+"-x"} "$UNPACK_DIR" ||
```
prakashsurya: How would you feel about moving the `popd` from a few lines below this one, to occur before this call to `prepare`.. and then we change this line to be:
```
"$UNPACK_DIR/prepare" ${opt_f:+"-f"} ${opt_s:+"-s"} ${opt_x:+"-x"} || die ...
```

My reasoning is this:

 1. This will ensure there's no processes with a reference on `$UNPACK_DIR` when `prepare` is called. While this might not be a problem today, I worry that if the process executing `unpack-image` is holding a reference on `$UNPACK_DIR`, it might restrict what `prepare` can do to that directory. I don't have any concrete examples of why this would be required, but I feel like it makes things a little more clean, as far as the assumptions that `prepare` can have about the state of the system.

 2. I don't see any reason as to why we need to pass `$UNPACK_DIR` as an argument to `prepare`, since `prepare` can easily determine this directory for itself via `"${BASH_SOURCE%/*}"` and/or `readlink -f "${BASH_SOURCE%/*}"`.. We even use this approach in other the upgrade scripts today.. So, we can simplify the interface between `unpack-image` and `prepare`, by removing the `$UPACK_DIR` argument.

prakashsurya: How do we want to handle "cleanup" when `prepare` fails?

Do we want `unpack-image` to be responsible for removing `$UNPACK_DIR` when `prepare` fails? Or do we want to `prepare` to be responsible for removing `$UNPACK_DIR` when it fails? Or do we want both to attempt the removal (presumably, one of the removals will fail and/or be a no-op, due to racing with the other)?

Right now, I think both scripts will attempt the removal, but I want to make sure this is by design, rather than by accident.

files/common/usr/bin/unpack-image
sanbhara: Can remove the comments but since this is the script that any customer orchestration would actually use, I believe we should leave it in there. Unless of course if these options are meant to be hidden (I do not get why the -s option should ever be exposed in production version of this script). But removing the comments would not effectively achieve hiding the options either so I am inclined to leave it in here. If you feel they're redundant in two scripts, I vote for removing the comments in the new prepare script instead. What do you think?

The shell format check complains about using uninitialized variables when using these when calling the prepare script, so will need to keep the initializations I believe.

files/common/usr/bin/unpack-image
sanbhara: Yes I had planned to do this but seem to have missed it. No reason for popd in unpack-image to happen after prepare script finishes.

sanbhara: This is by design. No harm in checking if this temp directory still exists after the prepare script returns and clean up in any case that the directory is left behind.

files/common/usr/bin/unpack-image
prakashsurya: We can suppress the static analysis error. I still vote to remove the declarations and the suppress the error, but if you disagree, I won’t push for it; I’m OK either way.

And yea, this script isn’t exposed to customers, but at the same time the changes I’m suggesting wouldn’t change the external interface of this script, so I don’t think it’d matter even if we did expose this to customers.

files/common/usr/bin/unpack-image
sanbhara: Have moved the popd up. Regarding the UNPACK_DIR parameter, I would like to be a little conservative and pass it in. There has been some suggestions that the replacement script to prepare in SUV be written in a different language than bash (python is primary suggestion) and I am not sure how reliable getting the directory using something like `os.path.dirname(__file__)` in python is, compared to the bash equivalent. What do you think?

files/common/usr/bin/unpack-image
sanbhara: Cool, I have removed the comments. I do like the initializations just for readability since the code where these variables are constructed do not make them obvious.

files/common/usr/bin/unpack-image
prakashsurya: I think it's still simple to get it with python:

> As initialized upon program startup, the first item of this list, path[0], is the directory containing the script that was used to invoke the Python interpreter.

See: https://docs.python.org/3/library/sys.html#sys.path

But if you prefer to pass in the directory using an argument, I'm OK with that.

files/common/usr/bin/unpack-image
```diff
@@ -113,42 +92,12 @@ fi
...
-
-[[ -d "$UPDATE_DIR/$VERSION" ]] && die "version $VERSION already exists"
-
-mv "$UNPACK_DIR" "$UPDATE_DIR/$VERSION" ||
-	die "failed to move unpacked upgrade image to $UPDATE_DIR/$VERSION"
+report_progress_inc 50 "Handoff unpack to prepare script."
 
-rm -f "$UPDATE_DIR/latest" || die "failed to remove 'latest' symlink"
-ln -s "$VERSION" "$UPDATE_DIR/latest" || die "failed to create 'latest' symlink"
+"$UNPACK_DIR"/prepare ${opt_f:+"-f"} ${opt_s:+"-s"} ${opt_x:+"-x"} "$UNPACK_DIR" ||
```
prakashsurya: This is relatively minor optimization, but what if we used `exec` here.. e.g.
```
-"$UNPACK_DIR"/prepare ${opt_f:+"-f"} ${opt_s:+"-s"} ${opt_x:+"-x"} "$UNPACK_DIR" ||
+exec "$UNPACK_DIR"/prepare ${opt_f:+"-f"} ${opt_s:+"-s"} ${opt_x:+"-x"} "$UNPACK_DIR" ||
    die "'prepare' hand-off failed"
-
-report_progress_inc 100 "Unpacking successful."
-
-exit 0
```
This way, we'll avoid creating another process to run `prepare`.. (we'll have to report "100" progress in `prepare`)

The only drawback to this, is we won't call `cleanup` from `unpack-image` anymore (at least, not once we call `prepare`).. so we'll rely more on the `cleanup` in `prepare`..

sanbhara:
> Looks good to me.. What's the plan for landing this? Can we land the appliance-build change first, and then this one afterwards, to avoid a flag day?

Yup, landing appliance-build first is the plan. That should not affect upgrade tests or any other tests. Once we have a new build with the new commit, we can plan to re-test ab-pre-push on this branch and then push it. I am guessing no VM can have a newer delphix-platform but still be on an older appliance-build.

prakashsurya:
:+1: Sounds good.

> I am guessing no VM can have a newer delphix-platform but still be on an older appliance-build.

It's difficult enough to do this, that it's not worth our trouble to support. It'd only occur if somebody manually built the delphix-platform package, and installed it, without going through the normal build/upgrade process. Definitely not a workflow we need to worry about.

files/common/usr/bin/unpack-image
sanbhara: exec replaces the process executable with the new script correct? I had considered three ways of invoking the prepare script, using `source` (not a good idea since that will fix the `prepare` script to be a shell script), executing the script directly and something like exec. It feels safest to invoke the other script in its own execution `shell` to handle unexpected failures in the script that's shipped with the new image. Since it's not a long running process, I feel we can keep it simple with the present flow. Let me know if you still feel strongly about using exec here.

files/common/usr/bin/unpack-image
sanbhara: Hmm just realized I am only using the `-s` option in the unpack-image script at the moment. I will continue to pass it into prepare since we are not yet sure if there will be necessity for signature checks within the prepare script for any additional artifacts like the verification package (there shouldn't be. The single SHA256SUMS file should in theory have the hashes for all artifacts and this file will be signed but I would rather pass in more info than less in case we need this param for our test environment).

In any case, given the parameter is used in unpack-image, I am considering moving the comments over from prepare to unpack-image instead.

sanbhara: Also realizing an empty string does not evaluate to false in bash. I am going to need separate variables to pass along to prepare like I had before.

sanbhara: Checking for empty string removes need for separate variable. Using that in latest upload.

files/common/usr/bin/unpack-image
prakashsurya: AFAIK, `exec` will just make it so the current process (`unpack-image`) is replaced with the new process (`prepare`).

It's practically the same as executing `prepare` without `exec`, except we won't return control back to `unpack-image` after `prepare` completes (which, to me, makes this hand-off even more clear, since it'd now be represented even in the process tree). Unless I'm mistaken, using `exec` won't change anything with respect to safety or the ability to deal with errors.

But this is minor, so I'll leave it up to you.

files/common/usr/bin/unpack-image
prakashsurya: I don't think we should pass in `-s` if it's not used by `prepare`. I also don't think `prepare` should accept `-s` if it doesn't use that option.

I can easily imagine `prepare` being modified later to remove `-s`, due to somebody noticing it's not used in `prepare`, but not realizing `unpack-image` uses it for no reason.

sanbhara: The concern here is with upgrading from Lumen to a future prepare script created in SUV to handle unpacking of the new meta-tar based upgrade image. At the moment I do not believe we have sufficient information to decide if this parameter will NOT be needed by that script.

sanbhara: Ill think a bit more about this and our plans for the build process in SUV and update the branch.

prakashsurya: > At the moment I do not believe we have sufficient information to decide if this parameter will NOT be needed by that script.

But this can be said about any possible option that can potentially be passed in. Based on this argument, we should accept all letters of the alphabet, and pass them to `prepare`, even if they're not used today.

Further, I don't think `-s` is even used by the product. IIRC, it's only there for development purposes, right? (I could be mistaken, I haven't reviewed the product code to verify this claim)

sanbhara: I think there is a misunderstanding with my statement. The concern is with respect to parameters being accepted by the lower version script which is a definite set we know of. To pass in all of these known set of parameters is all I am considering, given the fact that we are in effect splitting one script into two. Regarding the specific parameter -s, I too do not know where it is used but it is not passed in from the upgrade api that handles upload of the upgrade image. That code path only passes in the image file name so all the parameters are for internal use only, potentially in our qa tests.

prakashsurya: Perhaps I took it too far in my prior comment...

So, thinking a little more broadly, are you suggesting we adopt the design that `unpack-image` and `prepare` must maintain the same interface, and that `unpack-image` must communicate to `prepare` all of the options specified to it when it was executed?

My issue with this design, is `prepare` will need to work with multiple different versions of `unpack-image`. So, I can imagine a situation where we remove `-s` from `unpack-image` in some later release, or we change the meaning of `-s`.. I'm not sure how `prepare` would work in this situation, since `prepare` would have to match the interface of two different versions of `unpack-image`.

Perhaps we could make this work, though, since `prepare` will know the version of `unpack-image` being used, so it could interpret the options differently based on the version of `unpack-image` that's calling it?

If you're only suggesting that `prepare` accepts `-s`, without any broader design goals, I find it difficult to justify this, since `-s` has no meaning to `prepare` right now (and we have no way to test and verify `-s` works properly).

Ultimately, I'm OK with whatever approach you want to take here. I may not fully agree, but I don't think it's worth blocking this PR over it either. If you want to pass `-s` from `unpack-image` to `prepare`, that's fine with me.

sanbhara: > So, thinking a little more broadly, are you suggesting we adopt the design that unpack-image and prepare must maintain the same interface, and that unpack-image must communicate to prepare all of the options specified to it when it was executed?

Correct. The design reason is that at the time a particular version is released, all parameters used in its unpack-image script are deemed necessary. It would be reasonable to pass it along to the prepare script from a newer release and let that script handle the parameter correctly.

> My issue with this design, is prepare will need to work with multiple different versions of unpack-image. So, I can imagine a situation where we remove -s from unpack-image in some later release, or we change the meaning of -s.. I'm not sure how prepare would work in this situation, since prepare would have to match the interface of two different versions of unpack-image.

Good point. This is inherent in the design of SUV that the `prepare` script in SUV will handle a relatively complex matrix of upgrades since it will be a single script delphix-wide. So parameters will be just one aspect among many that could be treated in a conditional manner based on the current_version. Having said that, I do not expect the list of parameters to expand too large and neither to change too often so I think we are fine in that respect.

By removing a parameter, I am worried we might be handicapping ourselves when the time comes to write and ship the new SUV script. Since there are no security/performance concerns with this, let's pass in this parameter at this time.

prakashsurya: OK. I think this is making more sense to me now.

Can you document this design someplace? Maybe a comment in the “prepare” script?

sanbhara:
@prakashsurya ab-pre-push is clean on this now that the prepare script is merged. I have verified on the VM that it is using the updated unpack-image script and the latest upgrade images have the prepare script as well. I am ready to merge this now.

prakashsurya:
bors delegate+

PR #108, opened 2019-07-07
DLPX-64977 increase start rate limit for systemd-networkd

grwilson:


files/common/etc/systemd/system/systemd-networkd.service.d/override.conf
```diff
@@ -20,3 +20,12 @@
...
 Environment=SYSTEMD_LOG_LEVEL=debug
+
+#
+# We use "netplan apply" to make network changes which causes
+# a restart of the systemd-networkd service. Since we can
+# have any number of network configuration changes, we may exceed
+# the service's default restart limit causing the service to
+# fail. To avoid this we disable the restart limit here.
+#
+StartLimitBurst=0
```
pzakha: Should we put some reasonable limit in here to avoid having the service restart endlessly in case of failure?

grwilson: I'm open to a number but I'm not sure what that should be. Our unit tests generate tons of configuration changes and will be the first to hit the limit. I'm also not sure if it's better to have the service restart continuously or end up in a failed state. Either way, I'm happy to put a number here. Any suggestions?

pzakha: Anything that is above our unit test limit would do. Do you think something like 20 would be sufficient? Even 100 would be better than unlimited. The reason I don't like unlimited restart is that it fills up the logs fairly quickly and makes it harder to diagnose the original failure.

sebroy:
In the longer term, it seems like it would be good for us to think about how we apply multiple network configuration changes. There are a number of problems with the current approach, with each distinct piece of configuration coming from a separate web service API call, and being applied separately:
1. The problem outlined in this PR
2. Each individual API call may render the system unreachable, while taken as a whole, the set of calls is a complete configuration that makes the system reachable. For example, consider one of the most frequent set of network configuration operations, which is to replace the DHCP address that comes with the engine, with a static IP address, a static route, static DNS configuration, and a static hostname. That requires 5 distinct calls, but after the 1st operation (delete the DHCP address), the system is unreachable.

To solve that, I think we should perhaps add some kind of atomicity or a kind of transaction semantic to network configuration in order to apply a set of settings all at once. Either a commit operation after a set of distinct API calls, or one uber network configuration API object that takes a "final" configuration as an object, and the system figures out how to get there most optimally.

Anyway, I'm good with this PR, this is just what I was thinking when reading it.

files/common/etc/systemd/system/systemd-networkd.service.d/override.conf
grwilson: I ran some tests and using a value of 30 was enough to allow the network unit tests to pass. I will use that value.

grwilson:
@sebroy I completely agree.

grwilson:
ab-pre-push [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1586/)

PR #110, opened 2019-07-11
Don't use SSH idle timeout for internal variants

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1614/)

This change depends on the following PR: https://github.com/delphix/appliance-build/pull/330

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -193,15 +193,34 @@
...
+# our engine.
+#
+- lineinfile:
+    path: /etc/ssh/sshd_config
+    regexp: "^#?{{ item.key }} "
+    line: "{{ item.key }} {{ item.value }}"
+  with_items:
     - { key: "ClientAliveInterval", value: "1800" }
     - { key: "ClientAliveCountMax", value: "0" }
+  when:
```
pzakha: should we also add `internal-qa`?
If so, we can just use ansible's `match()` function to match all internal systems

prakashsurya: No, I don’t think so. I specifically left the QA variant out so it’d more closely match the external variants.

You think the QA variant should differ in this respect from the external variants?

pzakha: I was wondering if the idle timeout was causing any issues during testing. If not, I also think it's best to leave off `internal-qa` out of it so that it is closer to the external variant.

prakashsurya: I don't know of any issues with the timeout, so I'll leave the QA variant alone.

IMO, even if it did pose a problem to the qa-gate tests, I'd recommend we fix the qa-gate to account for the timeout, than modify the appliance, so that we remain consistent with the external variants.

Regardless, I'll leave this as-is.

pzakha: Might wanna mention in comments that we explicitly leave off the QA variant.

prakashsurya: I can do that.

I opted not to comment on that since the default stance should be to remain consistent with the external variants, which I'm adhering too. If I was doing something different than default, then I would have commented.

But I can still repeat the fact that we want the QA variant to remain consistent with the external variants.

pzakha: I think it's a good reminder for whoever is reading the code that this is our default stance.

PR #114, opened 2019-07-18
Fix dataset mountpoint permissions post migration for NFSv4 exports

don-brady:
[DLPX-65118](https://jira.delphix.com/browse/DLPX-65118)
Added section to `os-migration` script to fix dataset permissions after a migration from illumos.  For each `domain0` dataset with a default mountpoint, the parents need other-execute for NFSv4 exports.

**Testing**
1. [ab-pre-push --test-upgrade-from 5.3.5.0](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1712/)
2. Manually tested the script that fixes the permissions on a migrated engine
3. Ran the `shellcheck` and `shfmtcheck` checks against the os-migration script

### DIFF
--- a/files/common/var/lib/delphix-platform/os-migration
+++ b/files/common/var/lib/delphix-platform/os-migration
@@ -85,6 +85,27 @@ function perform_migration() {
 		die "Failed chowning /mds"
 
 	#
+	# For each domain0 dataset with a default mountpoint,
+	# the parents need other-execute for NFSv4 exports.
+	#
+	echo "Fixing mountpoint permissions for domain0 datasets"
+	DATASETS=$(zfs list -H -r -t filesystem -o mountpoint domain0 |
+		grep ^/domain0/ | xargs dirname | sort | uniq)
+
+	for DATASET in $DATASETS; do
+		if [[ "$DATASET" == "/domain0" ]]; then
+			continue
+		fi
+		PERMS=$(stat "$DATASET" -c %a 2>/dev/null)
+		RETURN=$?
+		if [[ $RETURN -eq 0 && "$PERMS" == "750" ]]; then
+			echo "chmod o+x $DATASET"
+			chmod o+x "$DATASET" ||
+				die "Failed chmod o+x $DATASET"
+		fi
+	done
+
+	#
 	# DLPX-63949: Postgres must be re-indexed after migration.
 	# See dx_pg_post_start.sh for more info.
 	#


files/common/var/lib/delphix-platform/os-migration
```diff
@@ -84,6 +84,28 @@ function perform_migration() {
...
+		if [[ "$DATASET" == "/domain0" ]]; then
+			continue
+		fi
+		PERMS=$(stat "$DATASET" -c %a 2>/dev/null)
+		RETURN=$?
+		if [[ $RETURN -eq 0 && "$PERMS" == "750" ]]; then
+			echo "Fixing execute permissions: $DATASET" \
+				>>"$MIGRATE_CONFIG_LOG"
+			echo "chmod +x $DATASET"
+			chmod +x "$DATASET" ||
```
sebroy: Do you mean `o+x` here since you say that only "other" needs execute permissions? In the java code that creates new datasets, I think only other gets the execute bit turned on.

don-brady: Good point! I change it to `o+x`

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -84,6 +84,28 @@ function perform_migration() {
...
+		grep ^/domain0/ | xargs dirname | sort | uniq)
+
+	for DATASET in $DATASETS; do
+		if [[ "$DATASET" == "/domain0" ]]; then
+			continue
+		fi
+		PERMS=$(stat "$DATASET" -c %a 2>/dev/null)
+		RETURN=$?
+		if [[ $RETURN -eq 0 && "$PERMS" == "750" ]]; then
+			echo "Fixing execute permissions: $DATASET" \
```
pzakha: Since we are already `echo`ing into the service log on the line below, we can probably remove this.

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -84,6 +84,28 @@ function perform_migration() {
 	chown -R postgres:postgres /mds ||
 		die "Failed chowning /mds"
 
+	#
```
pzakha: nit: we can add something like `echo "Fixing mountpoint permissions for domain0 datsets"` just so it's easier for someone reading the logs to have a grasp of where we are.

files/common/var/lib/delphix-platform/os-migration
don-brady: Good point.  Removing.

files/common/var/lib/delphix-platform/os-migration
don-brady: Added

sebroy:
bors delegate+

PR #115, opened 2019-07-19
rpc.nfsd ignores no-nfs-version parameter at reboot

don-brady:
[DLPX-64900](https://jira.delphix.com/browse/DLPX-64900)

This is a work-around for a bug in `rpc.nfsd(8)` where it ignores the `--no-nfs-version 4.1` parameter after a reboot.  We must prime the kernel NFS `version` so that the input string to disable 4.1 and 4.2 succeeds.

**Testing**
1. [ab-pre-push](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1730/)
2. Confirmed that the generated `dlpx-master` VM had the two new files and that reboots persists the expected nfs version (-2 +3 +4 -4.1 -4.2)
3. Manually ran builds of `shellcheck` and `shfmtcheck` checks against the new script

### DIFF
--- a/files/common/usr/lib/systemd/scripts/nfs-version-init.sh
+++ b/files/common/usr/lib/systemd/scripts/nfs-version-init.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+#
+# Copyright 2019 Delphix
+#
+
+#
+# Work around for kernel failing to parse the 'versions' input string
+# "-4.1 -4.2 -2 +3 +4" passed from rpc.nfsd after a reboot
+#
+# Prime the kernel version after a reboot to include +4 so that the -4.x works
+# Called by '/etc/systemd/system/nfs-config.service'
+#
+# We need this work-around until Ubuntu picks up nfs-utils 2.3.3
+#
+versions=$(cat /proc/fs/nfsd/versions)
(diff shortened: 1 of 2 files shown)


sebroy:
@prakashsurya I was going to mention the same, but Don pointed out that `delphix-platform` already has a similar override for rpc-mountd right now (`etc/systemd/system/nfs-mountd.service.d/override.conf), so this doesn't make things any worse.

prakashsurya:
@sebroy :+1: yea, I just noticed that too. I'm good landing this as-is.

files/common/usr/lib/systemd/scripts/nfs-version-init.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 by Delphix. All rights reserved.
```
prakashsurya: This copyright should be changed to follow the Apache convention (right?), which you used for the override file.

don-brady: yep. I'll update it

sebroy:
bors delegate+

PR #116, opened 2019-07-23
DLPX-61797 zpool create fails sporadically on AWS

jgallag88:
`zpool create` can fail on Xen because a udev rule for cdroms is triggered for the disks in the new pool, causing the device to be busy when `zpool` attempts to open it. We work around this by preventing `udev`'s cdrom rules from matching devices named `xvd*` (xen virtual device). We shouldn't need cdroms on AWS.

### DIFF
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -396,6 +396,31 @@
     - sshd
 
 #
+# On Xen, block devices, including cdroms, are named with the scheme /dev/xvdX.
+# Thus, the udev rules for cdroms are written to match devices with that naming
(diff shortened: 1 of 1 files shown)


jgallag88:
h2. Testing

Ran `zpool create` in a loop on an engine in AWS. Confirmed that the command was no longer sporadically failing, and confirmed using `bpftrace` that `cdrom_id` was no longer being invoked.

ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1756/console

jgallag88:
> With that said, would it better (i.e. more simple) to just remove this file altogether? or disable it some other way? Since we don't care about cdroms for our appliance, I'm just curious if we could fix the bug in a more simple way.

We could disable it by creating a symlink from `/etc/udev/rules.d/60-cdrom_id.rules` to `/dev/null`. The only reason I chose this way instead of that is because I thought this would be less confusing in the off-chance that we do need to use a cdrom for something in the future. I think both approaches are reasonable though.

PR #118, opened 2019-07-25
Use Azure-specific paths when importing domain0

pcd1193182:
On Azure, we want to use the azure-specific links to avoid issues where the by-id links change when the disk is expanded. Because there could be multiple `scsi` directories, we have to build a string that contains a `-d $dir` for each one when running the import command. This change is mutually dependent on the app-gate fix for DLPX-61215

Testing:
Migrated an engine from Illumos to Linux, verified that all disks (include rpool's disk) appear correctly configured. Verified that device removal/addition functioned after migration.

### DIFF
--- a/files/common/var/lib/delphix-platform/os-migration
+++ b/files/common/var/lib/delphix-platform/os-migration
@@ -63,10 +63,22 @@ function perform_migration() {
 		# In most cases, use the devices' by-id links to import
 		# domain0. This ensures persistent vdev names which can
 		# tolerate device reordering. AWS does not provide a by-id
-		# link, so we use simple device names (i.e. /dev/xvda)
+		# link, so we use simple device names (i.e. /dev/xvda). On
+		# Azure, however, we use the Azure-specific links, because the
+		# by-id links can change when the device is resized.
 		#
 		BY_ID_LINK="/dev/disk/by-id"
-		if [[ -d "$BY_ID_LINK" ]]; then
+		AZURE_LINK="/dev/disk/azure"
+		if [[ -d "$AZURE_LINK" ]]; then
+			dir_array=()
+			while read -r dir; do
+				dir_array+=(-d)
+				dir_array+=("$dir")
+			done < <(find "$AZURE_LINK" -maxdepth 1 -name "scsi*")
+			echo "zpool import -f ${dir_array[*]} domain0"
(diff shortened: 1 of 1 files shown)


files/common/var/lib/delphix-platform/os-migration
```diff
@@ -63,10 +63,21 @@ function perform_migration() {
...
 		BY_ID_LINK="/dev/disk/by-id"
-		if [[ -d "$BY_ID_LINK" ]]; then
+		AZURE_LINK="/dev/disk/azure"
+		if [[ -d "$AZURE_LINK" ]]; then
+			dir_array=()
+			while read -r dir; do
+				dir_array+=(-d)
+				dir_array+=("$dir")
+			done < <(find "$AZURE_LINK" -maxdepth 1 -name "scsi*")
+			zpool import -f "${dir_array[@]}" domain0 ||
```
pzakha: Can you log the exact command that we use to import the pool in each case. This will help us diagnose any potential issues.

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -63,10 +63,21 @@ function perform_migration() {
...
 		#
 		BY_ID_LINK="/dev/disk/by-id"
-		if [[ -d "$BY_ID_LINK" ]]; then
+		AZURE_LINK="/dev/disk/azure"
+		if [[ -d "$AZURE_LINK" ]]; then
+			dir_array=()
+			while read -r dir; do
+				dir_array+=(-d)
+				dir_array+=("$dir")
+			done < <(find "$AZURE_LINK" -maxdepth 1 -name "scsi*")
```
grwilson: I'm not sure azure supports more than 1 scsi controller. Have you seen instances where we have anything other than `scsi1`?

pcd1193182: I haven't, but given that it numbers the controller, I figured I would build the logic to handle it, since I wasn't at all sure either way. I can strip that out if you like; I'd also strip a little logic out of the app-gate review that handles it as well.

pcd1193182: I *do* know that the udev rules for azure have logic to create a scsi2 and scsi3 folder, but I don't know if those are ever triggered in practice.

PR #120, opened 2019-08-12
DLPX-65276 Create new dataset that will be shared across versions to store rollback and verify logs

vimalraj-anbarasu:
Problem: During upgrade if we decide to rollback, we need to collect the logs and place it in a location that can be accessed from the older location.

Solution: Create a new dataset that will be shared across versions. Place the rollback logs there.

git ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1920/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ debian/debhelper-build-stamp
 debian/delphix-platform-*
 debian/files
 debian/tmp
+
+#IntelliJ IDEA adds the below files
+delphix-platform.iml
+.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -21,3 +21,5 @@ debian/debhelper-build-stamp
 debian/delphix-platform-*
 debian/files
 debian/tmp
+delphix-platform.iml
```
prakashsurya: Can you add a brief comment above these to describe where these files would come from? I presume it's from some IDE?

vimalraj-anbarasu: Done

vimalraj-anbarasu:
Manually tested, rollback logs are placed in the newly created folder.

PR #126, opened 2019-08-29
LX-1897 migration: make sure sharenfs is not set on any dataset after migration

pzakha:
This makes sure any datasets that were still shared when migrating from Illumos will be unshared. We need this since we'll be switching from NFSv3 to NFSv4 for most datasets and the Illumos sharenfs arguments are not compatible with Linux anyway.

## Testing
migration run: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2030/

### DIFF
--- a/files/common/usr/lib/systemd/scripts/nfs-version-init.sh
+++ b/files/common/usr/lib/systemd/scripts/nfs-version-init.sh
@@ -14,7 +14,7 @@
 #
 versions=$(cat /proc/fs/nfsd/versions)
 if [[ "$versions" == "-2 -3 -4 -4.0 -4.1 -4.2" ]]; then
-	printf "+4\n" >/proc/fs/nfsd/versions
+	printf "+4\\n" >/proc/fs/nfsd/versions
 	echo "Initializing the NFS server version"
 fi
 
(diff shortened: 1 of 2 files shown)


pzakha:
cc: @sebroy @don-brady @grwilson

files/common/usr/lib/systemd/scripts/nfs-version-init.sh
```diff
@@ -14,7 +14,7 @@
 #
 versions=$(cat /proc/fs/nfsd/versions)
 if [[ "$versions" == "-2 -3 -4 -4.0 -4.1 -4.2" ]]; then
-	printf "+4\n" >/proc/fs/nfsd/versions
+	printf "+4\\n" >/proc/fs/nfsd/versions
```
don-brady: Why is this being escaped?  The kernel looks for this return char.

pzakha: It comes down to the same, it's just that shellcheck complains:
```
$ printf "+4\\n" > /tmp/1
$ printf "+4\n" > /tmp/2
$ md5sum /tmp/1 /tmp/2
4a61639514b388ae3d602fe6852aec76  /tmp/1
4a61639514b388ae3d602fe6852aec76  /tmp/2
```

PR #127, opened 2019-09-03
DLPX-62746 ClientAliveInterval needs to be adjusted for Azure DCoA VMs

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2064/)

jgallag88:
Is the plan to set the `client_alive_interval` back to 1800 for the internal variants for Azure as part of another change? I'm thinking we may want to include it in this change so that we don't regress any tests running in Azure.

prakashsurya:
Why would we want to set it to 1800 for our internal variants? We currently don't set it at all for all internal variants except "internal-qa", and I think "internal-qa" should match our external configuration.

If we want "internal-qa" to differ than our external configuration for this specific setting, I'd like us to discuss that with the rest of the team, and be explicit about that decision (since the point of "internal-qa" was to match our external variants as close as is reasonable).

prakashsurya:
I've opened https://github.com/delphix/delphix-platform/pull/128 to remove the timeout for the "internal-qa" variant; this PR should work seamlessly with that one.

PR #128, opened 2019-09-03
Apply SSH timeout only to external variants

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2067/)

jgallag88:
We could keep the 30 min timeout for the internal-qa variant, in which case the internal-qa variant would match the external one on all platforms except for azure. We know that the qa-gate tests work with the 30 minute timeout, and that would be a little closer to our ideal of keeping the qa variant as similar as possible to the eternal one. However, it might be a bit more complicated than necessary, so I'm also OK with this approach.

prakashsurya:
> We know that the qa-gate tests work with the 30 minute timeout

IMO, using 30 minutes here just because it happens to align with the current requirements of CRA and the qa-gate seems fragile and a little arbitrary; since if either of those two requirements changed (e.g. CRA moving to a 3 minute timeout, or qa-gate moving to even longer running commands), our solution might break down. By removing the timeout, our solution should be resilient to changes in the CRA requirements and the qa-gate.

With that said, I'm fine with whatever approach the team thinks is best; I'd be willing to change this PR if folks think using 30 minutes is better.

PR #129, opened 2019-09-03
Add "sdb" to package dependencies

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2072/)

debian/rules
```diff
@@ -53,6 +53,7 @@ DEPENDS += ansible, \
 	   openssl, \
 	   policykit-1, \
 	   rng-tools, \
+	   sdb, \
```
sdimitro: not a big deal but should this be grouped in the same statement with the above packages?
You don't really need `sdb` to `*create a working system*`.

prakashsurya: I can move it into the "for human convenience" group of packages (below), but we're going to use this in the product, so I didn't think that was appropriate.. but there's already packages in that list, that we use in the product (e.g. bcc), so maybe that's a better place to put this?

sdimitro: I'm not really an expert when looking at the code here but yeah that makes more sense to me at least.

PR #130, opened 2019-09-04
DLPX-65853 nfs-kernel-server - failed to init nfsdcltrack database

don-brady:
**Description**
Due to a permissions error, where `'/var/lib/nfs'` is owned by `statd`, the database used by `nfsdcltrack(8)` is never initialized.  The kernel nfs server code invokes `/sbin/nfsdcltrack` to initialize the database, which in turn fails when trying to make the containing `nfsdcltrack` directory.

This database is necessary for the NFSv4 server to track clients and notify them when the server is restarted.  The work around is to create the directory that holds the client tracking database upfront.

**Testing**

Manually tested the work around to confirm that the database is initialized.

ab-pre-push results:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2069/

### DIFF
--- a/files/common/etc/systemd/system/nfs-config.service.d/override.conf
+++ b/files/common/etc/systemd/system/nfs-config.service.d/override.conf
@@ -16,8 +16,9 @@
 
 #
 # Override the nfs-config service post execution parameters. We add an
-# ExecStartPost operation to work around an NFS bug where a no-nfs-version
-# is ignored after a reboot.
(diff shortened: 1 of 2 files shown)


don-brady:
@sebroy  We can certainly take a more direct approach and make the change in`nfs-utils` package.  I opened [nfs-utils PR-2](https://github.com/delphix/nfs-utils/pull/2) as an alternative to this PR.

PR #132, opened 2019-09-05
LX-1775 Post migration cleanup

pzakha:
See also related app-gate change: http://reviews.delphix.com/r/52321/

This performs a cleanup of Illumos file-systems and file permissions that is to be called after the no-rollback point has been reached after a migration upgrade. For debugging purposes we are not destroying the old Illumos datasets right after the migration upgrade, but defer this to the next upgrade.

As such, the cleanup is split into two phases `post-migration` and `final`:
We try to do most of the cleanup in the `post-migration` phase, except for the following things:
 - Destroying `rpool/versions` since it contains the Illumos files from previous upgrades. Also destroy snapshots required by this dataset.
 - Destroying the old `/export/home/` dataset as support has requested to keep it a round for some time.

## Future work
 - Invoke the command for the `final` cleanup (Should be done in DLPX-64438)
 - Add an upgrade-verification check to make sure that the `post-migration` cleanup succeeded.

## Testing
- On a migrated system, manually run `illumos-cleanup post-migration` and `illumos-cleanup final`.
- ab-pre-push for those changes only: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2080/
- migration (including app-gate changes): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2087/
- BlackBox pre_precheckin re-run as it failed above: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/git-blackbox/30846/


pzakha:
Putting this into `/opt/delphix/server/bin` would have been fine as well. I've put it into delphix-platform because this is where most of the other migration-related scripts are.

We do want to be able to call this script manually in case it fails.

pzakha:
cc: @sdimitro @shartse

files/common/usr/bin/illumos-cleanup
```diff
@@ -0,0 +1,185 @@
...
+
+function promote_dataset() {
+	local dataset="$1"
+	local origin
+
+	#
+	# To promote a dataset all the way to the top, zfs promote may need
+	# to be called multiple times if the dataset is a nested clone.
+	# In practice though, since we delete old os versions, the clone
+	# depth shouldn't go deeper than 2 or 3 levels.
```
prakashsurya: why is the limit 2 or 3? what is this derived from?

files/common/usr/bin/illumos-cleanup
```diff
@@ -0,0 +1,185 @@
...
+	#
+	for _ in {1..5}; do
```
prakashsurya: the comment says the limit is 2 or 3, so why do we use a limit of 5 here? should these be consistent with each other?

files/common/usr/bin/illumos-cleanup
```diff
@@ -0,0 +1,185 @@
...
+	#
+	# Some directories still have the old delphix:staff permissions.
+	# We want to convert those to the default Linux root:root.
+	# For /var/delphix, we only need to chown the directory itself
+	# since its contents should have been already chowned.
+	#
+	chown -R root: /var/dlpx-update /var/crash || errors=true
+	chown root: /var/delphix || errors=true
+
+	$errors && die "Some post-migration cleanup tasks have failed."
```
prakashsurya: out of curiosity, why don't we stop at the first error?

files/common/usr/bin/illumos-cleanup
pzakha: That's derived from the way we lay-out our file systems and delete older versions of upgrade images in `rpool/versions`.

For an initial Illumos install, for `/var/delphix` we have: 
`rpool/versions/[os-version]/dlpx/[dlpx-version]/install/var` that is installed, and then cloned into `rpool/versions/[os-version]/dlpx/[dlpx-version]/running/var`.

During upgrade we will have `rpool/versions/[os-version]/dlpx/[old-dlpx-version]/running/var` cloned into `rpool/versions/[os-version]/dlpx/[new-dlpx-version]/running/var`.

Then during migration we will clone `rpool/versions/[os-version]/dlpx/[dlpx-version]/running/var` into the Linux `/var/delphix` dataset.

In my testing, migrating from a fresh Delphix engine, /var/delphix required 2 promotes to get to the top. I'd expect it to require one more if we come from an already upgraded system.

files/common/usr/bin/illumos-cleanup
pzakha: I'll update the comment to mention that we want to stay on the safe side.

files/common/usr/bin/illumos-cleanup
pzakha: The initial idea was for the cleanup to be "best-effort", and not to be a hard requirement of the upgrade, so we'd want to cleanup as much datasets as possible even if there was a failure. Now, since we have some mandatory cleanup tasks (like the `zfs promote` calls), we could either error out right away if anything fails, or only fail if a critical task fails. I'd probably opt for the former since it's simpler.

files/common/usr/bin/illumos-cleanup
```diff
@@ -0,0 +1,185 @@
...
+[[ $# -gt 1 ]] && usage "too many arguments specified"
+[[ $# -eq 0 ]] && usage "too few arguments specified"
+
+[[ "$EUID" -ne 0 ]] && die "must be run as root"
+
+case "$1" in
+post-migration)
+	post_migration_cleanup
+	;;
+final)
```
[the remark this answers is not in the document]
pzakha: I'm not a big fan myself of leaving code that is not being tested. Since the `post-migrate` and `final` tasks are related to each-other, I wanted to make sure that the `final` task would work properly after the `post-migrate` one and to get the logic there.

That said, since I've already written and manually tested the code, I'd rather not to remove it since that would amount to wasted work. I can add a comment here instead pointing out to DLPX-64438 and add a reference to this PR in DLPX-64438.

files/common/usr/bin/illumos-cleanup
```diff
@@ -0,0 +1,185 @@
...
+		return 0
+	fi
+
+	var_delphix=$(zfs list -H -o name /var/delphix) ||
+		die "Failed to get /var/delphix dataset"
+
+	#
+	# Clear the user-property generated by the Illumos installer
+	#
+	zfs inherit org.opensolaris.caiman:install rpool || errors=true
```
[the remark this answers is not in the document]
pzakha: Yeah, I can make it more verbose

pzakha:
@prakashsurya, if you think that this is more appropriate for the app-gate, I can move it there and join it to my other review. 

Regarding `migrationctl status` I'm not planning to add it there for now since this script failing will now result in a failure of the upgrade. Once we add stress options to the app-stack part of the upgrade, I'll adjust `migrationctl` to do the right thing.

Regarding integration tests, I'm not sure how much value would they provide over the end-to-end tests that we are doing in migration testing. Let me know if you had something particular in mind.

prakashsurya:
> if you think that this is more appropriate for the app-gate, I can move it there and join it to my other review.

I do, but I'm OK with this if you think it's better; it's just difficult for me to determine where the conceptual boundaries are for our repositories when it comes to migration.

> Regarding `migrationctl status` ...

I don't know what users should expect when using `migrationctl`, so I guess that's fine.. just wanted to make sure that's not being modified by design, rather than by accident.

> Regarding integration tests ...

Well, my main concern is just that our migration end-to-end tests aren't very comprehensive; e.g. we always start with a "pristine" system of a certain version. With that said, after thinking about it some more, I doubt any mocked up integration test would give me any more confidence in this.

Rather, I think a better way to test this would be to do something akin to "domain restore"; where we'd build out a zpool layout from a "real" system, and verify our clean up works as we expect it to. I think this is similar in concept to how you've been testing migration against support bundles. Perhaps also using longer upgrade chains; e.g. from 5.1 to 5.2 to 5.3 to 6.0.. But I'd consider this out of scope for this PR, and probably more relevant to migration testing as a whole, than this specific change.

So, I'm fine with this as-is.

pzakha:
I'm closing this as I'll move the code into the app-gate.

PR #133, opened 2019-09-05
DLPX-65330 Latest public keys are missing in upgrade and migration images

prashks:
These changes install the correct image signing keys (both `registration` and `upgrade` keys) for the correct/current Delphix Release version being built.
The signing keys version to use is parameterized via `DELPHIX_SIGNATURE_VERSION` which is passed as environment variable via jenkins build job -> linux-pkg -> delphix_platform package build.  Corresponding jenkins changes are: [devops-gate ](http://reviews.delphix.com/r/52325/)

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -22,6 +22,8 @@ KERNEL_TYPE.esx := generic
 KERNEL_TYPE.gcp := gcp
 KERNEL_TYPE.kvm := kvm
 
+DELPHIX_SIGNATURE_TYPES = registration upgrade
+
 #
 # The following meta-package consolidates all the kernel packages required
 # by the Delphix Appliance for a given platform. Note that delphix-kernel
@@ -166,8 +168,9 @@ override_dh_install:
 		>debian/tmp/var/lib/delphix-appliance/platform
 	rm debian/tmp/var/lib/delphix-appliance/platform.in
 
-	./scripts/download-signature-key.sh upgrade 5.3 \
-		>debian/tmp/var/lib/delphix-appliance/key-public.pem.upgrade.5.3
+	for type in $(DELPHIX_SIGNATURE_TYPES) ; do \
+		./scripts/download-signature-key.sh $$type "$(DELPHIX_SIGNATURE_VERSION)" \
+		>"debian/tmp/var/lib/delphix-appliance/key-public.pem.$$type.$(DELPHIX_SIGNATURE_VERSION)" ; done
 
 	dh_install --autodest "debian/tmp/*"
 
--- a/files/common/usr/bin/get-appliance-version
+++ b/files/common/usr/bin/get-appliance-version
@@ -1,0 +1,23 @@
+#!/bin/bash
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -o errexit
+set -o pipefail
+
+zfs get -Hpo value \
+	"com.delphix:current-version" \
+	"$(dirname "$(zfs list -Hpo name /)")" | tr -d '\n'
--- a/files/common/usr/bin/get-property-from-image
+++ b/files/common/usr/bin/get-property-from-image
@@ -80,21 +80,23 @@ UNPACK_DIR=$(mktemp -d -p "$UPDATE_DIR" -t unpack.XXXXXXX)
 [[ -d "$UNPACK_DIR" ]] || die "failed to create unpack directory '$UNPACK_DIR'"
 pushd "$UNPACK_DIR" &>/dev/null || die "'pushd $UNPACK_DIR' failed"
 
-tar -x SHA256SUMS SHA256SUMS.sig.5.3 version.info -f "$UPGRADE_IMAGE_PATH" ||
+DELPHIX_SIGNATURE_VERSION=$(/usr/bin/get-appliance-version | cut -d - -f 1 | cut -d . -f 1-2)
+
+tar -x SHA256SUMS SHA256SUMS.sig."$DELPHIX_SIGNATURE_VERSION" version.info -f "$UPGRADE_IMAGE_PATH" ||
 	die 14 "failed to extract files from upgrade image '$UPGRADE_IMAGE_PATH'"
 
-for file in SHA256SUMS SHA256SUMS.sig.5.3 version.info; do
+for file in SHA256SUMS SHA256SUMS.sig."$DELPHIX_SIGNATURE_VERSION" version.info; do
 	[[ -f "$file" ]] || die 15 "image is corrupt; missing '$file' file"
 done
 
 if [[ -z "$opt_s" ]]; then
 	openssl dgst -sha256 \
-		-verify /var/lib/delphix-appliance/key-public.pem.upgrade.5.3 \
-		-signature SHA256SUMS.sig.5.3 \
+		-verify /var/lib/delphix-appliance/key-public.pem.upgrade."$DELPHIX_SIGNATURE_VERSION" \
+		-signature SHA256SUMS.sig."$DELPHIX_SIGNATURE_VERSION" \
 		SHA256SUMS >/dev/null ||
 		die 16 "image is corrupt; verification of 'SHA256SUMS' file," \
-			"using signature 'SHA256SUMS.sig.5.3'" \
-			"and key 'key-public.pem.upgrade.5.3' failed"
+			"using signature 'SHA256SUMS.sig.$DELPHIX_SIGNATURE_VERSION'" \
+			"and key 'key-public.pem.upgrade.$DELPHIX_SIGNATURE_VERSION' failed"
 fi
 
 awk '$2 == "version.info" { print $0 }' SHA256SUMS |
--- a/files/common/usr/bin/unpack-image
+++ b/files/common/usr/bin/unpack-image
@@ -110,14 +110,16 @@ for file in SHA256SUMS prepare; do
 	[[ -f "$file" ]] || die 15 "image is corrupt; missing '$file' file"
 done
 
+DELPHIX_SIGNATURE_VERSION=$(/usr/bin/get-appliance-version | cut -d - -f 1 | cut -d . -f 1-2)
+
 if [[ -z "$opt_s" ]]; then
 	openssl dgst -sha256 \
-		-verify /var/lib/delphix-appliance/key-public.pem.upgrade.5.3 \
-		-signature SHA256SUMS.sig.5.3 \
+		-verify /var/lib/delphix-appliance/key-public.pem.upgrade."$DELPHIX_SIGNATURE_VERSION" \
+		-signature SHA256SUMS.sig."$DELPHIX_SIGNATURE_VERSION" \
 		SHA256SUMS >/dev/null ||
 		die 16 "image is corrupt; verification of 'SHA256SUMS' file," \
-			"using signature 'SHA256SUMS.sig.5.3'" \
-			"and key 'key-public.pem.upgrade.5.3' failed"
+			"using signature 'SHA256SUMS.sig.$DELPHIX_SIGNATURE_VERSION'" \
+			"and key 'key-public.pem.upgrade.$DELPHIX_SIGNATURE_VERSION' failed"
 fi
 
 sha256sum -c SHA256SUMS >/dev/null ||
--- a/scripts/download-signature-key.sh
+++ b/scripts/download-signature-key.sh
@@ -30,11 +30,6 @@ fi
 set -o errexit
 set -o pipefail
 
-if [[ -z "$1" ]] || [[ -z "$2" ]]; then
-	echo "Must specify key 'type' and 'version'."
-	exit 1
-fi
-
 TYPE="$1"
 VERSION="$2"
 
@@ -46,7 +41,13 @@ VERSION="$2"
 # package will not be correct. When this package is built by our build
 # system and automation, these variables should be available.
 #
-if [[ -n "${DELPHIX_SIGNATURE_TOKEN:-}" ]] && [[ -n "${DELPHIX_SIGNATURE_URL:-}" ]]; then
+if [[ -n "${DELPHIX_SIGNATURE_TOKEN:-}" ]] &&
+	[[ -n "${DELPHIX_SIGNATURE_URL:-}" ]] &&
+	[[ -n "${DELPHIX_SIGNATURE_VERSION:-}" ]]; then
+	if [[ -z "$TYPE" ]] || [[ -z "$VERSION" ]]; then
+		echo "Must specify key 'type' and 'version'."
+		exit 1
+	fi
 	curl -s -S -u "$DELPHIX_SIGNATURE_TOKEN" \
 		"$DELPHIX_SIGNATURE_URL/$TYPE/keyVersion/$VERSION" |
 		jq -Mr .publicKey
--- a/scripts/docker-run.sh
+++ b/scripts/docker-run.sh
@@ -47,6 +47,7 @@ $DOCKER_RUN --rm \
 	--env AWS_SECRET_ACCESS_KEY \
 	--env DELPHIX_SIGNATURE_URL \
 	--env DELPHIX_SIGNATURE_TOKEN \
+	--env DELPHIX_SIGNATURE_VERSION \
 	--volume "$TOP:/opt/delphix-platform" \
 	--workdir "/opt/delphix-platform" \
 	delphix-platform "$@"


debian/rules
```diff
@@ -22,6 +22,19 @@ KERNEL_TYPE.esx := generic
 KERNEL_TYPE.gcp := gcp
 KERNEL_TYPE.kvm := kvm
 
+#
+# The DELPHIX_SIGNATURE_VERSION field refers to the current Delphix
+# Release version being built and only defaults to 6.0 if not set.
+# This is passed as an environment variable via jenkins job(s) that
+# launch the linux-pkg build for this package.
+#
+DELPHIX_SIGNATURE_VERSION := $(or $(DELPHIX_SIGNATURE_VERSION),6.0)
```
prakashsurya: why default to 6.0? this default will soon be incorrect, after we move off the 6.0 release, so I think it might be better to just not use any default value..

debian/rules
```diff
@@ -22,6 +22,19 @@ KERNEL_TYPE.esx := generic
...
+
+#
+# Signing key types
```
prakashsurya: IMO, this comment should be removed, since it isn't providing any more value than what the code itself already provides. Alternatively, we could leave a comment that describes why we have these two key types, and perhaps what their intended usage is.

debian/rules
```diff
@@ -166,8 +179,9 @@ override_dh_install:
 		>debian/tmp/var/lib/delphix-appliance/platform
 	rm debian/tmp/var/lib/delphix-appliance/platform.in
 
-	./scripts/download-signature-key.sh upgrade 5.3 \
-		>debian/tmp/var/lib/delphix-appliance/key-public.pem.upgrade.5.3
+	for keyType in $(SIGN_KEY_TYPES) ; do \
```
prakashsurya: we use [snake case](https://en.wikipedia.org/wiki/Snake_case) already in this file, so it feels inconsistent to use [camel case](https://en.wikipedia.org/wiki/Camel_case) for this one variable. I recommend we change `keyType` to `type` and avoid this issue entirely (IMO, we don't lose anything in terms of readability by doing this).

files/common/usr/bin/unpack-image
```diff
@@ -103,14 +104,34 @@ for file in SHA256SUMS prepare; do
...
+#
+# Provides the version of the latest public key in this engine.
+# Returns "6.0" if the latest key was made at version 6.0 or earlier.
+#
+function get_latest_public_key_version
+{
+	local public_key_dir=$1
+
+	# Public keys are named key-public.pem[.type.version] (for versions above 5.2)
+	(set -o pipefail; ls $public_key_dir/key-public.pem* \
```
prakashsurya: Can you please describe each step of this pipeline? As is, it's hard for me to determine exactly what each step is doing, and why it's doing what it's doing..

This code only has to handle version 6.0 and later (right?), so I'm concerned that if we simply copy what the app-gate code was doing, we may inherit additional complexity that is not needed here.

Further, I believe we'll only ever have a single key installed in `/var/lib/delphix-appliance` at any given time (right?), so I'm wondering if we can more simply just use:
```
$ ls /var/lib/delphix-appliance/key-public.pem.upgrade.*
/var/lib/delphix-appliance/key-public.pem.upgrade.5.3
```
?

debian/rules
prashks: Yeah since 6.0 will be the initial release, thought it would be better to default to that and also was thinking this could avoid any flag day currently.
I'll try removing it.

debian/rules
prashks: Yeah, I'll remove the comment.

debian/rules
prashks: Sounds good, will rename to just `type`.

files/common/usr/bin/unpack-image
prashks: Will get back to you offline on this.
If its handling only version 6.0 and later, agree with you on simplifying it per your suggestion above.

debian/rules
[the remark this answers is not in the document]
prashks: `DELPHIX_SIGNATURE_VERSION` could be missing when :
1)  We push these changes first (without the devops-gate changes) or
2) Building this directly on a bootstrap VM (like `./scripts/docker-run.sh make packages` or via linux-pkg build like `./buildpkg.sh delphix-platform`) for testing etc.

Other than that I don't know of any other workflows where this could be missing.

files/common/usr/bin/unpack-image
```diff
@@ -103,14 +104,34 @@ for file in SHA256SUMS prepare; do
...
+	    | sed 's/^.*key-public\.pem$/6.0/' \
```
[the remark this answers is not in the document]
prashks: Thanks for that insight.
Yes we're only putting keys for 6.0 or later in `/var/lib/delphix-appliance`, so will remove this.

debian/rules
rasantel: By the way, the name `DELPHIX_SIGNATURE_VERSION` might be a little confusing. It represents the version of the public key pair (upgrade, registration) to use hereon in the appliance. It's not necessarily the version of the signature included in the upgrade image where this package is bundled in.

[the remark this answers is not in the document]
rasantel: In that case, it might be safer to require `DELPHIX_SIGNATURE_VERSION` to be defined here and wait for the devops change before pushing this change. The `6.0` default might cause headaches down the road.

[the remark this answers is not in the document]
prakashsurya: I agree with @rasantel; I thought that was already the plan (to push the devops change first), but perhaps not.

For case (2), that's already broken, so I wouldn't worry about it; see here: https://jira.delphix.com/browse/TOOL-8270

prashks: Fixed.

debian/rules
prashks: Fixed.

debian/rules
prashks: Fixed.

files/common/usr/bin/unpack-image
prashks: Fixed.

debian/rules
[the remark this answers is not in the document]
prashks: Fixed.

prashks: For now decided to keep the same name, unless you feel strongly otherwise about this, thanks.

prashks:
@prakashsurya, @rasantel - I have updated this PR with latest changes to address the review comments and our offline discussion. 
Please re-review when you get a chance, thanks.

files/common/usr/bin/get-property-from-image
```diff
@@ -80,21 +80,21 @@ UNPACK_DIR=$(mktemp -d -p "$UPDATE_DIR" -t unpack.XXXXXXX)
 [[ -d "$UNPACK_DIR" ]] || die "failed to create unpack directory '$UNPACK_DIR'"
 pushd "$UNPACK_DIR" &>/dev/null || die "'pushd $UNPACK_DIR' failed"
 
-tar -x SHA256SUMS SHA256SUMS.sig.5.3 version.info -f "$UPGRADE_IMAGE_PATH" ||
+tar -x SHA256SUMS SHA256SUMS.sig.* version.info -f "$UPGRADE_IMAGE_PATH" ||
```
prakashsurya: Have you tested this to work? When I try it, it fails:
```
$ sudo get-property-from-image internal-dev.upgrade.tar VERSION
tar: Pattern matching characters used in file names
tar: Use --wildcards to enable pattern matching, or --no-wildcards to suppress this warning
tar: SHA256SUMS.sig.*: Not found in archive
tar: Exiting with failure status due to previous errors
get-property-from-image: failed to extract files from upgrade image '/export/home/delphix/internal-dev.upgrade.tar'
```

prakashsurya: Adding the `--wildcards` option might work.

If we do that, it'll successfully extract all signatures from the image; this isn't wrong, but we don't necessarily need all of them. We only need the one that matches the version of the private key installed in `/var/lib`. I'm OK using the wildcard and extracting more than is strictly necessary though, if it makes the code more simple.

files/common/usr/bin/get-property-from-image
```diff
@@ -80,21 +80,21 @@ UNPACK_DIR=$(mktemp -d -p "$UPDATE_DIR" -t unpack.XXXXXXX)
...
+for file in SHA256SUMS SHA256SUMS.sig.* version.info; do
 	[[ -f "$file" ]] || die 15 "image is corrupt; missing '$file' file"
 done
 
 if [[ -z "$opt_s" ]]; then
 	openssl dgst -sha256 \
-		-verify /var/lib/delphix-appliance/key-public.pem.upgrade.5.3 \
-		-signature SHA256SUMS.sig.5.3 \
+		-verify /var/lib/delphix-appliance/key-public.pem.upgrade.* \
+		-signature SHA256SUMS.sig.* \
```
prakashsurya: After adding the `--wildcards` option to the line above that I commented on, this fails:
```
dgst: Can only sign or verify one file.
get-property-from-image: image is corrupt; verification of 'SHA256SUMS' file, using signature 'SHA256SUMS.sig.*' and key 'key-public.pem.upgrade.*' failed
```

I believe this is because there's more than one signature contained in the upgrade image. Specifically, right now, there's the 5.3 and 6.0 signatures.

We'll need to ensure we only use the signature version that matches the private key version that's installed in `/var/lib`.. i.e. if the 6.0 key is installed, we want to verify against the 6.0 signature in the upgrade image (and ignore all others).

I think a good way "fix" this might be to parse the appliance version, to extract the currently running/installed version, store that in a variable early in the script, and then use that variable in the various place that we need it. I can discuss this with you later, if you like to pursue this approach.

prakashsurya: Here's an example for getting the version:
```
$ zfs get -Hpo value com.delphix:current-version "$(zfs list -Hpo name /)" | cut -d - -f 1 | cut -d . -f 1-2
6.0
```

If we go this route, it might be best to wrap the `zfs` calls in a new script; e.g. `get-appliance-version` (we have a script like this in the app-gate, but IMO, it probably makes most sense to have this in the delphix-platform respository).

debian/rules
```diff
@@ -22,6 +22,8 @@ KERNEL_TYPE.esx := generic
 KERNEL_TYPE.gcp := gcp
 KERNEL_TYPE.kvm := kvm
 
+SIGN_KEY_TYPES = registration upgrade
```
prakashsurya: To match the `DELPHIX_SIGNATURE_VERSION` variable name, can we change to to something like `DELPHIX_SIGNATURE_TYPES` to be a little more consistent?

debian/rules
```diff
@@ -166,8 +168,9 @@ override_dh_install:
 		>debian/tmp/var/lib/delphix-appliance/platform
 	rm debian/tmp/var/lib/delphix-appliance/platform.in
 
-	./scripts/download-signature-key.sh upgrade 5.3 \
-		>debian/tmp/var/lib/delphix-appliance/key-public.pem.upgrade.5.3
+	for type in $(SIGN_KEY_TYPES) ; do \
+		./scripts/download-signature-key.sh $$type "$(DELPHIX_SIGNATURE_VERSION)" \
```
prakashsurya: When the second parameter to `download-signature-key.sh` is empty, the script will fail.. this causes our PR checks to fail, and likely will also prevent us from using `git-linux-pkg` to build and deploy.

We should ensure building the package still works (i.e. doesn't fail) when building without the `DELPHIX_SIGNATURE_VERSION` variable being available. E.g. I'd be OK augmenting the `download-signature-key.sh` script to return "success" when an empty version is passed in, and simply have it result in nothing being done (like we do when `DELPHIX_SIGNATURE_URL` and `DELPHIX_SIGNATURE_TOKEN` are not specified).

files/common/usr/bin/get-property-from-image
prashks: Thanks @prakashsurya, yes this route seems better, will do.

debian/rules
prashks: Sounds good, will do.

debian/rules
prashks: Yeah earlier had a default for `DELPHIX_SIGNATURE_VERSION` and will make changes to follow similar to when `DELPHIX_SIGNATURE_URL` not specified.

files/common/usr/bin/get-property-from-image
prashks: Thanks, yes for now it makes sense to keep the code simpler, so will revert back to getting the appliance version using the zfs method and change this accordingly.

files/common/usr/bin/get-appliance-version
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 by Delphix. All rights reserved.
```
prakashsurya: This copyright should be updated to use the Apache format, which is different than what the app-gate uses (which I believe this is copied from, right?).

prashks:
**Testing:**
1) [git ab-pre-push with corresponding devops-gate changes]( http://collins.d.delphix.com:28391/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2/)

2) Cloned a DCoA VM from above build, and verified the following:
    - check if required key files are copied/installed by default
       `delphix@ip-10-110-221-2: $ ls -lrt /var/lib/delphix-appliance/`
      `total 4`
      `-rw-r--r-- 1 root root   4 Sep 23 22:04 platform`
      `-rw-r--r-- 1 root root 800 Sep 23 22:04 key-public.pem.upgrade.6.0`
      `-rw-r--r-- 1 root root 800 Sep 23 22:04 key-public.pem.registration.6.0`
      `delphix@ip-10-110-221-2: $`

    - check _get-appliance-version_
      `delphix@ip-10-110-221-2: $ get-appliance-version`
      `6.0.0.0-snapshot.20190923222040628+jenkins-selfservice-psreenivasa-devops-gate-master-appliance-build-master-pre-push-2`
      `delphix@ip-10-110-221-2: $`

    - verify _get-property-from-image_
      `delphix@ip-10-110-221-2: $ sudo get-property-from-image /var/tmp/internal-qa.upgrade.tar VERSION`
      `6.0.0.0-snapshot.20190923222040628+jenkins-selfservice-psreenivasa-devops-gate-master-appliance-build-master-pre-push-2`
      `delphix@ip-10-110-221-2: $`

    - verify _unpack-image_
      `delphix@ip-10-110-221-2:~$ sudo unpack-image /var/dlpx-update/upload/internal-qa.upgrade.tar`
`Progress increment: 19:08:35:071513396+0000, 10, Extracting upgrade image.`
`Progress increment: 19:09:55:668541957+0000, 40, Verifying format.`
`Progress increment: 19:10:23:997351732+0000, 50, Handoff unpack to prepare script.`
`Progress increment: 19:11:50:052506110+0000, 90, Prepare completed successfully.`
`Progress increment: 19:11:50:057885929+0000, 100, Unpacking successful.`
`delphix@ip-10-110-221-2:~$`
 
3) On a bootstrap VM, verify manual build:  
   `ubuntu@ip-10-110-216-82:delphix-platform$ ./scripts/docker-run.sh make packages`

files/common/usr/bin/get-appliance-version
prashks: Thanks, yes, fixed it to follow similar to other scripts (like `get-property-from-image`).

prashks:
**Testing update:**

1.  Repeat [git ab-pre-push](http://collins.d.delphix.com:28391/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5/console)

2. Cloned new VM from above build, and reverified the following:
   * check if required key files are copied/installed by default
      `delphix@ip-10-110-252-169: $ ls -lrt /var/lib/delphix-appliance/`
`total 4`
`-rw-r--r-- 1 root root   4 Sep 25 19:04 platform`
`-rw-r--r-- 1 root root 800 Sep 25 19:04 key-public.pem.upgrade.6.0`
`-rw-r--r-- 1 root root 800 Sep 25 19:04 key-public.pem.registration.6.0`
`delphix@ip-10-110-252-169: $ `

    * check _get-appliance-version_
      `delphix@ip-10-110-252-169: $ get-appliance-version`
`6.0.0.0-snapshot.20190925192019522+jenkins-selfservice-psreenivasa-devops-gate-master-appliance-build-master-pre-push-5`

    * verify _get-property-from-image_ (also with _bash -x_
      `delphix@ip-10-110-252-169: $ sudo get-property-from-image /var/tmp/internal-qa.upgrade.tar VERSION`
`6.0.0.0-snapshot.20190923222040628+jenkins-selfservice-psreenivasa-devops-gate-master-appliance-build-master-pre-push-2`

    * verify _unpack-image_
    `delphix@ip-10-110-252-169: $ sudo unpack-image /var/tmp/internal-qa.upgrade.tar `
`Progress increment: 23:13:44:000279224+0000, 10, Extracting upgrade image.`
`Progress increment: 23:15:57:385714867+0000, 40, Verifying format.`
`Progress increment: 23:16:18:842336802+0000, 50, Handoff unpack to prepare script.`
`Progress increment: 23:18:47:692650504+0000, 90, Prepare completed successfully.`
`Progress increment: 23:18:47:694742955+0000, 100, Unpacking successful.`

3. Manual build on a bootstrap VM
   `ubuntu@ip-10-110-216-82:delphix-platform$ ./scripts/docker-run.sh make packages`

prashks:
Repeat above tests and [git ab-pre-push](http://collins.d.delphix.com:28391/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6/console)

prakashsurya:
bors delegate+

scripts/download-signature-key.sh
```diff
@@ -46,7 +41,13 @@ VERSION="$2"
 # package will not be correct. When this package is built by our build
 # system and automation, these variables should be available.
 #
-if [[ -n "${DELPHIX_SIGNATURE_TOKEN:-}" ]] && [[ -n "${DELPHIX_SIGNATURE_URL:-}" ]]; then
+if [[ -n "${DELPHIX_SIGNATURE_TOKEN:-}" ]] &&
+	[[ -n "${DELPHIX_SIGNATURE_URL:-}" ]] &&
+	[[ -n "${DELPHIX_SIGNATURE_VERSION:-}" ]]; then
```
prakashsurya: It's worth point out, that we pass in `DELPHIX_SIGNATURE_VERSION` as a parameter to this script.

If we're going to check that this variable is set here, it might make sense to more simply remove the second parameter to this script, and instead use `DELPHIX_SIGNATURE_VERSION` instead of `VERSION`.

I don't think it's a big deal either way.. it won't practically change this script.. but it just struck me as odd, that we check for `DELPHIX_SIGNATURE_VERSION` here, but we also pass that variable in by the caller, and then use `VERSION` later, which happens to be derived from `DELPHIX_SIGNATURE_VERSION`.

prashks: Yeah thought about that, but felt like there are 2 workflows eventually invoking this - manual build (like a developer running make packages etc) where `VERSION` sounds relevant and the automated jenkins jobs where `DELPHIX_SIGNATURE_VERSION` becomes relevant.
So prefer to leave this as is for now.

PR #135, opened 2019-09-10
DLPX-66029 migration: network config should be migrated as early as possible

pzakha:
To finalize the migration of the network configuration, in the delphix-migration service we call `migrate_config.py post-upgrade`.

The issue is that we call this after a bunch of other steps. If those other steps fail, then we'd be potentially left with an unconfigured network. If the Delphix Engine was configured with a static IP then this would leave networking unconfigured (although cloud-init would then kick-in and configure it with DHCP if possible). In any case, it would be better to configure networking properly as soon as possible to facilitate debugging in case of issues in the delphix-migration service.

## Testing
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2111
- Palash tested this change with delphix-migration stress options.

### DIFF
--- a/files/common/var/lib/delphix-platform/os-migration
+++ b/files/common/var/lib/delphix-platform/os-migration
@@ -43,6 +43,18 @@ function perform_migration() {
 	__trigger_unset_stress_option \
 		"STRESS_MIGRATION_FAIL_BEGIN_MIGRATION_SERVICE"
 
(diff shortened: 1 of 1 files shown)


pzakha:
cc: @palashgandhi, @sdimitro

PR #136, opened 2019-09-11
delphix-migration should run after local-fs.target

pzakha:
See code comment for details.

Note that we couldn't do this originally because `local-fs.target` used to depend on ZFS mounts as well, but we want to run explicitly before ZFS.
Some of the calls that we do to `systemctl daemon-reload` come from `migrate_config` where we migrate the run-state of services such as masking and NTP.

## Testing
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2132


pzakha:
cc: @sebroy @grwilson @jgallag88

PR #138, opened 2019-09-16
Various improvements to allow migration stress testing in the mgmt stack

pzakha:
We add a new "migration complete" flag file that is set by the mgmt stack
when the migration upgrade is completed. This simplifies the logic in
migrationctl to determine when all migration steps are completed.

migrationctl resume functionality is improved to take into account services
that are auto-restarted on failure. Note that this doesn't apply to the mgmt
stack as auto-restart functionality is now disabled in the mgmt stack for
the duration of migration; this makes stress testing easier and also
improves the debugging experience by avoiding restarting multiple times on
the same failures.

## Dependencies
depends on app-gate: http://reviews.delphix.com/r/52562/
related appliance-build: https://github.com/delphix/appliance-build/pull/362

## Testing
See app-gate review.

### DIFF
--- a/files/common/usr/bin/migrationctl
+++ b/files/common/usr/bin/migrationctl
@@ -18,6 +18,7 @@
 import argparse
 import os
 import sys
+import time
 
 import sh
 
@@ -82,6 +83,14 @@ class Service:
             return False
 
     def start(self):
+        #
+        # If the service has failed and the restart counter reached its
+        # limit we must first reset the counter.
+        #
+        if self.is_failed():
+            if not self._command('reset-failed'):
+                warning('"systemctl reset-failed" failed for service %s'
+                        % self.name)
         started = self._command('start')
         if started and self.state != 'active':
             warning('Service %s start succeeded but state is %s'
@@ -133,6 +142,7 @@ class Service:
 class Migration:
     PERFORM_MIGRATION = '/var/delphix/migration/perform-migration'
     MIGRATION_CHECK_DIR = '/var/delphix/migration/illumos-config'
+    MIGRATION_COMPLETED = '/var/delphix/migration/migration-completed'
 
     def __init__(self):
         Migration.system_check()
@@ -158,22 +168,14 @@ class Migration:
             error("System doesn't come from Illumos")
             sys.exit(1)
 
-    def check_if_reboot_after_migration(self):
-        #
-        # After a successful run of delphix-migration, the service is
-        # disabled, so it will be inactive on reboot.
-        #
-        if self.svc_migration.is_inactive() and \
-                not os.path.exists(Migration.PERFORM_MIGRATION):
-            return True
-        return False
+    @staticmethod
+    def get_mgmt_upgrade_stage():
+        root_dataset = sh.zfs('list', '-H', '-o', 'name', '/').strip()
+        root_container = os.path.dirname(root_dataset)
+        return sh.zfs('get', '-H', '-o', 'value', 'com.delphix:mgmt_upgrade',
+                      root_container).strip()
 
     def status(self):
-        if self.check_if_reboot_after_migration():
-            print('System rebooted after a successful run of %s.'
-                  % self.svc_migration)
-            sys.exit(0)
-
         Service.print_header()
         for service in self.services:
             service.pretty_print_state()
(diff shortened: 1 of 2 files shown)


files/common/usr/bin/migrationctl
```diff
@@ -158,22 +168,14 @@ class Migration:
...
-        #
-        if self.svc_migration.is_inactive() and \
-                not os.path.exists(Migration.PERFORM_MIGRATION):
-            return True
-        return False
+    @staticmethod
+    def get_mgmt_upgrade_stage():
+        root_dataset = sh.zfs('list', '-H', '-o', 'name', '/').strip()
+        root_container = os.path.dirname(root_dataset)
+        return sh.zfs('get', '-H', '-o', 'value', 'com.delphix:mgmt_upgrade',
```
prakashsurya: Are you changing the upgrade logic on linux to start using dataset properties? IIRC, we don't use dataset properties right now, and instead use the `/var/dlpx-update/upgrade.properties` file.. if you're changing this, where can I read more about the motivations for doing so?

pzakha: I'm not changing this, I'm just re-using whatever was already being set. It's possible that it is leftover code from Illumos, but you can check http://reviews.delphix.com/r/52562/ for more details about the existing logic and the changes I've done. If this is obsolete on Linux, I'm open to use `upgrade.properties` instead.

prakashsurya: Perhaps I'm wrong (after looking at some of the app-gate code), and we currently (already) set/use dataset properties for Linux upgrade?

pzakha: Yeah, we already do use them. From my understanding those dataset properties were meant to be used in the rollback code, although I'm not sure that is the case on Linux anymore.

prakashsurya: Looking at code on master, it looks like we use this (only) to determine if we should create the pool checkpoint or not. Can you clarify why querying and using this property here is useful?

If you want to know if the checkpoint exists or not, I think it'd be better to query for the checkpoint itself, rather than this property.

IMO, the app-gate should use a different mechanism for implementing how it creates the checkpoint, and not use a dataset property. So, if that changes, it'd be good to not have to update this code as well.

pzakha: Do you think it would be fine to do this as a follow-up change? I'm not really doing anything beyond leveraging existing logic, and once we have stress testing automated for those new options it will be easier to refactor this code.

prakashsurya: > Do you think it would be fine to do this as a follow-up change?

IMO, what I'm suggesting will make the code more robust, and easier to read and understand. Can you explain why integrating the code as-is is better than what I'm suggesting?

pzakha: I'll talk to you offline to understand better the implications of this change on regular non-migration upgrade.

prakashsurya: Here's my suggestion(s):

1. If you want to show the status of the checkpoint that's created during upgrade, then check that directly using `zpool status`
2. If you want to show if the "post-upgrade cleanup" code has occurred, then I'd suggest we use the `/var/dlpx-update/upgrade.properties` file
3. If you want to capture as much info as possible, to help diagnose a failure situation, then we have the support bundle for this.

I've opened [DLPX-66104](https://jira.delphix.com/browse/DLPX-66104) to track the work to stop using this dataset property on Linux.

pzakha: Those are great suggestions, although we will currently be missing some information. For instance the dataset property can have 5 states and gives us a quick idea of where we are:
`-, createCheckpoint, domainUpgraded, discardCheckpoint, postCleanup`

Once we implement DLPX-66104, some of those states might make it into `upgrade.properties`, so we can update migrationctl to include this information. I've opened https://jira.delphix.com/browse/DLPX-66108 to track this.

Given that I'm not sure if the work above will make it into this release, I'd rather leave the querying of the dataset properties in here for now as it provides convenience.

prakashsurya: > Those are great suggestions, although we will currently be missing some information.

Can you also add this information to support bundles, as I suggested in (3) in my comment above. If this is information we'll need to debug failures, we need to have this information captured in support bundles.

pzakha: Good point, I'll add `migrationctl status` output to the support bundle. Created https://jira.delphix.com/browse/DLPX-66110 to track this.

prakashsurya: Where appropriate, please also ensure we capture equivalent information for non-migration upgrade cases, too; e.g. where we can use a single solution/command/etc. for both, migration and non-migration, I think we should.

pzakha:
cc: @sebroy @sdimitro

PR #139, opened 2019-09-17
DLPX-65501 uploading same upgrade image twice has "unexpected" error

vimalraj-anbarasu:
Instead of tags unpack script will accept different exit codes that will be translated in app-stack.
Corresponding app-gate change: http://reviews.delphix.com/r/52590

ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2191/

### DIFF
--- a/files/common/usr/bin/get-property-from-image
+++ b/files/common/usr/bin/get-property-from-image
@@ -20,8 +20,15 @@ set -o pipefail
 UPDATE_DIR=${UPDATE_DIR:-/var/dlpx-update}
 
 function die() {
+	exit_code=$1
+	# Use first argument as exit code, if provided.
+	if [[ -z "${exit_code//[0-9]/}" ]]; then
+		shift
+	else
+		exit_code=1
+	fi
 	echo "$(basename "$0"): $*" >&2
-	exit 1
+	exit ${exit_code}
 }
 
 function usage() {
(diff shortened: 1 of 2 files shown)


sanbhara:
Looks good. I suppose if prepare were to send an unrecognized new exit code, we will just default to exit code 1 in the app stack which is fine.

files/common/usr/bin/get-property-from-image
```diff
@@ -53,49 +60,49 @@ shift $((OPTIND - 1))
 [[ $# -gt 2 ]] && usage "too many arguments specified"
 [[ $# -lt 2 ]] && usage "too few arguments specified"
 
-[[ "$EUID" -ne 0 ]] && die "must be run as root"
-[[ -d "$UPDATE_DIR" ]] || die "$UPDATE_DIR does not exist"
+[[ "$EUID" -ne 0 ]] && die 1 "must be run as root"
```
prakashsurya: for all cases where `die 1 ...` is used, should we just use `die ...` since `1` is the default when not specified?

vimalraj-anbarasu:
> out of curiousity, why do you start at 11, for all the non-default error codes?

Exit codes 0, 1 and 2 are reserved. I thought will reserve exit codes till 10 for generic internal errors from our scripts.

files/common/usr/bin/get-property-from-image
vimalraj-anbarasu: We can, I don't have a preference. Will update the scripts.

PR #141, opened 2019-09-19
DLPX-36241 Version List Not Updated After Uploading Upgrade Version

AdityaPrasad-dlpx:
**Solution**: 
Check if the management stack is active. If so, pass the `system version` command along to the CLI to trigger an upgrade version list refresh. This will cause the UI to update the version list after an upgrade image is uploaded using the command line tool `unpack-image`.

### DIFF
--- a/files/common/usr/bin/unpack-image
+++ b/files/common/usr/bin/unpack-image
@@ -133,4 +133,8 @@ return=$?
 
 report_progress_inc 100 "Unpacking successful."
 
+if [[ $(systemctl is-active delphix-mgmt.service) == "active" ]]; then
+	su sysadmin -c "system version" >/dev/null
+fi
+
 exit 0


files/common/usr/bin/unpack-image
```diff
@@ -133,4 +133,8 @@ return=$?
 
 report_progress_inc 100 "Unpacking successful."
 
+if [[ $(systemctl is-active delphix-mgmt.service) == "active" ]]; then
+	su sysadmin -c "system version" >/dev/null
```
prakashsurya: Why would we want to do this here, as opposed to solving the problem via a change to the app-gate (or qa-gate)?

At first glance, I feel like this is the wrong place to be doing this. As is, this script has no knowledge of the app-gate, and I think it should remain that way. Instead, I'd recommend any users of this script handle updating the app-gate version list on their own, rather than having this script be responsible for that.

PR #142, opened 2019-09-24
DLPX-66005 Unquiesced VDBs cause domain0 to fail to mount due to a bug in zfs inherit sharenfs

pzakha:
zfs inherit sharenfs automatically mounts each dataset which has
sharenfs set. However the parent datasets will not be mounted if
they do not have sharenfs set, which will result in the creation
of mount directories in the wrong file systems and will later
cause zfs mount -a to fail with "directory not empty". As a
work around to that bug, we make sure to call zfs mount -a first.

## Testing
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2221/
- manually tested migration with VDBs that failed to quiesce and verified that migration completed properly.


pzakha:
cc: @sebroy @sdimitro @shartse

PR #153, opened 2019-10-26
DLPX-65491 Invalid argument when mounting ZFS filesystem

grwilson:


grwilson:
[ab-pre-push](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2376/)

files/common/lib/sysctl.d/50-override.conf
```diff
@@ -71,3 +71,13 @@ net.ipv4.tcp_rmem = 4096 4194304 16777216
 # allows the script that generates the hprof dump to run.
 #
 vm.overcommit_memory = 1
+
+#
+# By default all mounts are "shared" with other mount peer groups.
+# As a results any new mount will propagate the mount event to their peer
+# groups. This can result in inflating the number of mounts for that
+# mount namespace resulting in it hitting the mount max value prematurely.
+# To avoid this, we increase the mount-max value to 3 times the default
```
pzakha: Can you elaborate how you picked the number 3? Doesn't this just mean that we push the limit a bit further, and that we'd hit it if we spin up more VDBs?

grwilson: For us to hit this would require us to have about 150K filesystems mounted that were then part of the `systemd-resolved` mount namespace. Even 50K filesystem is unlikely in our customer base but the scalability servers have seen numbers this high and that's where we were hitting this bug.

pzakha: I'm fine with that, we should probably just mention that 150k file systems limit in this comment.

Setting the mounts as private would probably indeed cause issues during upgrade-verify since we are bind mounting domain0 in the container, right?

Also, when running upgrade-verify, would that result in even more mounts in the `systemd-resolved` namespace?

grwilson: I tested that the upgrade container will add another factor to our already inflated count. So with that in mind, to support 100,000 filesystems we need to set the max to 3 times that value. If we start adding additional containers, we will have to adjust this.

PR #155, opened 2019-10-29
DLPX-66227 Disk I/O scheduler should be `noop` rather than default `cfq`

tonynguien:
Disk should have `noop` I/O scheduler for optimal performance since ZFS will schedule I/Os.

### DIFF
--- a/files/common/etc/default/grub.d/override.cfg
+++ b/files/common/etc/default/grub.d/override.cfg
@@ -36,6 +36,12 @@ GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT mitigations=off"
 GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT ipv6.disable=1"
 
 #
+# On Delphix appliance, disks should have 'noop' I/O scheduler since they
+# would be under ZFS control.
+#
+GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT elevator=noop"
+
+#
 # Increase the amount of memory reserved for the crash kernel.
 #
 # Empirically, it seems we need about 512M to get it to boot so


files/common/etc/default/grub.d/override.cfg
```diff
@@ -35,6 +35,12 @@ GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT mitigations=off"
 #
 GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT ipv6.disable=1"
 
+#
+# On Delphix appliance, disks should have 'noop' I/O scheduler since they
+# would be under ZFS control.
+#
+GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT elevator=noop"
```
prakashsurya: have we considered the potential affects this may have on the IO report card? we run that on "raw" disks, right? so, there's no ZFS IO scheduling to take advantage of there.. I wonder if we would want to use "noop" in that case?

tonynguien: Yes, we get better IO Report Card results with 'noop' scheduler. I have results in DLPX-66227 JIRA case.

prakashsurya: Ah, sorry.. I see that was already considered in the bug. Thanks.

tonynguien:
> LGTM

Thanks!

tonynguien:
> After you push this, can you also open an appliance-build review for setting this during the first boot after migration?
> The new option should be added here: https://github.com/delphix/appliance-build/blob/master/live-build/misc/migration-scripts/dx_apply#L285 (you should test the appliance-build change by running `git ab-pre-push --test-upgrade-from 5.3.6.0`)

Thanks for the note. I filed DLPX-66946 for that task.

sebroy:
bors delegate+

PR #159, opened 2019-11-13
DLPX-67281 Network configuration not migrated because of multiple netplan files

sdimitro:
# Commit Description

The root-cause of this issue is that the service that generates the
default netplan file for cloud-init (named cloud-init-local) can run
at the same time as delphix-migration which writes our own custom
netplan file on-disk (and potentially deletes the default one if the
timing is right). Unfortunately, timing is not always right and due
to the above raace between the two services we end up with two
netplan files that can have conflicting info.

This change ensures that the migration service runs after cloud-init-local
so the default netplan file is always generated before ouyr custom one
takes its place.

Note again that this is a migration-only issue that can happen on
first boot. We disable the cloud-init-local service from regenerating
its netplan file for subsequent boots.

# Testing
 
(pending test results)


pzakha:
That meant to be `bors delegate+`, my mistake.

sdimitro:
Discussed the event of `s/bors r+/bors degelate+` with Pavel - I had done some preliminary testing but since this is a race I wanted to get a few more runs in to be sure that we don't hit this anymore. I'll update this PR with the results and in the first sight of a failure we can always revert

sdimitro:
Here are a couple of test runs were the tests that were failing are passing:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/2688/console
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/2689/console

PR #160, opened 2019-11-13
 DLPX-67375 [Backport of Issue DLPX-67281 to 6.0.0.0] Network configuration not migrated because of multiple netplan files

sdimitro:
The root-cause of this issue is that the service that generates the
default netplan file for cloud-init (named cloud-init-local) can run
at the same time as delphix-migration which writes our own custom
netplan file on-disk (and potentially deletes the default one if the
timing is right). Unfortunately, timing is not always right and due
to the above raace between the two services we end up with two
netplan files that can have conflicting info.

This change ensures that the migration service runs after cloud-init-local
so the default netplan file is always generated before ouyr custom one
takes its place.

Note again that this is a migration-only issue that can happen on
first boot. We disable the cloud-init-local service from regenerating
its netplan file for subsequent boots.


sdimitro:
PR for original issue can be found here -> https://github.com/delphix/delphix-platform/pull/159

files/common/lib/systemd/system/delphix-migration.service
```diff
@@ -37,11 +37,22 @@
...
+# netplan files that can have conflicting information for networkd.
+# To avoid this scenario, we make sure that this service runs AFTER
+# cloud-init-local.
+#
 [Unit]
 Description=Delphix OS Migration Service
 PartOf=delphix.target
 DefaultDependencies=no
 After=systemd-udev-settle.service
+After=cloud-init-local.service
```
grwilson: Out of curiosity, do we have to worry about running before any other cloud-init services? In other words, does this also need a `Before=cloud-init.service`?

sdimitro: That's a great question. In term of anything that we manage by netplan there should be nothing to worry about. Outside of that though, one could bring up the question of `resolv.conf`. Looking at the logs in our machines and the source code of cloud-init it is not obvious to me which service handles the creation of that file by default. If it is indeed `cloud-init` and not `cloud-init-local` then we should probably do what you suggest. I can open an issue about this so it is tracked somewhere as this is a backport and I wouldn't like to introduce any differences between this and its original commit on `master`.

sdimitro: Issued https://jira.delphix.com/browse/DLPX-67388

PR #161, opened 2019-11-14
DLPX-67394 Increase postgres service timeout during migration (Part 2 of 2)

pzakha:
Part 1 of 2: http://reviews.delphix.com/r/54101

See JIRA for description.

Caveat:
For the internal-dev variant we deploy other override.conf files under `/etc/systemd/...`, so the files for the same service under `/run/systemd/...` are ignored by systemd. An alternative would be to edit `/etc/systemd` files instead, but the logic would be more complex and prone to failure, so I've opted for this instead.

## Testing
Testing this change only:
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2490/

Testing both parts together: see http://reviews.delphix.com/r/54101


sebroy:
bors delegate+

PR #164, opened 2019-11-18
Add performance-diagnostics package

jgallag88:


jgallag88:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2510/

debian/rules
```diff
@@ -137,6 +137,7 @@ DEPENDS += bcc-tools, \
 	   memstat, \
 	   ncdu, \
 	   pciutils, \
+	   performance-diagnostics, \
```
pzakha: Are any of these scripts going to be called by code from the app-stack? If so, then we should move this as a requirement of the virtualization package.

jgallag88: Yes, eventually, so the virtualization package will need to add a dependency on `performance-diagnostics`. However, some of the scripts could be useful in debugging performance issues in other contexts, such as on dc1, so I figured it made sense to have this in the base image. I'm open to other thoughts though.

sebroy: I think it should be here so that it can be included on DCoL. Brad will add it as a dependency on virtualization as part of his change to re-work dlpx-app-gate to use these scripts.

pzakha: Sounds good!

We can add it to delphix-virtualization separately. I think `delphix-virtualization` is missing other packages in this list e.g. `bpftrace` and maybe a few other

PR #170, opened 2019-12-09
Disable Ubuntu motd banner and provide a Delphix-specific one.

sebroy:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2577/

Before:
```
guild login: sysadmin
Password:
Welcome to Ubuntu 18.04.3 LTS (GNU/Linux 5.0.0-36-generic x86_64)
 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage
 * Overheard at KubeCon: "microk8s.status just blew my mind".
     https://microk8s.io/docs/commands#microk8s.status
 * Canonical Livepatch is available for installation.
   - Reduce system reboots and improve kernel security. Activate at:
     https://ubuntu.com/livepatch
The programs included with the Ubuntu system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.
Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.
guild.dc1>
```
After:
```
guild login: sysadmin
Password:
Last login: Mon Dec  9 16:01:40 UTC 2019 on ttyS0
Welcome to Delphix 6.1.0.0-snapshot.20191127042627846
guild.dc1>
```

### DIFF
--- a/files/common/etc/update-motd.d/00-delphix
+++ b/files/common/etc/update-motd.d/00-delphix
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 full_version=$(/usr/bin/get-appliance-version)
-version=$(echo $full_version | awk -F'-' '{print $1}')
-build=$(echo $full_version | sed 's/.*-\(.*\)+.*/\1/')
+version=$(echo "$full_version" | awk -F'-' '{print $1}')
+build=$(echo "$full_version" | sed 's/.*-\(.*\)+.*/\1/')
 
 printf "Welcome to Delphix %s (%s)\n" "$version" "$build"
--- a/files/common/etc/update-motd.d/00-delphix
+++ b/files/common/etc/update-motd.d/00-delphix
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+full_version=$(/usr/bin/get-appliance-version)
+version=$(echo $full_version | awk -F'-' '{print $1}')
+build=$(echo $full_version | sed 's/.*-\(.*\)+.*/\1/')
+
+printf "Welcome to Delphix %s (%s)\n" "$version" "$build"
--- a/files/common/etc/update-motd.d/00-delphix
+++ b/files/common/etc/update-motd.d/00-delphix
@@ -1,7 +1,4 @@
 #!/bin/sh
 
-full_version=$(/usr/bin/get-appliance-version)
-version=$(echo "$full_version" | awk -F'-' '{print $1}')
-build=$(echo "$full_version" | sed 's/.*-\(.*\)+.*/\1/')
-
-printf "Welcome to Delphix %s (%s)\n" "$version" "$build"
+version=$(/usr/bin/get-appliance-version | awk -F'+' '{print $1}')
+printf "Welcome to Delphix %s\n" "$version"
--- a/files/common/etc/update-motd.d/00-delphix
+++ b/files/common/etc/update-motd.d/00-delphix
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-version=$(/usr/bin/get-appliance-version | awk -F'+' '{print $1}')
+version=$(get-appliance-version | awk -F'+' '{print $1}')
 printf "Welcome to Delphix %s\n" "$version"
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -478,3 +478,16 @@
     path: /etc/default/kdump-tools
     regexp: '^#?MAKEDUMP_ARGS='
     line: 'MAKEDUMP_ARGS="-c -d 31 --message-level 22 --private-page-filter 0x2F5ABDF11ECAC4E"'
+
+#
+# Disable all motd scripts except for those provided by Delphix
+#
+- find:
+    paths: /etc/update-motd.d
+    excludes: '*-delphix*'
+  register: motd_files
+
+- file:
+    path: "{{ item.path }}"
+    mode: 0644
+  with_items: '{{ motd_files.files }}'
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -480,9 +480,9 @@
     line: 'MAKEDUMP_ARGS="-c -d 31 --message-level 22 --private-page-filter 0x2F5ABDF11ECAC4E"'
 
 #
-# Disable all motd scripts except for those provided by Delphix. We start by
-# removing executable permissions on every script in /etc/update-motd.d except
-# those that have "-delphix" in their filename.
+# Disable all motd scripts except for those provided by Delphix by removing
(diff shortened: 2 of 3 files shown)


files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -478,3 +478,16 @@
...
+
+#
+# Disable all motd scripts except for those provided by Delphix
+#
+- find:
+    paths: /etc/update-motd.d
+    excludes: '*-delphix*'
+  register: motd_files
+
+- file:
```
prakashsurya: Can you comment on why it's most appropriate to do this configuration here, in the delphix-platform package, vs. appliance-build? Should we do it in both places?

I ask, because we remove the `/etc/issue` file currently in the appliance-build configuration of the initial VM.. Perhaps we should remove that configuration from appliance-build, and move that to here (for consistency's sake)?

With the configuration here, it'll be applied to the VM image such that the system is configured correctly on first boot (prior to the platform service running), right?

files/common/etc/update-motd.d/00-delphix
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+full_version=$(/usr/bin/get-appliance-version)
```
prakashsurya: why use `/usr/bin` explicitly, vs. relying on using the version in the `PATH`?

files/common/etc/update-motd.d/00-delphix
```diff
@@ -0,0 +1,7 @@
...
+version=$(echo "$full_version" | awk -F'-' '{print $1}')
+build=$(echo "$full_version" | sed 's/.*-\(.*\)+.*/\1/')
```
pzakha: I think we should display anything before the `+` sign as the version, so your example would become:
```
Welcome to Delphix 6.1.0.0-snapshot.20191127042627846
```
Most release builds will have an empty "build" field.

I've been using [this doc](https://docs.google.com/document/d/1K9-7TY0LdLtdT-MilECptO_6taH3edrz5pjH--LBy3s/edit#heading=h.pc1842v34c8x) as reference.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
pzakha: I think we should try to do most of the logic in `delphix-platform` rather than `appliance-build` as most of the appliance-build customizations are not carried over automatically on (not-in-place) upgrade.

sebroy: My thinking and understanding was that our ideal was to have `appliance-build` remain (where possible) a generic appliance build mechanism, while Delphix use-case specific configuration should be done in `delphix-platform`, or in specific packages that are specific to that use case (e.g. virtualization or masking). My understanding was also that the `delphix-platform` ansible role does get applied to the VM during the `appliance-build` process (prior to creating the initial disk image).

I've deployed a DCoA VM based on my `ab-pre-push` job above, and the changes were there on first boot.

files/common/etc/update-motd.d/00-delphix
sebroy: You're right, thanks for pointing that out. I'll take the suggestion and simply print everything before the `+` sign as-is.

files/common/etc/update-motd.d/00-delphix
sebroy: Just a habit from having had to fix broken scripts that accidentally use incorrect versions of executables because they assumed someone had set PATH correctly for them.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
sebroy: I neglected to reply to the point about being consistent with the handling of other files such as `/etc/issue`. I think in that case the file gets removed by the build so that other packages can deliver that file without conflict perhaps?

prakashsurya: w.r.t. `/etc/issue`.. I was asking more about if we should remove the removal that's being done via appliance-build, and move it into here, so all of the logic affecting the login message is contained in a consistent place.

files/common/etc/update-motd.d/00-delphix
prakashsurya: IMO, it'd be better to use what's in the PATH; which would also be consistent with how you're calling awk/sed.

sebroy:
Update on this PR: During testing, I noticed that the `motd-news.service` was in failed state. In its journal was:
```
Dec 10 19:27:08 guild.dc1 systemd[1]: Starting Message of the Day...
Dec 10 19:27:08 guild.dc1 systemd[13914]: motd-news.service: Failed to execute command: Permission denied
Dec 10 19:27:08 guild.dc1 systemd[13914]: motd-news.service: Failed at step EXEC spawning /etc/update-motd.d/50-motd-new
```
This lead me down the rabit hole of motd-news.service, which fetches dynamic content from the Internet every 12 hours via a motd-news.timer systemd unit. I'm going to be updating this PR to disable this, as not only does this fail if motd scripts are disabled, but we don't want engines reaching out to the public Internet for dynamic content from Ubuntu.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
grwilson: We will probably need to move `/etc/issue` to `delphix-platform` when we add the IP address to the login prompt.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -478,3 +478,39 @@
...
+# 3. The service explicitly runs the /etc/update-motd.d/50-motd-news script
+#    and fails if that script isn't executable (we disable the executable
+#    bit on it above).
+#
+# As such, the service needs to be disable/masked, which involves both
+# the service unit along with the timer unit that is tasked with running
+# it every 12 hours.
+#
+- command: systemctl mask motd-news.service
+- command: systemctl mask motd-news.timer
```
prakashsurya: I think it'd be better to disable these via the packaging hooks, like we do for other services.

prakashsurya: see here: https://github.com/delphix/delphix-platform/blob/master/debian/postinst

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
sebroy: That's true. I'll leave that logic alone and we can address it then.

PR #179, opened 2020-01-10
Add more systemd-related debug symbol packages 

jgallag88:
This adds debug symbols for several libraries used by systemd and dbus. Parts of these libraries provide core bits of infrastructure used by systemd and dbus, so having debug info is important for being able to debug those programs effectively.

ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2677/

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -61,7 +61,12 @@ DEPENDS += ansible, \
 
 # Debugging symbols for packages pulled in by the the above dependencies
 DEPENDS += systemd-dbgsym, \
+	   libsystemd0-dbgsym, \
+	   libblkid1-dbgsym, \
+	   libmount1-dbgsym, \
+	   libuuid1-dbgsym, \
 	   dbus-dbgsym, \
+	   libdbus-1-3-dbgsym, \
 	   openssh-server-dbgsym, \
 	   openssh-client-dbgsym, \
 	   coreutils-dbgsym, \


sdimitro:
Not sure if it worths it or the space consumption would be too much but here are some of the most common dbg symbol packages for libraries that I've seen when I open systemd with `drgn`:
```
libblkid1-dbgsym
libseccomp2-dbgsym
libselinux1-dbgsym
libpam0g-dbgsym
libaudit1-dbgsym
libkmod2-dbgsym
libapparmor1-dbgsym
libacl1-dbgsym
libcryptsetup12-dbgsym
libcap2-dbgsym
libgcrypt20-dbgsym
libip4tc0-dbgsym
libidn11-dbgsym
liblzma5-dbgsym
liblz4-1-dbg
libpcre3-dbg
libuuid1-dbgsym
libcap-ng0-dbgsym
libattr1-dbgsym
libdevmapper1.02.1-dbgsym
libargon2-0-dbgsym
libjson-c3-dbgsym
libgpg-error0-dbgsym
libudev1-dbgsym
```

sdimitro:
`libuuid` and `libblkid` must be very common among userland utils because I've even seen them in `ztest`

jgallag88:
One of the reasons that I was hesitant to just add debug symbols for everything, apart from size, was that there isn't a good way to say 'install the dbgsym for package X and all of its dependencies'. As far as I can see, you have to manually list the -dbgsym package for each dependency, which means that if the dependencies change, you could end up pulling in packages you don't actually need, or trying to pull in packages that don't even exist any more.

That said, I don't think the bar deciding to add a package should be that high. If we suspect something should be useful, we may as well just add it, so I've added dbgsym packages for `libuuid` and `libblkid`.

sdimitro:
The above makes sense to me. Thanks for looking into it again!

pzakha:
Note that ab-pre-push didn't succeed, probably due to infrastructure issues.

jgallag88:
I cancelled that run and started this one after modifying the PR: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2678/

This should be ready to land. If someone with the appropriate permissions wants to merge this that would be great.

PR #182, opened 2020-01-13
DLPX-68014 iSCSI service interruption during deferred upgrade

pzakha:
## Pull request checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Build was run locally and any changes were pushed

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: DLPX-68014
The current behaviour is that the rtslib-fb-targetctl service restarts during a deferred upgrade, causing iSCSI-based VDBs to be inaccessible for a certain time. 

## What is the new behavior?

The new behaviour is that rtslib-fb-targetctl is not restarted on upgrade anymore.

This is achieved with 2 changes:
1. We do not re-enable the delphix-migration service when we upgrade the delphix-platform package.
2. We remove delphix-migration as a part of delphix.target.
Either one of those changes should be theoretically sufficient to prevent this bug, but we are doing them both since that is the correct behaviour for this service.

- [ ] Yes
- [x] No

## Testing
- manually upgraded delphix-platform and restarted `delphix.target`. Verified that delphix-migration.service was not re-enabled and that rtslib-fb-targetctl.service was not restarted.
- ab-pre-push (testing upgrade): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2687/
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2686/

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -49,9 +49,16 @@ configure)
 
 	systemctl enable auditd.service
 	systemctl enable delphix.target
-	systemctl enable delphix-migration.service
 	systemctl enable delphix-platform.service
 	systemctl enable systemd-networkd.service
+
+	#
+	# Only enable the delphix-migration service on initial install, not
+	# on upgrade.
+	#
+	if [[ -z "$2" ]]; then
+		systemctl enable delphix-migration.service
+	fi
 
 	if ! id -u postgres >/dev/null; then
 		# When installing postgres, a postgres user is created unless it
--- a/files/common/lib/systemd/system/delphix-migration.service
+++ b/files/common/lib/systemd/system/delphix-migration.service
@@ -49,7 +49,6 @@
 #
 [Unit]
 Description=Delphix OS Migration Service
-PartOf=delphix.target
 DefaultDependencies=no
 After=systemd-udev-settle.service
 After=cloud-init-local.service
@@ -88,7 +87,7 @@ RemainAfterExit=yes
 ExecStartPost=/bin/systemctl disable --no-reload delphix-migration.service
 
 [Install]
-WantedBy=delphix.target
+WantedBy=default.target
 
 #
 # By Setting RequiredBy dependencies in the Install section, this enforces


debian/postinst
```diff
@@ -49,10 +49,17 @@ configure)
...
-	systemctl enable delphix-migration.service
 	systemctl enable delphix-platform.service
 	systemctl enable systemd-networkd.service
 
+	#
+	# Only enable the delphix-migration service on initial install, not
+	# on upgrade.
+	#
+	if [[ -z "$2" ]]; then
+		systemctl enable delphix-migration.service
```
prakashsurya: Did you consider removing this enable altogether, and then having a different change to appliance-build.. to enable this service via the appliance-build ansible logic or the migration image creation logic?

I don't think we ever want to enable the migration service, except when performing a migration, so it might be better to only enable it when building the migration artifact, since that will be a migration specific step.

For example, I don't think we should have migration enabled on initial boot, for a DE that started on 6.0, and thus was never migrated. As far as I understand this change, this way, we'll still start the migration service for initial 6.0 installs.

pzakha: Yeah, right now delphix-migration always starts on first boot and then disables itself after running successfully. Having the migration service enabled only for the migration artifact would probably have been cleaner, that said we'll be already shipping 6.0.0.0 with the current behaviour, so if we were to change this we'd then have differing behaviours depending on the version.

Looking at appliance-build, we currently use the same `binary` directory for generating all the artifacts, so we'd need to make a separate copy of the directory (in a similar way that `90-raw-disk-image.binary` does it) with the change to enable the `delphix-migration service` there. That should work but that would require some extra testing.

Do you think we should dig into this more?

prakashsurya: > so if we were to change this we'd then have differing behaviours depending on the version.

I don't think we should limit improvement in new versions, simply because old versions wouldn't get that same improvement.

> Do you think we should dig into this more?

It's up to you; if this works, I think it's fine. I'll approve as-is.

PR #180, opened 2020-01-13
DLPX-68017 generic images should use NoCloud

grwilson:

PR #183, opened 2020-01-13
DLPX-68021 cloud-init should exclude vendor data

grwilson:


grwilson:
This has been testing in combination with https://github.com/delphix/delphix-platform/pull/180

PR #199, opened 2020-02-03
DLPX-68141 [Backport of Issue DLPX-68025 to 6.0.1.0] migration: clearing sharenfs on domain0 datasets takes a long time

pzakha:
Instead of calling "zfs inherit" recursively on domain0, which tries to
clear sharenfs on every single dataset under domain0, the new approach is
to first identify which datasets have sharenfs set and then clear them one
by one. Datasets that are not mounted are ignored since clearing sharenfs
on unmounted datasets can end-up mounting them without first mounting
their parents, which will automatically create directories for their
mountpoints and later prevent the parent datasets from being mounted
due to their mount directories being not empty.

## Pull request checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Build was run locally and any changes were pushed

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

## Other information

Note that I'm not opening a PR on master because we are not allowing migrations to 6.1 (Also my testing on master was failing because of DLPX-68071).

## Testing

- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2811/
- environment resiliency sanity test: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/3284/
- manually tested the script with some datasets unmounted

### DIFF
--- a/files/common/var/lib/delphix-platform/clear-sharenfs
+++ b/files/common/var/lib/delphix-platform/clear-sharenfs
@@ -1,0 +1,65 @@
+#!/usr/bin/python3
+
+import subprocess
+
+lines = subprocess.check_output(
+    "zfs list -t filesystem -r -H -o name,sharenfs,mounted domain0",
+    shell=True).decode('utf-8').splitlines()
+datasets = [line.split('\t') for line in lines]
+#
+# Filter out datasets with sharenfs == "off" and sharenfs == "-"
+#
+datasets = [ds for ds in datasets if ds[1] not in ('off', '-')]
+
+#
+# List all datasets that have sharenfs property source set to "local".
+# Those are the datasets for which we explicitly set the sharenfs property.
+# By default, datasets have sharenfs property source set to "default", meaning
+# that they inherit whatever value is set on their parent. Therefore we
+# filter out all datasets that do not have sharenfs source set to local since
+# their sharenfs setting will be updated automatically.
+#
+sharenfs_local = subprocess.check_output(
+    "zfs get -r -t filesystem -o name -Hs local sharenfs domain0",
+    shell=True).decode('utf-8').splitlines()
+datasets = [ds for ds in datasets if ds[0] in sharenfs_local]
+
+#
+# We only clear sharenfs for datasets that are mounted as in some cases
+# zfs can mount datasets while clearing their sharenfs property and mount
+# children without mounting the parent datasets first. Given that ZFS
+# automatically creates directories for mountpoints, mounting children
+# before the parents would end up creating sub-directories under their
+# parent's mountpoints and so parents would fail to mount later. This
+# is something we really want to avoid, so as a precaution we make sure
+# that all the datasets that we manipulate are already mounted.
+#
+unmounted_datasets = [ds for ds in datasets if ds[2] != 'yes']
+datasets = [ds for ds in datasets if ds not in unmounted_datasets]
+
+if unmounted_datasets:
+    print("WARNING: Not clearing sharenfs property for the following %d "
+          "datasets because they are unmounted:" % len(unmounted_datasets))
+    for ds in unmounted_datasets:
+        print("  %s" % ds[0])
+
+if datasets:
+    #
+    # We must use zfs inherit to clear the sharenfs property instead
+    # of setting sharenfs=off as doing the latter would set the SOURCE
+    # of the sharenfs property to "local" instead of "default". The
+    # app-stack shares timeflows by setting sharenfs on the timeflows and
+    # letting the setting propagate to its children, however this
+    # would not longer work if SOURCE for the sharenfs property is set to
+    # "local" for the children.
+    #
+    # We want to call zfs inherit on the parent datasets before the children,
+    # as such we sort the dataset names by length before traversing them.
+    #
+    datasets.sort(key=lambda ds: len(ds[0]))
+    print("Clearing sharenfs property for the following %d datasets:"
+          % len(datasets))
+    for ds in datasets:
+        print("  %s" % ds[0])
+        subprocess.check_call(["zfs", "inherit", "sharenfs", ds[0]])
+    print("Clearing sharenfs properties completed.")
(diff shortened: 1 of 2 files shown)


sebroy:
How much of a win is this over `zfs inherit -r ...`?

pzakha:
I haven't done performance evaluations but it's significantly faster. I'd say orders of magnitude faster.

pzakha:
So I've just done a quick test with 2488 mounts.

With this fix, it took 1 second with no datasets with sharenfs set, 7 seconds with 30 datasets and 9 seconds with 60 datasets. Note that on another try with 60 datasets it took much more time at 50 seconds, which is probably because something was going on in the background writing do domain0 and extending txg sync times.

Without this fix, it took 85 seconds with no datasets, 225 seconds with 60 datasets shared. 

The important part to remember is that with this fix we only sync once for each dataset that has sharenfs set, whereas without this fix we sync once for each dataset on the system.

files/common/var/lib/delphix-platform/clear-sharenfs
```diff
@@ -0,0 +1,48 @@
...
+
+if datasets:
+    #
+    # We must use zfs inherit to clear the sharenfs property instead
+    # of setting sharenfs=off as doing the latter would set the SOURCE
+    # of the sharenfs property to "local" instead of "default". The
+    # app-stack shares timeflows by setting sharenfs on the timeflows and
+    # letting the setting propagate to its children, however this
+    # would not longer work if SOURCE for the sharenfs property is set to
+    # "local" for the children.
```
sebroy: What you have works, and is a huge improvement. Just thinking out loud: If the children's `sharenfs` properties are already set to `inherit`, then theoretically, we don't need to set those again to `inherit`, and presumably, there are many more of those than there are parents. I wonder if a more efficient approach could be to simply set `sharenfs` to `off` on those datasets that already have a `SOURCE` of `local`, and ignore any whose source is `inherit` given that they'll automatically get that setting.

files/common/var/lib/delphix-platform/clear-sharenfs
```diff
@@ -0,0 +1,48 @@
...
+# children without mounting the parent datasets first. Given that ZFS
+# automatically creates directories for mountpoints, mounting children
+# before the parents would end up creating sub-directories under their
+# parent's mountpoints and so parents would fail to mount later. This
+# is something we really want to avoid, so as a precaution we make sure
+# that all the datasets that we manipulate are already mounted.
+unmounted_datasets = [ds for ds in datasets if ds[2] != 'yes']
+datasets = [ds for ds in datasets if ds not in unmounted_datasets]
+
+if unmounted_datasets:
```
grwilson: What are the ramifications of skipping these datasets as it pertains to migration?

files/common/var/lib/delphix-platform/clear-sharenfs
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/python3
+
+import subprocess
+
+lines = subprocess.check_output(
+    "zfs list -t filesystem -r -H -o name,sharenfs,mounted domain0",
```
grwilson: I'm wondering if it would be faster to just look for any filesystem which has the `sharenfs` property set and the `source` is set to local. For example:

I've set the `sharenfs` property at a higher level and let datasets inherit it below:
```
# zfs list -t filesystem -r -H -o name tank
tank
tank/1
tank/1/2
tank/1/2/3
tank/1/2/3/4
tank/1/2/3/4/5
# zfs get -r -t filesystem -o name -Hs local sharenfs tank
tank/1/2
```

sebroy: Right, that was my feedback as well. I think it's worth doing this given how much boot time this part of the migration workflow is taking up. Anything we can do to squeeze seconds or minutes out of that boot time the better, especially if it's a simple thing like this.

files/common/var/lib/delphix-platform/clear-sharenfs
pzakha: I haven't tested what exact ramifications would there be, I'd guess there might be issues starting those datasets. I could try to test this out by sharing some datasets on Illumos, unmounting them and then going through the migration.

That said, I'm not sure if there are any cases where we would have unmounted datasets with sharenfs set, so I've just added this code as a failsafe.

files/common/var/lib/delphix-platform/clear-sharenfs
pzakha: Yeah, this is similar to George's suggestion. I think filtering only datasets that have `SOURCE` set to `local` is reasonable and should give us the benefit of less calls to spa_sync. I'll try that out.

pzakha:
@grwilson @sebroy Please see the updated version.

I've tested this manually by running on longevity (without the `zfs inherit ...` command).

grwilson:
@pzakha you're trying to get the union of filesystems with properties which have `sharenfs` set and are currently mounted. I proposed using the `zfs get -s local sharenfs` to get the first which should yield a smaller list of datasets. From there you could just parse `/proc/self/mounts` to determine if it's mounted. This would eliminate having to traverse all the datasets twice with the 2 zfs commands. Given the current implementation using two iterations of zfs, I wonder if your first suggestion wouldn't be faster. I suspect that what I outline above may be faster but I would only suggest implementing this if we think it's significantly faster. I was trying to simplify things.

pzakha:
@grwilson Doing 2 zfs calls should be relatively fast. I was thinking of starting with `zfs get -s local sharenfs` but that would have required to do separate calls (either zfs get mount or parsing /proc/self/mounts) for each dataset which could take more time. I think this new implementation is a good mix of simplicity and performance and it does give us the significant performance boost of not calling `zfs inherit sharenfs` on datasets we do not need to.

grwilson:
I'm good with it.

sebroy:
@pzakha how does this latest iteration compare in terms of time vs your original implementation?

pzakha:
@sebroy I haven't timed it, but it seemed faster. Given that we perform significantly less txg syncs with this iteration I'd say it's worthwhile to have this change.

PR #200, opened 2020-02-06
DLPX-68392 performance regression on linux kernel 5.3 due to init_on_alloc

ahrens:
The 5.3 linux kernel adds a new feature which allows pages to be zeroed when allocating or freeing them: init_on_alloc and init_on_free. init_on_alloc is enabled by default on Ubuntu. ZFS allocates and frees pages frequently (via the ABD structure), e.g. for every disk access. The additional overhead of zeroing these pages is significant. I measured a ~40% regression in performance of an uncached "zfs send ... >/dev/null".
init_on_alloc enabled: http://pharos.delphix.com/shares/export/home/mahrens/FlameGraph/29956.svg
init_on_alloc disabled: http://pharos.delphix.com/shares/export/home/mahrens/FlameGraph/14672.svg
We should disable this in the product by setting init_on_alloc=0 in the GRUB kernel boot parameters.
Linux kernel commit: https://github.com/torvalds/linux/commit/6471384af2a6530696fc0203bafe4de41a23c9ef

testing: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2826/

### DIFF
--- a/files/common/etc/default/grub.d/override.cfg
+++ b/files/common/etc/default/grub.d/override.cfg
@@ -98,3 +98,18 @@ GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT elevator=noop"
 #
 GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT crashkernel=256M,high"
 GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT crashkernel=256M,low"
+
+#
+# Disable unnecessary zeroing of allocated pages.
+#
+# The 5.3 linux kernel adds a new feature which allows pages to be
+# zeroed when allocating or freeing them: init_on_alloc and
+# init_on_free. init_on_alloc is enabled by default on Ubuntu. ZFS
+# allocates and frees pages frequently (via the ABD structure), e.g. for
+# every disk access. The additional overhead of zeroing these pages is
+# significant. I measured a ~40% regression in performance of an
+# uncached "zfs send ... >/dev/null".
+#
+# Upstream commit: 6471384af2a6530696fc0203bafe4de41a23c9ef
+#
+GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT init_on_alloc=0"


pzakha:
Can you add a link to a git ab-pre-push run?

files/common/etc/default/grub.d/override.cfg
```diff
@@ -98,3 +98,18 @@ GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT elevator=noop"
...
+# zeroed when allocating or freeing them: init_on_alloc and
+# init_on_free. init_on_alloc is enabled by default on Ubuntu. ZFS
+# allocates and frees pages frequently (via the ABD structure), e.g. for
+# every disk access. The additional overhead of zeroing these pages is
+# significant. I measured a ~40% regression in performance of an
+# uncached "zfs send ... >/dev/null".
+#
+# Upstream commit: 6471384af2a6530696fc0203bafe4de41a23c9ef
+#
+GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT init_on_alloc=0"
```
grwilson: Do we also need to disable `init_on_free`?

grwilson:
I found a blog post where the writer, talking about `init_on_alloc`, states  "The performance impact of the former under most workloads appears to be under 1%, if it’s measurable at all." Guess the workload their testing is not zfs. :)

files/common/etc/default/grub.d/override.cfg
ahrens: init_on_free is disabled by default on Ubuntu.

prakashsurya: If we specifically need `init_on_free=0`, I think it might be good to set that explicitly here, just in case Ubuntu ever decides to change that default setting..

ahrens:
@pzakha yes sorry I forgot to add that to the initial comment.  Updated.  I am assuming that the setting will be ignored on any older kernels that we may be running.  Do you think we need to test that, and if so do you know which platforms are running kernels <5.3 on master?

pzakha:
@ahrens The only kernel running 5.3 on master is the generic one (that translates to esx and kvm platforms). I think extraneous settings will be ignored, but to be safe it's easy to test by setting it manually on a system with an older kernel.

pzakha:
@prakashsurya Regarding deferred upgrade, I believe this setting will only be needed once users reboot anyway since they will be running an older kernel before that.

ahrens:
@prakashsurya that's right.  Since we don't have any releases on 5.3, this is fine.

@pzakha I tested setting a bogus variable and the system booted fine.

pzakha:
Note that we may need to add this setting to migration boot options later, however this should not be an issue for now as we are planning to use the 5.0 kernel for 6.0.1.0.

PR #206, opened 2020-02-25
DLPX-67907 Upgrade verification / apply fails if rpool/crashdump is full or larger than default

sumedhbala-delphix:
Bugfix: On older engines with smaller roool sizes, support sometimes manually sets the crashdump quota to a higher number to allow system dumps to be stored. The app-gate has logic to clean the dumps. 
Currently the quota is set to 50% which can fail if crashdump is using >50%.  "msg": "cannot set property for 'rpool/crashdump': size is less than current used or reserved space\n"
This takes care of the case by taking into account both used space and size of rpool. If the dumps get cleaned up, on the next upgrade quota will be set back to 50%.

Testing: 
git ab-pre-push --test-upgrade-from 6.0.1.0: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2971/
Rerun of upgrade which failed due to a mssql bug: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/3485/

Upgrade testing: Verified on an engine which has crashdump >50% of rpool, upgrade verification fails without the fix but passes with the fix.

Followed the readme and made sure "./scripts/docker-run.sh make packages" works.

### DIFF
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -61,24 +61,38 @@
 
 #
 # We want the quota on the crashdump dataset to be sized accordingly
-# based on the capacity of rpool. Thus, we use this command to query the
-# rpool size, and then use this value later to set the crashdump quota.
+# based on the capacity of rpool and space already used by crashdump.
+# We pick greater of ( 50% of rpool, space used by crashdump).
+# We use the following commands to query the rpool size and used
+# crashdump space, and use these values later to set the crashdump quota.
+# We also bump up the used space by 1GB in case it is currently filling
+# up to avoid a race
+# This is done to ensure its contents can't run rpool out of space
 #
+
+- zfs:
+    name: rpool/crashdump
+    state: present
+  when: not ansible_is_chroot
+
 - command: zpool list -Hpo size rpool
   register: rpool_size_bytes
   when: not ansible_is_chroot
 
-#
-# Reconfigure the crashdump dataset. The initial configuration was performed
-# by appliance build so we only update the configuration here. In particular,
-# we want to enforce a quota to ensure its contents can't run rpool out
-# of space.
-#
+- zfs_facts:
+    name: rpool/crashdump
+    recurse: no
+    parsable: yes
+    properties: used
+    type: filesystem
+  register: crashdump
+  when: not ansible_is_chroot
+
 - zfs:
     name: rpool/crashdump
     state: present
     extra_zfs_properties:
-      quota: "{{ rpool_size_bytes.stdout|int / 2 }}"
+      quota: "{{ [rpool_size_bytes.stdout|int / 2, crashdump.ansible_facts.ansible_zfs_datasets[0].used|int + 1073741824]| max }}"
   when: not ansible_is_chroot
 
 #


files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -61,24 +61,36 @@
...
+    type: filesystem
+  register: crashdump
+  when: not ansible_is_chroot
+
 - zfs:
     name: rpool/crashdump
     state: present
     extra_zfs_properties:
-      quota: "{{ rpool_size_bytes.stdout|int / 2 }}"
+      quota: "{{ [rpool_size_bytes.stdout|int / 2, crashdump.ansible_facts.ansible_zfs_datasets[0].used|int + 1]| max }}"
```
pzakha: Per curiosity, why the `+1`?

sumedhbala-delphix: I added it for safety. I once hit a failure where setting the quota to the exact number failed. I have not been able to repro the problem though.

prakashsurya:
Out of curiosity, rather than setting the quota to higher than 50%, would it also fix the bug if we ignored the error? i.e. always attempt to set the quota to 50%, but ignore errors when that fails?

Or is it actually better to still attempt to set the quota, even if we have to go higher than 50% of rpool size? e.g. perhaps to ensure the used space by the dataset doesn't grow any larger than that?

sumedhbala-delphix:
> Out of curiosity, rather than setting the quota to higher than 50%, would it also fix the bug if we ignored the error? i.e. always attempt to set the quota to 50%, but ignore errors when that fails?
> 
> Or is it actually better to still attempt to set the quota, even if we have to go higher than 50% of rpool size? e.g. perhaps to ensure the used space by the dataset doesn't grow any larger than that?

I think it is definitely safer to set it to prevent rpool from getting full.

prakashsurya:
> I think it is definitely safer to set it to prevent rpool from getting full.

I agree.

Part of why I ask, though, is this method is still prone to failure since the crashdump dataset will likely be mounted when this is run. Thus, it's possible for the "used" size to report something, and then for that to increase before we set the quota (e.g. because some userspace program crashed in between those two operations); if this were to occur, we'd fail (assuming we're using the "used" space to set the quota).

How do we want to handle this potential race condition?

sumedhbala-delphix:
> > I think it is definitely safer to set it to prevent rpool from getting full.
> 
> I agree.
> 
> Part of why I ask, though, is this method is still prone to failure since the crashdump dataset will likely be mounted when this is run. Thus, it's possible for the "used" size to report something, and then for that to increase before we set the quota (e.g. because some userspace program crashed in between those two operations); if this were to occur, we'd fail (assuming we're using the "used" space to set the quota).
> 
> How do we want to handle this potential race condition?

Instead of adding 1, I could change to to bump by 1GB , since the commands are run close to each other, it should cover the race?

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
pzakha: I think the 1G is safer in this case. It is possible that some metadata was updated for that dataset after mounting it, as such the difference in size would be more than 1. Given that you did run into a failure, it would probably be safer to go this way.

prakashsurya:
Why do we think 1G is any safer? What is the typical size of a dump file?

In the edge cases where we think this new logic will kick in, what do we expect "used" to be relative to the rpool's capacity? e.g. could it be large enough, such that "used + 1G" is more than the rpool capacity (at which case, I'd expect setting the quota would fail)?

I'm not necessarily against this approach, I just feel like there's some potential failure scenarios present, and I'm not sure if those are practical concerns to be worried about or not.

What alternative solutions are available to determine which quota we should use, apart from the max like you're doing here? e.g. could we check rpool capacity, and if it's less than some threshold, then set the quota to something higher than 50%? In these "small rpool" cases, what's a typical quota that is used by support?

prakashsurya:
@sumedhbala-delphix should we close this out, now that we've landed #209?

PR #208, opened 2020-03-20
DLPX-68763 crashdump quota should not get reset on upgrade

prakashsurya:


prakashsurya:
This depends on: https://github.com/delphix/appliance-build/pull/452

`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3089/)

prakashsurya:
cc: @sumedhbala-delphix

sumedhbala-delphix:
I will close https://github.com/delphix/delphix-platform/pull/206 once this gets merged,

Basic question: how will this work on migration? On illumos we set the quota to 25% but the quota on new installs on linux is 50%?

prakashsurya:
> how will this work on migration? On illumos we set the quota to 25% but the quota on new installs on linux is 50%?

As far as I understand, we'll keep the existing quota after upgrade/migration, so it'll stay at 25%. This is by design, though. If we don't want to update the quota on upgrade, this is the trade off. In general, as far as this is quota concerned, "migration" is no different "upgrade".

With that said, I'm not sure why we moved from 25% to 50% during the development of the 6.0 release. If that was due to something Linux specific, perhaps we should add some logic post-migration to handle this difference? Although, if a 25% quota worked fine on illumos, I think it should work fine on Linux, too. So, my initial thought is we should just treat this the same as any other upgrade, and maintain the existing 25% quota (or whatever it may be on the engine).

prakashsurya:
> Do we have plans to backport this to 6.0?

Yes, we have a backport bug open for 6.0.2 here: https://jira.delphix.com/browse/DLPX-68769

> If so this would mean that all the Illumos systems migrated to 6.0.0.0 and 6.0.1.0 first will have a 50% quota, but those migrated directly to 6.0.2.0 (or wherever we decide to backport) would have the 25% quota set on Illumos.

I presume that is the intention; i.e. we want to preserve the original quota, rather than modify it on upgrade. @grwilson can you comment on this?

grwilson:
Could we keep the same behavior by having the migration scripts update the quota?

pzakha:
> Could we keep the same behavior by having the migration scripts update the quota?

We can add similar logic to https://github.com/delphix/delphix-platform/pull/206 into the `migrate-os` script.

prakashsurya:
OK, so here's what we want to accomplish:

1. for illumos to linux migrations, we want to set the quota to 50% (where possible)
2. for linux to linux upgrades, we don't want to modify the quota at all

This PR current implements (2). I'll look into the changes required to implement (1).

I hadn't realized we wanted to do (1), which is why I hadn't implemented that part, but I'll look into that now.

sumedhbala-delphix:
>     1. for illumos to linux migrations, we want to set the quota to 50% (where possible)
 What does "where possible" mean? If the used space is >50%, will we be setting it to = ~used space?

prakashsurya:
> What does "where possible" mean? If the used space is >50%, will we be setting it to = ~used space?

I think we should set the quota on migration to a minimum of 50%. So, if used is greater than (or equal to) 50%, that necessarily implies the quota is already greater than (or equal to) 50%, so in that case I'd suggest we not modify the quota at all.

What do you think?

prakashsurya:
@pzakha since the migration related files have been removed from master, am I good to land this as-is? I'll open a backport with this change, and include the migration specific changes in the backport PR.

PR #209, opened 2020-03-24
DLPX-68769 [Backport of DLPX-68763 to 6.0.2.0] crashdump quota on rpool gets reset on upgrade

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3104/)
`git-ab-pre-push --test-upgrade-from 5.3.7.0` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3105/)

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -63,6 +63,19 @@ function perform_migration() {
...
+	CRASHDUMP_QUOTA="$(echo "$(zpool list -Hpo size rpool) / 2" | bc)"
+	if [[ "$CRASHDUMP_USED" -lt "$CRASHDUMP_QUOTA" ]]; then
+		echo "zfs set quota=${CRASHDUMP_QUOTA}b rpool/crashdump"
+		zfs set quota="${CRASHDUMP_QUOTA}b" rpool/crashdump ||
+			die "Failed to set quota on rpool/crashdump;" \
+				"used ($CRASHDUMP_USED)," \
+				"target quota ($CRASHDUMP_QUOTA)"
+	else
+		echo "note: rpool/crashdump used ($CRASHDUMP_USED) >=" \
+			"target quota ($CRASHDUMP_QUOTA)"
```
pzakha: So the idea here is that the quota is either set to more than half the pool or not set at all. If the quota is set to more than half the pool, this was probably done deliberately by support. However, if the quota is not set at all, do you think we should still set it to some value (similar to #206) to prevent running the rpool out of space in the future?

prakashsurya: > However, if the quota is not set at all, do you think we should still set it to some value

Under what conditions would there be no quota set at all, and when those conditions are met, why would it be correct to set a quota?

> So the idea here is that the quota is either set to more than half the pool or not set at all.

The idea is to set the quota to a minimum of half of rpool's size, while attempting to preserve any manual tuning of the quota previously done by support.

If "used" is already greater than that size, then the quota has already been modified by support to match this criteria (minimum of half of rpool size), so we do nothing.

If "used" is already greater due to there not being any quota at all, the removal of the quota would have also been a modification by support, so we do nothing. In this case, I don't know why it'd be considered safe to add a quota after it was manually removed; presumably it was removed for good reason, and that reasoning would still apply on linux.

pzakha: > If "used" is already greater due to there not being any quota at all, the removal of the quota would have also been a modification by support, so we do nothing. In this case, I don't know why it'd be considered safe to add a quota after it was manually removed; presumably it was removed for good reason, and that reasoning would still apply on linux.

That sounds reasonable to me. Let's have supports opinion on this perhaps?

prakashsurya: I don't understand the need to get support's opinion. Previously on illumos, we wouldn't change/set a quota on upgrade, so this is no different in that specific case. With that said, sure, go ahead and pull in somebody from support to weigh in.

prakashsurya:
I discussed this with Anthony offline, and he's OK moving forward with this as-is, so I'll merge this.

PR #210, opened 2020-03-24
DLPX-66808 Boot/splash screen for Linux

grwilson:
This PR add the `delphix-startup-screen` as a python3 program and converts the keyboard logic to use `localectl`. The delphix console is started on both `ttyS0` and `tty1` and it runs as a pre-exec script in the `serial-getty@` and `getty@` services. We override only the `getty@tty1.service` (I think linux can have up to 6 ttys).

### DIFF
--- a/files/common/usr/bin/delphix-startup-screen
+++ b/files/common/usr/bin/delphix-startup-screen
@@ -1,0 +1,394 @@
+#!/usr/bin/python3
+#
+# Copyright 2020 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""
+This program displays the delphix console screen which includes
+the IP address, hostname, and services (description and status).
+"""
+
+import curses
+import curses.ascii
+import curses.textpad
+import sys
+import os
+import logging
+import subprocess
+from typing import List, Any, Tuple, Generator
+
+STATUS_HEADER: str = "STATUS\t\t\tSERVICE\t\t\t"
+LAYOUT_STR: str = "Keyboard layout (press 9 to change):"
+CONSOLE_STR: str = "Press any other key for console access."
+KEY_LAYOUT_SELECT_STR: str = "Select a keyboard layout by number [1-%d]:"
+LOGO: str = "Delphix Version "
+
+log_file: str = "/tmp/dpx_startup_screen.log"
+
+website: str = "www.delphix.com"
+
+
+def get_svcs() -> List[str]:
+    """
+    Retrieve the service name and state for each service
+    that relies on 'delphix.target'. This will dynamically
+    update based on services which are WantedBy delphix.target.
+    """
+    cmd: List[str] = ['systemctl', 'show', '-p', 'Wants', 'delphix.target']
+    cp = subprocess.run(cmd,
+                        stdout=subprocess.PIPE,
+                        universal_newlines=True,
+                        check=True)
+    return cp.stdout.split("=", 1)[1].split()
+
+
+def run_svcs(options: str = None) -> Generator:
+    """
+    Run the command to retrieve the status of all the services. The options
+    argument can be used to obtain additional output from the 'systemctl'
+    command. By default we only obtain the 'ActiveState' of the services.
+    Each line is separated by a newline and an extra newline is added to
+    separate services.
+
+    For example when requesting 'ActiveState' and 'Description' the output
+    would look like this:
+    'Delphix management service\nactive\n\nDelphix Masking Service\ninactive\n'
+
+    After invoking splitlines(), we filter out any empty lines, so that we
+    end up with a List of strings:
+    ['Delphix management service', 'active',
+    'Delphix Masking Service', 'inactive']
+    """
+
+    fields: str = 'ActiveState'
+    cmd: List[str] = ['systemctl', 'show', '--value', '-p']
+    if options:
+        fields += ',' + options
+    cmd.append(fields)
+    for svc in get_svcs():
+        cmd.append(svc)
+    cp = subprocess.run(cmd,
+                        stdout=subprocess.PIPE,
+                        universal_newlines=True,
+                        check=True)
+    yield from filter(None, cp.stdout.splitlines())
+
+
+def get_keyboard_layout() -> str:
+    """
+    Return the current keyboard layout.
+    """
+    cmd: List[str] = ['localectl', 'status']
+    cp = subprocess.run(cmd,
+                        stdout=subprocess.PIPE,
+                        universal_newlines=True,
+                        check=True)
+    output: List[str] = cp.stdout.split()
+    return output[output.index("Layout:") + 1]
+
+
+def set_keyboard_layout(layout: str) -> subprocess.CompletedProcess:
+    """
+    Set the keyboard layout based on the user selection.
+    """
+    cmd: List[str] = ['localectl', 'set-x11-keymap', layout, 'pc105']
+    subprocess.run(cmd, shell=False, check=True)
+    return subprocess.run('setupcon', shell=False, check=True)
+
+
+def get_valid_keyboard_layouts() -> List[str]:
+    """
+    Return valid keyboard layouts supported by the system.
+    """
+    cmd: List[str] = ['localectl', '--no-pager', 'list-x11-keymap-layouts']
+    cp = subprocess.run(cmd,
+                        stdout=subprocess.PIPE,
+                        universal_newlines=True,
+                        check=True)
+    return cp.stdout.splitlines()
+
+
+def load_header(stdscr) -> Any:
+    """
+    Display the header information for the main screen.
+    """
+    WIN_LEN, WIN_HT = set_common_variables(stdscr)
+    cmd = ['get-appliance-version', '--patch']
+    cp = subprocess.run(cmd,
+                        stdout=subprocess.PIPE,
+                        universal_newlines=True,
+                        check=True)
+    version: str = cp.stdout
+
+    stdscr.clear()
+    stdscr.addstr(1, 2, LOGO + str(version), curses.A_BOLD)
+    stdscr.addstr(1, WIN_LEN - len(website) - 2, website, curses.A_BOLD)
+
+    win = stdscr.subwin(WIN_HT - 4, WIN_LEN - 4, 3, 2)
+    win.bkgd(' ', curses.color_pair(2))
+    win.box()
+    return win
+
+
+def is_server_down() -> bool:
+    """
+    Determine if the server is down by looking at the state of all the
+    services. If any service is 'failed' then we declare the server down.
+    """
+    for i in run_svcs():
+        if i.strip() == "failed":
+            return True
+    return False
+
+
+def is_server_up() -> bool:
+    """
+    Determine if all the services are 'active' or 'inactive'. Inactive
+    is considered up since the service may not be enabled.
+    """
+    for i in run_svcs():
+        if i.strip() not in ("active", "inactive"):
+            return False
+    return True
+
+
+#
+# Sets the common variables.
+#
+def set_common_variables(stdscr) -> Tuple[int, int]:
+    """
+    Initialize curses colors and settings.
+    """
+    if curses.has_colors():
+        curses.start_color()
+        curses.init_pair(1, curses.COLOR_BLACK, curses.COLOR_WHITE)
+        curses.init_pair(2, curses.COLOR_WHITE, curses.COLOR_BLUE)
+        curses.init_pair(3, curses.COLOR_BLACK, curses.COLOR_GREEN)
+        curses.init_pair(4, curses.COLOR_BLACK, curses.COLOR_YELLOW)
+        curses.init_pair(5, curses.COLOR_BLACK, curses.COLOR_RED)
+        stdscr.bkgd(' ', curses.color_pair(1))
+
+    (Y, X) = stdscr.getmaxyx()
+
+    return X, Y
+
+
+def getstatus_pair(lst) -> Generator:
+    """
+    Process the status and return a list which is comprised of
+    ['service description', 'service status'].
+    """
+    for i in range(0, len(lst), 2):
+        yield lst[i:i + 2]
+
+
+def getstatus() -> str:
+    """
+    Each service returns 2 lines -- one with the description and the
+    next line with the state.
+    """
+    status = ""
+    for i in getstatus_pair(list(run_svcs("Description"))):
+        #
+        # Switch the state and description
+        #
+        status += f"{i[-1]}\t{i[0]}\n"
+    return status
+
+
+def get_network_status() -> Tuple[str, str]:
+    """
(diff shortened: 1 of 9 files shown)


prakashsurya:
Do we want the `delphix-startup-screen` service to be restarted after an upgrade? If so, we might want to hook it up with the `delphix.target` service, like we do for some other services.

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,330 @@
...
+    out = run_svcs()
+    for i in out.splitlines(False):
+        #
+        # Skip any empty lines
+        #
+        if not i.strip():
+            continue
+
+        if i.strip() == "failed":
+            return True
```
sdimitro: Couldn't this be
```
for i in out.splitlines():    # I think False is the default?
    if i.strip() == "failed":   # If i.strip() is empty we'll be skipping anyways
        return True
return False
```

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,330 @@
...
+    out = run_svcs().rstrip("\n")
+    for i in out.splitlines(True):
+        #
+        # Skip any empty lines
+        #
+        if not i.strip():
+            continue
+
+        if i.strip() != "active" and i.strip() != "inactive":
+            return False
```
sdimitro: What's different here from `is_server_down()` that we are passing `True` to `splitlines()` ?
Does it matter if we are going to `strip()` things anyways? also like before couldn't we get rid of `if not i.strip():` ?

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,330 @@
...
+        if i.strip() == "failed":
+            return True
+    return False
+
+def is_server_up():
+    '''
+    Determine if all the services are 'active' or 'inactive'. Inactive
+    is considered up since the service may not be enabled.
+    '''
+    out = run_svcs().rstrip("\n")
```
sdimitro: Again what is different here from `is_server_down()` that we have the extra `rstrip("\n")`?

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,330 @@
...
+    next line with the state. In addition, a newline separates each
+    service.
+    '''
+    out = run_svcs("Description").splitlines()
+    str = ""
+    for i in range(0, len(out), 3):
+        #
+        # Switch the state and description
+        #
+        str = str + out[i+1] + "\t" + out[i] + "\n"
```
sdimitro: Also here (assuming the rename of `str` you can do:
```
status_str += f"{out[i+1]}\t{out[i]}\n"
```

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,330 @@
...
+
+def get_valid_keyboard_layouts():
+    cmd = ['localectl', '--no-pager', 'list-x11-keymap-layouts']
+    cp = subprocess.run(cmd, stdout=subprocess.PIPE, universal_newlines=True)
+    return cp.stdout.splitlines()
+
+def load_header(stdscr):
+    global logo
+
+    cmd = ['/opt/delphix/server/bin/get_appliance_version']
```
prakashsurya: Let's use `get-appliance-version` instead, which should be in the default PATH.

We should probably remove the `get_appliance_version` script that's delivered by the app-gate, to avoid this sort of confusion.

prakashsurya: We don't want to use `get_appliance_version` since that'd cause the platform package to be dependent on the virtualization package (at least for this script), which we don't want. The `get-appliance-version` script is delivered by the platform package, so there's no package dependency issues this way.

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,330 @@
...
+    cp = subprocess.run(cmd, stdout=subprocess.PIPE, universal_newlines=True)
+    version = cp.stdout.split('-', 1)[0]
```
prakashsurya: I don't feel strongly, but we could potentially add an option to the `get-appliance-version` script to break up the version information like you're doing here.. this way, that same functionality could be use elsewhere, rather than forcing the consumers to know how to properly split up the string that is returned.

e.g. something like a `--major` and/or `--minor` options.. or maybe allow a printf style string to be supplied.. 

With that said, I'm good with what you have as-is.

files/common/usr/bin/delphix-startup-screen
grwilson: This is not necessary and I've removed this.

files/common/usr/bin/delphix-startup-screen
grwilson: The problem is the way that the service information is returned. Here's an example of the output after `splitlines()`:
```
     ['active', '', 'inactive']
```
So you'll notice that there is an empty string that separates the information so if we don't skip the empty string then we'll report the server as not being 'up'

files/common/usr/bin/delphix-startup-screen
grwilson: I've cleaned both of these functions up and will post a new version.

files/common/usr/bin/delphix-startup-screen
grwilson: What do you think about making the default always report the first part of the version and having `--full` or `--debug` to get the snapshot info?

prakashsurya: Seems reasonable; I'm not sure how that'll interact with other software using that script, but seems reasonable to me at first thought.

grwilson: I looked at the other consumers of `get-appliance-version` in `delphix-platform` repo and I think your original proposal of `--major` made more sense. I was trying to determine the use cases for `--minor` and `--patch`. Right now they return back just that portion of the version but I'm not sure how useful that would be especially since I'm not currently restricting the options to be mutually exclusive. If I make the options mutually exclusive then we can either have `--minor` return back just the minor portion of the version or return back the whole version minus the patch portion. In the latter case, it makes the need for `--patch` unnecessary since it would overlap with the default case. Let me know what you think.

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,400 @@
...
+    services. If any service is 'failed' then we declare the server down.
+    """
+    for i in run_svcs():
+        #
+        # Empty strings separate services, skip those.
+        #
+        line: str = i.strip()
+        if line and line.strip() == "failed":
+            return True
+    return False
```
sdimitro: [1] if `line` is equal to `i.strip()` then later when you do `line.strip()` it should be equal to `i.strip().strip()`. Is there a reason why we are stripping twice?
 
[2] In our previous conversation you said that the output that we iterate over is `['active', '', 'inactive']` and that's why you are checking if `line` is equal to the empty string, correct? My point before was that you can still compare the empty string to "failed" and continue to the next iteration of the loop:
```
$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "" == "failed"
False
```
That said I understand if you want to keep it like that to warn the reader that `line` can also be an empty string, but I think the comment could do that for you.

Similarly for the `is_server_up()` function.

files/common/usr/bin/delphix-startup-screen
prakashsurya: > If I make the options mutually exclusive then we can either have --minor return back just the minor portion of the version or return back the whole version minus the patch portion.

Yea, I think it makes sense to make them exclusive, and return the version up to the option specified; i.e. major is just the two major number, minor is the 3 numbers, patch is the 4 numbers, etc.

> In the latter case, it makes the need for --patch unnecessary since it would overlap with the default case. 

Would `--patch` also return the snapshot information? e.g. on my system:
```
6.1.0.0-snapshot.20200325201630621+jenkins-ops-devops-gate-master-appliance-build-master-post-push-2822
```
I could imagine `--patch` only returning `6.0.0.0`

I could also imagine a `--snapshot` option that returns: `6.1.0.0-snapshot.20200325201630621` (this would be used by the MOTD script)

and then the default, that'd return the entire string.

> I looked at the other consumers of get-appliance-version in delphix-platform repo

There's also "consumers" in the app-gate.. those are currently using the app-gate's `get_appliance_version` script, but arguably those should be using the platform script.. e.g. if we change how the version is stored, we'd update the platform script, and the app-gate would not have to know about that.

But I think all the consumers of the app-gate would be able to use the new options that you mentioned (e.g. `--major`, `--minor`, etc.).

files/common/usr/bin/delphix-startup-screen
```diff
@@ -0,0 +1,400 @@
+#!/usr/bin/python3
+"""
+This program displays the delphix console screen which includes
```
prakashsurya: I believe this comment can be placed below the copyright comment.. IIRC, this is the convention we use in the SDB repo (docstring below the copyright).

files/common/usr/bin/delphix-startup-screen
grwilson: I have reworked the functions so take another look.

files/common/usr/bin/delphix-startup-screen
grwilson: Here's the output of all the options:
```
# ./get-appliance-version 
6.1.0.0
# ./get-appliance-version --patch
6.1.0.0
# ./get-appliance-version --snapshot
6.1.0.0-snapshot.20200325231145826
# ./get-appliance-version --major
6.1
# ./get-appliance-version --minor
6.1.0
# ./get-appliance-version --debug
6.1.0.0-snapshot.20200325231145826+jenkins-ops-devops-gate-master-appliance-build-master-post-push-2824
```

files/common/usr/bin/get-appliance-version
```diff
@@ -18,6 +18,76 @@
...
+	echo "               information, if it exists."
+	echo "-M, --major -- displays the major version"
+	echo "-m, --minor -- displays the major and minor version"
+	echo "-p, --patch -- displays the major, minor, and patch version"
+	echo "-s, --snapshot -- displays the major, minor, patch, and " \
+		"snapshot information"
+	exit 2
+}
+
+options=$(getopt -l "major,minor,patch,snapshot,debug" -o "Mmpsd" -- "$@")
```
prakashsurya, line 38: Can we support a `-h` and/or `--help` option, that'd print the usage information?

e.g. right now:
```
$ ./get-appliance-version -h
getopt: invalid option -- 'h'
```

files/common/usr/bin/get-appliance-version
```diff
@@ -18,6 +18,76 @@
...
+done
+
+$opt_debug && echo "$output" | tr -d '\n'
+$opt_major && echo "$output" | cut -d'.' -f1,2 | tr -d '\n'
+$opt_minor && echo "$output" | cut -d'.' -f1-3 | tr -d '\n'
+$opt_patch && echo "$output" | cut -d'-' -f1 | cut -d'.' -f1-4 | tr -d '\n'
+$opt_snapshot && echo "$output" | cut -d'+' -f1
+
+if ! $opt_major && ! $opt_minor && ! $opt_patch && ! $opt_debug; then
+	echo "$output" | cut -d'-' -f1 | tr -d '\n'
```
prakashsurya, line 92: Why don't we want the default behaviour to print out the full version?

e.g.
```
$ ./get-appliance-version
6.1.0.0
```

The actual "appliance version" includes all the additional information, which is what's displayed in the UI and tracked in MDS. This enables us to do things like trunk to trunk upgrades, since the actual version is different. This means, any app-gate consumers that need the full version will use the `--debug` option, which feels a little awkward to me.

I don't feel strongly; just wanted to share my thoughts as I'm reading over this change...

files/common/usr/bin/get-appliance-version
```diff
@@ -18,6 +18,76 @@
...
+	esac
+done
+
+$opt_debug && echo "$output" | tr -d '\n'
+$opt_major && echo "$output" | cut -d'.' -f1,2 | tr -d '\n'
+$opt_minor && echo "$output" | cut -d'.' -f1-3 | tr -d '\n'
+$opt_patch && echo "$output" | cut -d'-' -f1 | cut -d'.' -f1-4 | tr -d '\n'
+$opt_snapshot && echo "$output" | cut -d'+' -f1
+
+if ! $opt_major && ! $opt_minor && ! $opt_patch && ! $opt_debug; then
```
prakashsurya, line 91: I believe this is broken when the `-s` option is used; e.g.
```
$ ./get-appliance-version -s
6.1.0.0-snapshot.20200330155231164
6.1.0.0
```

Additionally, I think it might be cleaner to rely on setting one of the `opt_` variables to true, to trigger the default behaviour. This way, there's no "special case" for the default handling.

files/common/usr/bin/get-appliance-version
```diff
@@ -18,6 +18,76 @@
...
+	echo "-p, --patch -- displays the major, minor, and patch version"
+	echo "-s, --snapshot -- displays the major, minor, patch, and " \
+		"snapshot information"
+	exit 2
+}
+
+options=$(getopt -l "major,minor,patch,snapshot,debug" -o "Mmpsd" -- "$@")
+eval set -- "$options"
+
+output=$(zfs get -Hpo value \
```
prakashsurya, line 41: rather than add the `tr -d` to all the `echo` lines below, I think it might be better to do the `tr -d` call once here, or remove the call entirely (I can't remember why it was added to begin with).

files/common/usr/bin/get-appliance-version
prakashsurya: > Why don't we want the default behaviour to print out the full version?

Sorry, just to be clear, I mean why don't we want to print the "debug" version by default..

grwilson: Sure, I wasn't aware that we stored this.

files/common/usr/bin/get-appliance-version
grwilson: Yeah, now that everything is mutually exclusive, I can handle this better in the case statement.

files/common/usr/bin/get-appliance-version
grwilson: Sounds good.

files/common/usr/bin/get-appliance-version
grwilson: Sure

files/common/usr/bin/get-appliance-version
prakashsurya: image

Yes, currently the app-gate uses all of the version information, including the snapshot and build information. Additionally, the `VERSION` of an upgrade image, also includes all of the snapshot and build information in it.

prakashsurya: With that said, I think we could probably make do with the default only including the snapshot information, and not the build/jenkins information.

I think the main constraint is the version is different between builds, so that we can easily support trunk to trunk upgrades. I believe the snapshot information alone satisfies that, so I'd be good with using `-s` as the default mode.

files/common/usr/bin/get-appliance-version
grwilson: turns out that `cut` will add a newline so to keep the old behavior I need to do the `tr -d` after each command.

files/common/usr/bin/get-appliance-version
grwilson: I'm assuming the snapshot and jenkins info only exists on non-released builds. Is that true? I think leaving the old semantics is the safest thing since that's what we've been using. Anybody wanting some portion of the version should supply the necessary option.

prakashsurya: > Is that true?

Yes, as far as I understand, that's true.

> I think leaving the old semantics is the safest thing since that's what we've been using.

Sounds good to me.

PR #212, opened 2020-03-26
Replace TravisCI with Github Actions

prakashsurya:
This change adds the necessary files and logic to take advantage of
GitHub Actions for running our automated tests. Currently we rely on
TravisCI to do this, but we'd like to transition over to GitHub Actions,
for the following reasons:

    1. The Actions UI is more integrated with GitHub PRs
    2. Best practices at Delphix suggests using Actions

Additionally, this change removes integration with the "bors" bot.


shartse:
I don't have a lot of experience with github actions, but this looks reasonable to me. 

One small suggestion I have is to reword the name of the PR - as it is it's a little unclear if we're moving to or from TravisCI. Maybe: "Replace TravisCI with Github Actions" or "Leverage Github Actions instead of TravisCI"

prakashsurya:
> Maybe: "Replace TravisCI with Github Actions"

Done!

PR #217, opened 2020-04-08
DLPX-69340[Backport of DLPX-69339 to 6.0.2.0] migration should handle overridden physical block sizes

grwilson:
This portion of the fix consumes an overridden ashift value and sets the pool-wide ashift property so that any devices added in the future will use the overridden ashift value.

The pre-upgrade portion of this change can be found here: http://reviews.delphix.com/r/57404

### DIFF
--- a/files/common/var/lib/delphix-platform/os-migration
+++ b/files/common/var/lib/delphix-platform/os-migration
@@ -130,6 +130,22 @@ function perform_migration() {
 		"STRESS_MIGRATION_FAIL_AFTER_DOMAIN0_IMPORT"
 
 	#
+	# The ashift value was stored as a user property on the domain0
+	# dataset so must wait till the pool is imported to read it.
+	# If the value of dlpx:ashift is "-" or "0" then we skip setting
+	# the ashift property. Otherwise, we set the pool-wide ashift
+	# property to the ashift value that was uses prior to migration.
+	# We always remove the user property we set on domain0.
+	#
+	ashift=$(zfs get -Ho value dlpx:ashift domain0)
+	if [[ "$ashift" != "-"  && "$ashift" != "0" ]]; then
+		echo "setting ashift value to $ashift"
+		zpool set ashift=$ashift domain0 ||
+			die "Failed to set ashift value on domain0"
+	fi
+	zfs inherit dlpx:ashift domain0
+
+	#
 	# Note that we must mount the ZFS datasets before clearing sharenfs
 	# as calling after sometimes fails (see DLPX-66005 for more info).
 	#


grwilson:
ab-pre-push containing both appstack and delphix-platform changes: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3247/

Created an image and installed it on a system with an overidden block size:

```
root@gwilson-5380:~# zfs get dlpx:ashift domain0
NAME     PROPERTY     VALUE        SOURCE
domain0  dlpx:ashift  12           local
root@gwilson-5380:~# zpool get ashift domain0
NAME     PROPERTY  VALUE   SOURCE
domain0  ashift    12      local
```

files/common/var/lib/delphix-platform/os-migration
```diff
@@ -129,6 +129,15 @@ function perform_migration() {
...
 		"STRESS_MIGRATION_FAIL_AFTER_DOMAIN0_IMPORT"
 
+	#
+	# The ashift value was stored as a user property on the domain0
+	# dataset so we can't read it until after we've imported the pool.
+	# A value of 0 is the default setting.
+	#
+	ashift=$(zfs get -Ho value dlpx:ashift domain0)
+	zpool set ashift=$ashift domain0 ||
+		die "Failed to set ashift value on domain0"
```
pzakha, line 139: - Could we skip setting this if ashift is 0 or empty?
- Do we need to re-import domain0 for this setting to take effect?

grwilson: We could skip the setting if it's 0 but is there a reason we would want to add more code here? We don't need to re-import for this to take effect.

pzakha: We don't need to do this check per-say, but the idea was to keep changes between migrated and freshly created systems to a minimum.

In the same spirit, we might also want to unset `dlpx:ashift` after we are done with this. 

I'm considering creating a document outlining all of the customizations we do to customer engines, so we can keep track of them in one place. This will make it easier to make sure we do the right testing when needed.

shartse: I think it also might be worth adding a case handling a case where `dlpx:ashift` doesn't get created. What I'm imagining is: `ashift` is non-zero, but `migrate_config.py` fails to create `dlpx:ashift`, and then in this script `$(zfs get -Ho value dlpx:ashift domain0)` fails silently and `ashift` is gets set to 0 because of bash default values. Then we'd inadvertently change `ashift` to 0.

PR #220, opened 2020-05-12
Use "delphix/actions" for shellcheck and shfmt

prakashsurya:


prakashsurya:
This is depends on: https://github.com/delphix/actions/pull/3 and https://github.com/delphix/actions/pull/4

PR #225, opened 2020-05-28
DLPX-69937 [Backport of DLPX-69870 to 6.0.3.0] Upgrade verification fails on GCP because google-instance-setup.service hangs inside container

sanbhara:
google-instance-setup.service hangs within a systemd-nspawn container. This service is unnecessary for the upgrade container. Solution is to add an override parameter that adds a "!container" condition to the service execution.

git ab-pre-pusht: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3530/

### DIFF
--- a/files/common/lib/systemd/system/google-instance-setup.service.d/override.conf
+++ b/files/common/lib/systemd/system/google-instance-setup.service.d/override.conf
@@ -1,0 +1,11 @@
+#
+# During upgrade verification, the root filesystem will be booted as a
+# container using systemd-nspawn. Engines running on GCP have the
+# google-instance-setup service enabled by default. When this service
+# runs in a container, it fails to activate and cause the start container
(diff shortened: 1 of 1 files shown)


prakashsurya:
This is a backport, so I'd rather not add differences between master and our 6.0 branches. If we want to modify the solution we used on master, we should do that on master, and then backport that to 6.0.

With that said, I think it's fine for this file to reside on all platforms, even if we only install that package on GCP; I don't think it'll have any effect on non-GCP platforms.

pzakha:
Ah okay, I didn't notice it was a backport. As long as it doesn't cause any unwanted behaviour it shouldn't be a big deal.

sanbhara:
@pzakha ya I ran ab-pre-push on the master PR and that was clean so this override file should be harmless if the underlying service does not exist on a platform.

PR #230, opened 2020-06-11
DLPX-70300 Upgrade verification hangs on GCP because systemd-logind service gets into a restart loop within the container

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3647/)

prakashsurya:
@sanbhara FWIW, I tested this manually, by creating the override files, restarting the host, and then creating and starting a container. I welcome any additional testing you want to do.

We might also want to verify that upgrading an existing system will work, since any existing system will not start with these override files on the host, when the container is initially created and started; I haven't yet tested this.

sanbhara:
@prakashsurya I do not expect this fix will allow us to upgrade an existing system that does not have this patch. If for some reason, a customer chooses to deploy an engine on a version that does not have this fix, that engine will need to be hotfixed in order to upgrade.

I plan to go through the upgrade process manually since we do not seem to run upgrade-testing in ab-pre-push if the platform chosen is anything but AWS or ESX.

Let me know if I am missing something in my understanding of the fix.

prakashsurya:
Yea, thinking a little more, I agree with all of that.

sebroy:
Keep in mind that the hangs were were seeing may have been unique to our internal-dev variant's use of LDAP (which our customers can't configure). This issue may not be as severe on a real internal-qa or external variant.

prakashsurya:
@sebroy I think you're right. I tried this workflow on a QA variant image, and the container started up just fine the first time. So, this does appear to be an internal-dev issue only.

sanbhara:
That's great to know. FWIW I tested out verification with the fix and it went through fine.

PR #232, opened 2020-06-18
DLPX-70614 [Backport of DLPX-70512 to 6.0.3.0] Long zio delays likely caused by Linux IO scheduler write throttling logic

pzakha:
Disable Linux IO scheduler write throttle logic for all disks.

## Testing
- Manually verified on a system that the tunable was applied and that write throttle was effectively disabled (wb_normal = 0):
```
$ for lat in /sys/block/sd*/queue/wbt_lat_usec; do echo "$lat: $(cat $lat)"; done
/sys/block/sda/queue/wbt_lat_usec: 0
/sys/block/sdb/queue/wbt_lat_usec: 0
/sys/block/sdc/queue/wbt_lat_usec: 0
/sys/block/sdd/queue/wbt_lat_usec: 0

$ sudo bash -c 'for norm in /sys/kernel/debug/block/sd*/rqos/wbt/wb_normal; do echo "$norm: $(cat $norm)"; done'
/sys/kernel/debug/block/sda/rqos/wbt/wb_normal: 0
/sys/kernel/debug/block/sdb/rqos/wbt/wb_normal: 0
/sys/kernel/debug/block/sdc/rqos/wbt/wb_normal: 0
/sys/kernel/debug/block/sdd/rqos/wbt/wb_normal: 0
```
- ab-pre-push (esx): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3691/
- ab-pre-push (aws): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3692/

### DIFF
--- a/files/common/lib/udev/rules.d/60-dlpx-schedulers.rules
+++ b/files/common/lib/udev/rules.d/60-dlpx-schedulers.rules
@@ -1,0 +1,6 @@
+#
+# Delphix: this setting disables the write throttle algorithm that is used in
+# the Linux IO scheduler for all block devices as we've seen problems with it.
+# We explicitly disable it for zvols as it doesn't seem to apply to them anyway.
+#
+ACTION=="add|change", SUBSYSTEM=="block", KERNEL!="zd*", ATTR{queue/wbt_lat_usec}="0"


files/common/lib/udev/rules.d/60-dlpx-schedulers.rules
```diff
@@ -0,0 +1,5 @@
+#
+# Delphix: this setting disables the write throttle algorithm that is used in
+# the Linux IO scheduler as we've seen problems with it.
+#
+ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="sd*", ATTR{queue/wbt_lat_usec}="0"
```
sebroy, line 5: I wonder if this set of matching rules could be inadvertently excluding some block devices that we care about. I think, Pavel, you're going to verify that and adjust if needed.

grwilson: Yeah, we probably need too do something like `KERNEL=="sd*|xvd*|virtio*'

pzakha: Yeah, I've tested this and it definitely doesn't apply to xvd/nvme/etc.
I'm considering removing KERNEL altogether, it seems to work. The downside is that it applies to other block devices as well, such as `loop` and `sr` and `zd`. It probably doesn't matter much for `loop` and `sr`, but I think for `zd` we'll probably want to exclude it, given the following behaviour:
```
$ cat /sys/block/zd0/queue/wbt_lat_usec
cat: /sys/block/zd0/queue/wbt_lat_usec: Invalid argument
$ sudo bash -c 'echo 0 >/sys/block/zd0/queue/wbt_lat_usec'
$ cat /sys/block/zd0/queue/wbt_lat_usec
0
```
I guess the safer way would be to selectively include, at the risk of missing new ones, rather than do a selective exclude and then accidentally mess with devices that we shouldn't be messing with.

sebroy: The impact of missing new ones would be rather huge (system-wide hang on a customer system at some point in the future requiring a patch release). What's the impact of killing wbt across the board? The kernel allows one to compile it out of the system altogether (via `CONFIG_BLK_WBT`), so I'm guessing this is not disastrous in theory.

pzakha: I don't know what's the impact of echoing `0` for zvols, which had `cat /sys/block/zd0/queue/wbt_lat_usec` return EINVAL. To be safe, I could set it across the board, but specifically exclude zvols.

pzakha:
## Retesting latest code

- ab-pre-push (aws): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3693/
- ab-pre-push (esx): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3694/

I've created a zvol on Azure, here's the output:
```
$ for lat in /sys/block/*/queue/wbt_lat_usec; do echo "$lat: $(cat $lat)"; done
/sys/block/loop0/queue/wbt_lat_usec: 0
/sys/block/loop1/queue/wbt_lat_usec: 0
/sys/block/loop2/queue/wbt_lat_usec: 0
/sys/block/loop3/queue/wbt_lat_usec: 0
/sys/block/loop4/queue/wbt_lat_usec: 0
/sys/block/loop5/queue/wbt_lat_usec: 0
/sys/block/loop6/queue/wbt_lat_usec: 0
/sys/block/loop7/queue/wbt_lat_usec: 0
/sys/block/sda/queue/wbt_lat_usec: 0
/sys/block/sdb/queue/wbt_lat_usec: 0
/sys/block/sdc/queue/wbt_lat_usec: 0
/sys/block/sdd/queue/wbt_lat_usec: 0
/sys/block/sde/queue/wbt_lat_usec: 0
/sys/block/sr0/queue/wbt_lat_usec: 0
cat: /sys/block/zd0/queue/wbt_lat_usec: Invalid argument
/sys/block/zd0/queue/wbt_lat_usec:
```

Here's on gcp:
```
$ for lat in /sys/block/*/queue/wbt_lat_usec; do echo "$lat: $(cat $lat)"; done
/sys/block/loop0/queue/wbt_lat_usec: 0
/sys/block/loop1/queue/wbt_lat_usec: 0
/sys/block/loop2/queue/wbt_lat_usec: 0
/sys/block/loop3/queue/wbt_lat_usec: 0
/sys/block/loop4/queue/wbt_lat_usec: 0
/sys/block/loop5/queue/wbt_lat_usec: 0
/sys/block/loop6/queue/wbt_lat_usec: 0
/sys/block/loop7/queue/wbt_lat_usec: 0
/sys/block/nvme0n1/queue/wbt_lat_usec: 0
/sys/block/nvme1n1/queue/wbt_lat_usec: 0
/sys/block/nvme2n1/queue/wbt_lat_usec: 0
/sys/block/nvme3n1/queue/wbt_lat_usec: 0
cat: /sys/block/zd0/queue/wbt_lat_usec: Invalid argument
/sys/block/zd0/queue/wbt_lat_usec:
```

Note that we get `cat: /sys/block/zd0/queue/wbt_lat_usec: Invalid argument` as `wbt_lat_usec` is undefined by default on zvols, unless we explicitly set it via udev rules or by first writing to `/sys/block/zd0/queue/wbt_lat_usec`.

PR #234, opened 2020-07-08
Add recovery environment dependency to delphix-platform

pcd1193182:
Small change to add the recovery environment to delphix-platform. ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3727/

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -104,6 +104,12 @@ DEPENDS += delphix-buildinfo-kernel,
 DEPENDS += pam-challenge-response, \
 	   pam-challenge-response-dbgsym,
 
+#
+# The recovery environment is required for the product to be debuggable in
+# worst-case scenarios.
+#
+DEPENDS += recovery-environment,
+
 # Platform-specific dependencies
 DEPENDS.aws = nvme-cli,
 DEPENDS.azure = walinuxagent, linux-cloud-tools-azure,


debian/rules
```diff
@@ -104,6 +104,12 @@ DEPENDS += delphix-buildinfo-kernel,
 DEPENDS += pam-challenge-response, \
 	   pam-challenge-response-dbgsym,
 
+#
+# The recovery environment is required for the product to be debuggable in
+# worst-case scenarios.
+#
+DEPENDS += recovery-environment
```
pzakha, line 111: This will create a circular dependency since the recovery environment depends on delphix-platform: https://github.com/delphix/recovery-environment/blob/master/debian/control#L12

pzakha: Also it seems like you are missing a `comma` at the end of the line `DEPENDS += recovery-environment `.

I believe we've added the depend on `delphix-platform` because the recovery environment needed some of the files that were present in there. I think we may need to rethink how we pull-in some of the package dependencies into the product.

pcd1193182: Yeah, we have it as a dependency because debian/postinst copies a bunch that we get from other packages. Maybe we can modify the recovery-environment depends to be more specific, so it only depends on the specific packages it's copying things from; the kernel, the networking config, etc. Or, alternatively, can we have whatever part of the build process says "install delphix-platform and its dependencies on the VM" also install recovery-environment. Preferences?

prakashsurya: Why is the circular dependency an problem?

pzakha: Yeah, so ideally we'd have the recovery environment depend on the specific packages that it cares about.
> Or, alternatively, can we have whatever part of the build process says "install delphix-platform and its dependencies on the VM" also install recovery-environment. Preferences?
This would also require some unsavory changes in the upgrade logic, which ideally we wouldn't mess with. 

Also another thing we may want to consider is what happens when we do an upgrade. If say we upgrade both `delphix-kernel` and the `recovery-environment` package, as part of the Delphix upgrade process, I don't know if there is an ordering guarantee that it will first upgrade delphix-kernel and THEN upgrade the recovery-environment. The `Depends` section guarantees that some version of `delphix-kernel` will be there when the recovery-environment package is installed (and postinst script is run), but I'm not sure about the rest.

prakashsurya: > I don't know if there is an ordering guarantee ...

Why is the order important? What order do we require?

pzakha: > Why is the circular dependency an problem?

This could either break install/upgrade or have apt arbitrarily pick a point where to break the cyclic dependency, potentially installing `recovery-environment` before `delphix-platform` and thus breaking its install.

pzakha: > > I don't know if there is an ordering guarantee ...
> 
> Why is the order important? What order do we require?

Because of how the postinst script copies files from the dependencies: https://github.com/delphix/recovery-environment/blob/master/debian/postinst
It would change which version of the files is copied.

prakashsurya: > This could either break install/upgrade or have apt arbitrarily pick a point where to break the cyclic dependency, potentially installing recovery-environment before delphix-platform and thus breaking its install.

Can you provide some details on how this would occur? I'm not sure how exactly that'd occur for our use case(s) (breaking the installation and/or upgrade).. Just trying to understand..

IIRC, the dependency graphs is determined when "apt-get install" is invoked.. so even though at some point while "apt-get install" is running (i.e. when packages are being removed, installed, etc. by apt-get), the graph might not be coherent, I'm not sure if that's a problem; the end result will always be a coherent graph of package dependencies (assuming "apt-get install" exits successfully).

> Because of how the postinst script copies files from the dependencies

IIRC, all packages in a given "apt-get install" invocation are installed first, and then each of those package's "postinst" script runs after all binaries are installed/updated..

so, for example, on upgrade when we "apt-get install delphix-platform", we'll install delphix-platform and all dependencies, and only run the "postinst" hooks after all packages are installed.. which I think is what we want.. i.e. we'll run "postinst" after the proper binaries are in place.

Please correct me if I'm wrong, though.. I'm assuming we don't care about the order in which each "postinst" hook for each package is run.. and we only care that the recovery environment's "postinst" hook runs after all new binaries are installed in the filesystem..

prakashsurya: While I'm not 100% sure of this, I think the package hook execution order follows the order of dependencies.. so, for example, if the "recovery-environment" depends on "delphix-platform", and "delphix-platform" depends on "delphix-kernel".. the "postinst" hooks for each package will follow the order of: first "delphix-kernel", then "delphix-platform", then "recovery-environment". I'd have to do some testing and/or reading of the source code to double check this, though.

pzakha: > have apt arbitrarily pick a point where to break the cyclic dependency

Actually now that I think more about it I'm not sure about this anymore. I do recall that this is how systemd would address cyclical dependencies, but I'm not as sure about apt.

With a quick search online regarding cyclical dependencies with apt people report installation / upgrade issues because of it. There does seem to be ways to address them (e.g. if A depends on B and B depends on A, then doing `apt install A B`, or `apt install -f`) but I think in general if we can avoid them it would make us less likely to fall into dependency hell down the road.

> Please correct me if I'm wrong, though.. I'm assuming we don't care about the order in which each "postinst" hook for each package is run.. and we only care that the recovery environment's "postinst" hook runs after all new binaries are installed in the filesystem..

I think you are right regarding the `postinst` scripts, usually the files for the packages are deployed first and then the `postinst` scripts for the packages are run as part of the "Config" step. There might perhaps be some exceptions if there are `Pre-Depends` dependencies but in general that's likely the case. Something we'd want to verify though.

PR #241, opened 2020-07-29
Automatically determine image variant if not specified

prakashsurya:
This change optimizes the UX of "download-latest-image" script, such
that the user does not necessarily have to specify the image variant to
be downloaded.

This is useful for the case where the script is run on an existing
Delphix appliance, and the user wants the script to download the image
variant that matches the appliance it's being run on.

In this case, the script can (and now does, with this change) detect the
variant of the appliance, and download the corresponding image, without
forcing the user to specify which variant to download.

### DIFF
--- a/files/common/usr/bin/download-latest-image
+++ b/files/common/usr/bin/download-latest-image
@@ -22,7 +22,7 @@ function die() {
 
 function usage() {
 	echo "$(basename "$0"): $*" >&2
-	echo "Usage: $(basename "$0") [-f] [-N] [-b branch] <variant>"
+	echo "Usage: $(basename "$0") [-f] [-N] [-b branch] [variant]"
 	exit 2
 }
 
@@ -43,10 +43,18 @@ done
 shift $((OPTIND - 1))
 
 [[ $# -gt 1 ]] && usage "too many arguments specified"
-[[ $# -eq 0 ]] && usage "too few arguments specified"
 
 $opt_f && rm -f "latest" >/dev/null 2>&1
 [[ -f "latest" ]] && die "file 'latest' already exists"
+
+VARIANT="$1"
+if [[ -z "$VARIANT" ]]; then
+	PLATFORM=$(cat "/var/lib/delphix-appliance/platform")
+	[[ -n "$PLATFORM" ]] || die "platform could not be determined"
+
+	VARIANT=$(cat "/usr/share/doc/delphix-entire-$PLATFORM/variant")
+	[[ -n "$VARIANT" ]] || die "variant could not be determined"
+fi
 
 #
 # We don't want to delete the "latest" file if it already exists and the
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3796/)

grwilson (approved):
This is great! Can you also backport this?

prakashsurya:
@grwilson sure, see here: #254

files/common/usr/bin/download-latest-image
```diff
@@ -43,11 +43,21 @@ done
...
 $opt_f && rm -f "latest" >/dev/null 2>&1
 [[ -f "latest" ]] && die "file 'latest' already exists"
 
+VARIANT="$1"
+if [[ -z "$VARIANT" ]]; then
+	PLATFORM=$(cat "/var/lib/delphix-appliance/platform")
+	[[ -n "$PLATFORM" ]] || die "platform could not be determined"
+
+	VARIANT=$(cat "/usr/share/doc/delphix-entire-$PLATFORM/variant")
+	[[ -n "$PLATFORM" ]] || die "variant could not be determined"
```
pzakha, line 56: copy-paste typo: `-n "$PLATFORM"`

pzakha, line 56: Actually, I've been thinking it would be very convenient to have a short helper script like `/usr/bin/get-appliance-variant` that would do this. I think it would be especially useful for folks outside of the systems platform team as remembering the path of the variant file is not as trivial.

files/common/usr/bin/download-latest-image
```diff
@@ -43,11 +43,21 @@ done
...
+fi
+
+[[ -n "$VARIANT" ]] || die "upgrade image variant is invalid: '$VARIANT'"
```
pzakha, line 59: This line seems redundant.

files/common/usr/bin/download-latest-image
```diff
@@ -69,8 +79,8 @@ aws s3 cp \
...
 	. || die "failed to download file: 'latest'"
 
-$opt_f && rm -f "$1.upgrade.tar" >/dev/null 2>&1
-[[ -f "$1.upgrade.tar" ]] && die "image $1.upgrade.tar already exists"
+$opt_f && rm -f "$VARIANT.upgrade.tar" >/dev/null 2>&1
+[[ -f "$VARIANT.upgrade.tar" ]] && die "image $VARIANT.upgrade.tar already exists"
 
-aws s3 cp "s3://snapshot-de-images/$(cat latest)/upgrade-artifacts/$1.upgrade.tar" . ||
-	die "failed to download file: '$1.upgrade.tar'"
+aws s3 cp "s3://snapshot-de-images/$(cat latest)/upgrade-artifacts/$VARIANT.upgrade.tar" . ||
```
pzakha, line 85: nit: might be useful to echo the full S3 path for debugging/informational purposes.

pzakha (approved):

files/common/usr/bin/download-latest-image
prakashsurya: I disagree.

files/common/usr/bin/download-latest-image
prakashsurya: Sure, can you open a bug for this suggestion?

pzakha: I can, but I figured you might as well do it now as it would take 2 minutes and avoid all the bureaucracy of creating a new bug, then a separate PR, etc.

files/common/usr/bin/download-latest-image
prakashsurya: I can remove this.

files/common/usr/bin/download-latest-image
prakashsurya: Well, I'm not convinced that exposing the variant is something we should be doing. I can't recall when I needed that information, outside of this script, and I prefer to expose as little information to external consumers as is required. So, without more information, I can't say that I agree that making the variant easier to retrieve is a universally good thing to do.

Which is why I think you'd be the better person to open the bug and make the argument for creating such a script.

pzakha: okay, feel free to drop this then.

PR #245, opened 2020-08-27
DLPX-71608 Remove delphix-buildinfo packages as they will no longer be built

pzakha:
With the work done in TOOL-9778, each package will be built individually and we will no longer build delphix-buildinfo-userland and delphix-buildinfo-kernel packages, so those have to be removed from the delphix-platform's package dependencies.

Those packages are not currently used by the product, but are useful for QA/developers to figure out what git commit was used to build packages that are installed on the appliance.

We will need to think of an alternative way to get the buildinfo onto the engine as it was quite useful, and I've proposed an idea in DLPX-71683.

## Testing
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3935/


pzakha:
> Does this mean we can no longer map the software that's installed on the system, to the git repository information that was use to build that software?

That's right. I've updated the description.

brad-lewis (approved):

PR #249, opened 2020-10-19
Move various configuration files out of "/etc"

prakashsurya:
This change moves various files that were previously installed into
"/etc" (for no good reason) out of "/etc". The problem of installing
into "/etc" is, these files are than treated differently by the package
manager on upgrade due to them being considered "conffiles".

Thus, to simplify the semantics of how these files will be handled on
upgrade, it's best to install into directories other than "/etc" when
possible, which is what this change does.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4170/)

prakashsurya:
List of confiles before this change:
```
$ cat conffiles
/etc/cloud/cloud.cfg.d/91-delphix-platform.cfg
/etc/cron.daily/delphix-upgrade-logs
/etc/default/grub.d/override.cfg
/etc/modprobe.d/10-lockd.conf
/etc/modprobe.d/10-zfs.conf
/etc/modprobe.d/blacklist-delphix.conf
/etc/sudoers.d/delphix
/etc/sysctl.d/30-nfsv3-ports.conf
/etc/systemd/journald.conf.d/override.conf
/etc/systemd/network/80-container-host0.network.d/override.conf
/etc/systemd/network/80-container-ve.network.d/override.conf
/etc/systemd/system.conf.d/override.conf
/etc/systemd/system/console-setup.service.d/override.conf
/etc/systemd/system/systemd-networkd.service.d/override.conf
/etc/udev/rules.d/999-aws-ebs-nvme.rules
/etc/udev/rules.d/999-delphix-systemd.rules
/etc/update-motd.d/00-delphix
```

List of conffiles after this change:
```
$ cat conffiles
/etc/cloud/cloud.cfg.d/91-delphix-platform.cfg
/etc/cron.daily/delphix-upgrade-logs
/etc/default/grub.d/override.cfg
/etc/sudoers.d/delphix
/etc/update-motd.d/00-delphix
```

pzakha (approved):
Change LGTM.
Have you verified that all of those new locations are indeed ingested by the system? I've noticed some of them go into `/lib`, others to `/usr/lib`; I know the configuration managers have precedence rules for which locations have priority over other locations (with /etc usually at the top priority), so I think it would be worthwhile to verify that the new locations are not overridden by other config files on the system.

prakashsurya:
I haven't done that yet, but it's on my list of things to do before landing this.

prakashsurya:
I've verified all the files still take effect in their new location.

grwilson (approved):

PR #260, opened 2021-01-12
DLPX-73666 file "/var/run/reboot-required" is not properly generated on DEFERRED upgrades

prakashsurya:
This is a simple change to add the "update-notifier-common" package as a
dependency. This package is required for the "/var/lib/reboot-required"
file to be properly generated during a upgrade of the appliance.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -60,7 +60,8 @@ DEPENDS += ansible, \
 	   python3, \
 	   rng-tools, \
 	   systemd-container, \
-	   ubuntu-minimal,
+	   ubuntu-minimal, \
+	   update-notifier-common,
 
 # Debugging symbols for packages pulled in by the the above dependencies
 DEPENDS += systemd-dbgsym, \


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4582/)

pzakha (approved):
We might also want to check that there are no unwanted side effects from installing this package. For instance, looking at the list of files installed we have:
- New conf files in /etc/apt/apt.conf.d
- New cron files /etc/cron.daily/update-notifier-common and /etc/cron.weekly/update-notifier-common
- New files in /etc/update-motd.d

sebroy (approved):

sanbhara (commented):
Do we want to update the dbus.postinst script to use notify-reboot-required as well? Or is that file in a different repo?

prakashsurya:
The dbus package should work fine with or without this change; I don't see any need to modify it.

prakashsurya:
> Or is that file in a different repo?

For completeness, that would be a change to a different repository; but as I said, I don't think it's necessary.

sanbhara (approved):

pzakha:
@prakashsurya have you checked on the concerns I've raised in my comment above and how that could affect the product?

prakashsurya:
@pzakha I did, I didn't find anything concerning.

PR #263, opened 2021-01-26
Move open-iscsi override out of /etc

don-brady:
## Description
We want the new `open-iscsi.service.d/override.conf` file (from #262) to reside under `/lib` not `/etc`

See PR #249 --  Move various configuration files out of "/etc"

## Testing
ab-pre-push
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4670/

Confirmed that the override file now lives under `/lib/systemd/system/`
Confirmed that `systemctl cat open-iscsi` contains the override content


prakashsurya (approved):

pzakha (approved):
Good catch.

PR #264, opened 2021-02-04
DLPX-68851 auditd service fails within upgrade container

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4719/)

prakashsurya:
This is a partial solution, which also needs the following app-gate change: http://reviews.delphix.com/r/65576/

prakashsurya:
Additionally, I think this solution may be insufficient as it stands. For systems with the auditd override already in place in `/etc`, we'll need to first remove that file (and run `systemctl daemon-reload`) prior to upgrading the auditd package.

Currently, with the proposed solution here (including the app-gate change), we wouldn't attempt to remove the old configuration in `/etc` until after we attempt to upgrade auditd; so upgrade will have failed before we have a chance to fix the configuration.

I believe a full solution here would architecturally benefit from the "upgrade healer" concept we've discussed in the past, where we could "fix" the existing system, prior to attempting the upgrade. Since we don't have a good mechanism for doing this in today's upgrade architecture, we'll likely have to add a "hack", where we remove the file in `/etc` via upgrade's `execute` script (i.e. an additional change to appliance-builds upgrade script) or perhaps add a `preinst` hook for this platform package which removes the file in `/etc` (and run `systemctl daemon-reload`).

prakashsurya:
I've updated the app-gate portion of this fix to handle existing systems by adding the following commands to the "preinst" hook of the delphix-virtualization package:
{noformat}
rm -f /etc/systemd/system/auditd.service.d/override.conf
systemctl daemon-reload
{noformat}
With that in place, we should be good to land this and the app-gate portion (tests are running to verify).

sanbhara (approved):

don-brady (approved):

PR #266, opened 2021-02-25
Fix several issues with delphix console screen

grwilson:
DLPX-71998 Chatty systemd events for getty, serial-getty services take up a large percentage of syslog
DLPX-72681 delphix-startup-screen fails with static IP address
DLPX-73286 systemd is restarting locale service every 30 seconds
DLPX-73423 delphix-startup-screen crashes if there's no default route

This changes the getty services to launch the startups screen directly and it's the startup screen that invokes login. This prevents the need for the getty service to restart itself frequently and now the localed and getty restarts only happen when the login session is completed. If nobody logs in via the console then we don't need to restart the service. A further enhancement would be to not display the keyboard layout on the main screen. This would prevent localed from starting unless you change the keyboard layout.

### DIFF
--- a/files/common/usr/bin/delphix-startup-screen
+++ b/files/common/usr/bin/delphix-startup-screen
@@ -23,10 +23,12 @@ import curses
 import curses.ascii
 import curses.textpad
 import sys
+import signal
 import os
 import logging
 import subprocess
 from typing import List, Any, Tuple, Generator
+from netifaces import interfaces, ifaddresses, AF_INET
 
 STATUS_HEADER: str = "STATUS\t\t\tSERVICE\t\t\t"
 LAYOUT_STR: str = "Keyboard layout (press 9 to change):"
@@ -201,6 +203,11 @@ def getstatus() -> str:
     status = ""
     for i in getstatus_pair(list(run_svcs("Description"))):
         #
+        # Don't print any inactive services
+        #
+        if i[-1].strip() == "inactive":
+            continue
+        #
         # Switch the state and description
         #
         status += f"{i[-1]}\t{i[0]}\n"
@@ -209,22 +216,17 @@ def getstatus() -> str:
 
 def get_network_status() -> Tuple[str, str]:
     """
-    Returns a tuple of (hostname, ipaddr) for the main interface.
+    Returns a tuple of (hostname, ipaddrs).
     """
 
-    #
-    # We get the primary address based on the default route. The
-    # address we use follows the 'src' key in the output.
-    #
-    cmd = ['ip', 'route', 'show', '0.0.0.0/0']
-    cp = subprocess.run(cmd,
-                        stdout=subprocess.PIPE,
-                        universal_newlines=True,
-                        check=True)
-    output = cp.stdout.split()
-    ipaddr = output[output.index("src") + 1]
+    ipaddrs = []
+    for interface in interfaces():
+        if interface == "lo":
+            continue
+        for link in ifaddresses(interface)[AF_INET]:
+            ipaddrs.append(link['addr'])
     hostname = os.uname()[1]
-    return (hostname, ipaddr)
+    return (hostname, ", ".join(ipaddrs))
 
 
 # pylint: disable-msg=too-many-locals
@@ -239,11 +241,11 @@ def display_status(stdscr, win):
     layout = get_keyboard_layout()
 
     width = X - 10
-    height = Y - 8
+    height = Y - 6
     status = 0
 
     x = int((X - width) / 2) + 2
-    y = (Y - height - 1)
+    y = (Y - height)
 
     win.clear()
     win.box()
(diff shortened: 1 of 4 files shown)


grwilson:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4828/

files/common/usr/bin/delphix-startup-screen
```diff
@@ -272,7 +274,9 @@ def display_status(stdscr, win):
 
         (hostname, ipaddr) = get_network_status()
         netwin.clear()
-        netwin.addstr(0, 0, "Host: http://" + hostname + "/", curses.A_BOLD)
+        if hostname:
+            hostname = "http://" + hostname
```
pzakha, line 278: nit: should we show `https` instead of `http` here?

files/common/usr/bin/delphix-startup-screen
```diff
@@ -272,7 +274,9 @@ def display_status(stdscr, win):
 
         (hostname, ipaddr) = get_network_status()
```
pzakha, line 275: nit: ipaddr -> ipaddrs

files/common/usr/bin/delphix-startup-screen
```diff
@@ -272,7 +274,9 @@ def display_status(stdscr, win):
...
         netwin.clear()
-        netwin.addstr(0, 0, "Host: http://" + hostname + "/", curses.A_BOLD)
+        if hostname:
+            hostname = "http://" + hostname
+        netwin.addstr(0, 0, "Host: " + hostname, curses.A_BOLD)
         netwin.addstr(1, 0, "IP: " + ipaddr, curses.A_BOLD)
```
pzakha, line 280: nit: IP -> IPs; ipaddr -> ipaddrs

files/common/usr/bin/delphix-startup-screen
```diff
@@ -390,5 +394,11 @@ def installer_main(stdscr):
 
 if __name__ == '__main__':
     logging.basicConfig(filename=log_file, level=logging.DEBUG)
-
+    signal.signal(signal.SIGTTOU, signal.SIG_IGN)
+    signal.signal(signal.SIGTSTP, signal.SIG_IGN)
+    signal.signal(signal.SIGHUP, signal.SIG_IGN)
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
+    signal.signal(signal.SIGQUIT, signal.SIG_IGN)
```
pzakha, line 401: Can you comment on why we ignore those signals in particular?

grwilson:
If you think we don't need existing keyboard layout then I can remove it right now. It's a real simple change.

grwilson:
Screen Shot 2021-02-25 at 11 00 15 AM

files/common/usr/bin/delphix-startup-screen
grwilson: We are ignoring them so that you can't quit the login process when we exec it.

grwilson:
Most of my testing has been on ESX and azure where the console renders correctly. 

- I have confirmed that the screen will display the login prompt for 5 minutes and if nobody logs in then it will redisplay the status screen again. 
- I have confirmed that messages have been greatly reduced  since we no longer restart service frequently
-  confirmed that localed only restarts when a login process terminates

pzakha:
> I have confirmed that messages have been greatly reduced since we no longer restart service frequently
confirmed that localed only restarts when a login process terminates

Is there other automation that start a login process?

files/common/usr/bin/delphix-startup-screen
pzakha: Who would sent those signals normally? if someone presses enter to login to the console and then changes their mind and presses ctrl-C, what would happen before and after this change?

pzakha:
FYI, it seems like pylint has found a few issues: https://github.com/delphix/delphix-platform/pull/266/checks?check_run_id=1980352590

files/common/usr/bin/delphix-startup-screen
grwilson: Before if you press ctrl-C then it just prints on the screen. If we don't ignore the signal then the python process exits and restarts. So ignoring these signals allow the login behavior to stay the same.

pzakha: Got it, thanks, could you add this info as a comment in the code?

pzakha (approved):

grwilson:
> > I have confirmed that messages have been greatly reduced since we no longer restart service frequently
> > confirmed that localed only restarts when a login process terminates
> 
> Is there other automation that start a login process?

The login process will only be invoked by the startup screen and that will get automatically restarted by the getty service. I will verify that single-user logins are not impacted.

files/common/usr/bin/delphix-startup-screen
```diff
@@ -209,22 +216,17 @@ def getstatus() -> str:
 
 def get_network_status() -> Tuple[str, str]:
     """
-    Returns a tuple of (hostname, ipaddr) for the main interface.
+    Returns a tuple of (hostname, ipaddrs) for the main interface.
```
sebroy, line 219: comment nit: we're not considering any interface as special anymore

sebroy (commented):
What happens if the delphix-startup-screen process dies with a python stack trace? Does it get automatically re-started?

grwilson:
> What happens if the delphix-startup-screen process dies with a python stack trace? Does it get automatically re-started?

Yes, it gets restarted. One somewhat negative side effect is that a bug in the screen does prevent login from being exec-ed. I wonder if I could catch all errors and ignore them so that we can always get a login prompt. I will look at that.

sebroy (approved):

PR #274, opened 2021-03-18
DLPX-74854 Update /usr/bin/get-appliance-version to also report the h…

matt-skinner-delphix:
## Pull request checklist

- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Build was run locally and any changes were pushed

[Test run](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4946/)

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Current script does not report the hotfix version. 
Issue Number: DLPX-74854

## What is the new behavior?

Due to changes in https://github.com/delphix/appliance-build/pull/507 and with ongoing work in hotfixes, the version is stored in a ZFS property. For the sake of consistency the hotfix version should be reported in this tool. Because we are not changing the Delphix version string, the hotfix version is not included in the version string, but is a separate value.

- [ ] Yes
- [x] No

### DIFF
--- a/files/common/usr/bin/get-hotfix-version
+++ b/files/common/usr/bin/get-hotfix-version
@@ -1,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright 2021 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -o errexit
+set -o pipefail
+
+function usage() {
+	echo "Usage: $(basename "$0")"
+	echo
+	echo "Display the hotfix version information for the appliance."
+	exit 2
+}
+
+if [[ $# -gt 0 ]]; then
+	echo "No arguments are supported."
+	usage
+fi
+
+output=$(zfs get -Hpo value "com.delphix:hotfix-version" "$(dirname "$(zfs list -Hpo name /)")")
+if [[ "$output" != "-" ]]; then
+	echo -n "$output"
+fi


prakashsurya (commented):
Since the hotfix-version field is completely separate from the current-version field, and the implementation of retrieving and parsing is independent, how would you feel about creating a new script (e.g. `get-hotfix-version`) to retrieve the hotfix version?

I might be able to be convinced otherwise, but I sort of feel that the two versions are distinct enough, that perhaps using a separate script might be more appropriate. This might also make it a little more apparent when reading consumer logic, that we're getting the hotfix version; e.g. we'd see `get-hotfix-version` rather than `get-appliance-version -f`, which to me is easier to immediately see what "version" we're retrieving.

What do you think?

matt-skinner-delphix:
@prakashsurya I am open to a new script. I'll wait a little before proceeding to see if others have other ideas and opinions.

files/common/usr/bin/get-hotfix-version
```diff
@@ -0,0 +1,40 @@
...
+set -o pipefail
+
+function usage() {
+	echo "Usage: $(basename "$0")"
+	echo
+	echo "Display the hotfix version information for the appliance."
+	exit 2
+}
+
+function hotfix_version() {
```
prakashsurya, line 28: Lets remove this function and just inline it at the bottom where this is currently called. Since this function is only called in one place, I don't see much benefit to using a separate function (and I don't foresee any future changes where we'd want the separate function).

files/common/usr/bin/get-hotfix-version
```diff
@@ -0,0 +1,40 @@
...
+	zfs get -Hpo value "com.delphix:hotfix-version" "$(dirname "$(zfs list -Hpo name /)")" | tr -d '\n'
+}
+
+options=$(getopt -l "" -o "" -- "$@")
+eval set -- "$options"
```
prakashsurya, line 33: What benefit do we gain by using `getopt` here? Since we don't support options, would it be more simple to remove this part?

files/common/usr/bin/get-hotfix-version
```diff
@@ -0,0 +1,40 @@
...
+
+function usage() {
+	echo "Usage: $(basename "$0")"
+	echo
+	echo "Display the hotfix version information for the appliance."
+	exit 2
+}
+
+function hotfix_version() {
+	zfs get -Hpo value "com.delphix:hotfix-version" "$(dirname "$(zfs list -Hpo name /)")" | tr -d '\n'
```
prakashsurya, line 29: When the hotfix-version field doesn't exist (e.g. no hotfix is present), what do we want to return?

matt-skinner-delphix: Fixed to return nothing.

files/common/usr/bin/get-hotfix-version
```diff
@@ -0,0 +1,44 @@
...
+	echo "Display the hotfix version information for the appliance."
+	exit 2
+}
+
+function hotfix_version() {
+  local output
+	output=$(zfs get -Hpo value "com.delphix:hotfix-version" "$(dirname "$(zfs list -Hpo name /)")" | tr -d '\n')
+	if [[ "$output" != "-" ]]; then
+	   echo "$output"
+	fi
```
prakashsurya: Not sure if you think this would be "too clever" and make the code more difficult to understand, but we could accomplish by adding `sed 's/^-$//'` to the pipeline.. E.g.
```
#
# When the "hotfix-version" property doesn't exist (e.g. when no hotfix has been applied), we want to return the
# empty string. We accomplish this by replacing the "-" string returned by "zfs get" with the empty string.
#
zfs get -Hpo value "com.delphix:hotfix-version" "$(dirname "$(zfs list -Hpo name /)")" | sed 's/^-$//' | tr -d '\n'
```

The comment helps make it clear what we're doing, and then we have one less variable and conditional; which makes it more simple, IMO.

I'm good with what you have, but just wanted to offer a potential alternative solution.

files/common/usr/bin/get-hotfix-version
```diff
@@ -0,0 +1,41 @@
...
+	echo
+	echo "Display the hotfix version information for the appliance."
+	exit 2
+}
+
+function hotfix_version() {
+  local output
+	output=$(zfs get -Hpo value "com.delphix:hotfix-version" "$(dirname "$(zfs list -Hpo name /)")")
+	if [[ "$output" != "-" ]]; then
+	   echo "$output" | tr -d '\n'
```
prakashsurya, line 32: Since we're using `echo` here, can we use `echo -n` and drop the use of `tr`?

prakashsurya (approved):

files/common/usr/bin/get-hotfix-version
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+#
+# Copyright 2020 Delphix
```
prakashsurya, line 3: 2021

files/common/usr/bin/get-hotfix-version
```diff
@@ -0,0 +1,37 @@
...
+if [[ $# -gt 0 ]]; then
+	echo "No arguments are supported."
+	usage
+fi
+
+output=$(zfs get -Hpo value "com.delphix:hotfix-version" "$(dirname "$(zfs list -Hpo name /)")")
+if [[ "$output" != "-" ]]; then
+	echo -n "$output"
+fi
+
```
prakashsurya: trailing line

prakashsurya (approved):
Looks good! Thank you.

files/common/usr/bin/get-hotfix-version
matt-skinner-delphix: We probably need these script checkers are part of this repo so the developer can run them rather than relying on reviewers and these checkers entirely. It dirty ups the review to have these revisions rather than doing it on the dev's machine.

prakashsurya: I understand the concern, and for these simple checks that may be a reasonable possibility, but for more complicated checks that may not always be achievable (e.g. in other repositories, we do more complicated checks, and duplicating that test environment outside of github actions would not be possible); that is, unless the actions can be run locally.

grwilson (approved):

PR #314, opened 2021-07-29
DLPX-76802 Starting Verification container enables IP forwarding on host

sanbhara:
WIP. It has been observed that simply starting the verification container enables ip forwarding on the host. This seems to persist post a DEFERRED upgrade, leading to potential performance regression in networking. While the change in this diff prevents the ip forwarding from being set, I am not yet certain verification will complete successfully.

The relevant property can be viewed on the engine using below command :
```
sudo sysctl net.ipv4.ip_forward
net.ipv4.ip_forward = 0
```
git ab-pre-pusht: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5798/

# Problem:

# Solution:

# Testing

### DIFF
--- a/files/common/lib/systemd/network/80-container-vz.network.d/override.conf
+++ b/files/common/lib/systemd/network/80-container-vz.network.d/override.conf
@@ -1,3 +1,2 @@
 [Network]
-LinkLocalAddressing=no
 IPMasquerade=no


prakashsurya (commented):
Seems reasonable to me.

Out of curiosity, could we also achieve this by a setting at the individual container level? E.g. we set some network settings for the container, when the container configuration is created (i.e. via the `upgrade-container` script), so I'm just curious if we could fix this issue by changing the configuration of the container.. 

I ask since I think this change will likely require a "hotfix" to apply.. and/or have the customer upgrade to a fixed version.. where that upgrade would have "broken" the system.. so if we could modify the upgrade container configuration, we may be able to fix this via upgrade, and without needing a hotfix.

sanbhara:
> Seems reasonable to me.
> 
> Out of curiosity, could we also achieve this by a setting at the individual container level? E.g. we set some network settings for the container, when the container configuration is created (i.e. via the `upgrade-container` script), so I'm just curious if we could fix this issue by changing the configuration of the container..
> 
> I ask since I think this change will likely require a "hotfix" to apply.. and/or have the customer upgrade to a fixed version.. where that upgrade would have "broken" the system.. so if we could modify the upgrade container configuration, we may be able to fix this via upgrade, and without needing a hotfix.

Good points. We did have a similar concerns during the fixes we made for getting verification to work on GCP instances, but in that case since no customers were recorded as having engines yet on GCP, we got away with just making the change that is expected to be in the lower version. Here we would need to try and alleviate the issue even for engines that do not have these override properties.

If we need to create the container with the property setting that's different from the default, we would need to use systemd-nspawn directly to create the container, rather than machinectl since machinectl only takes very limited set of properties. Another method could be generating the necessary override file directly in the container's root fs in the /var/lib/machines/... location after creating the container and before starting it.

In either case, I have since found that turning off port forwarding on the host breaks upgrade verification so there is more investigation that is needed for this.

prakashsurya:
>  we would need to use systemd-nspawn directly to create the container

> Another method could be generating the necessary override file directly in the container's root fs in the /var/lib/machines/... location after creating the container and before starting it.

IIRC, this is exactly what we do, already.

> In either case, I have since found that turning off port forwarding on the host breaks upgrade verification so there is more investigation that is needed for this.

What fails with it turned off?

prakashsurya:
> IIRC, this is exactly what we do, already.

Yea, starting about [here](https://github.com/delphix/appliance-build/blob/master/upgrade/upgrade-scripts/upgrade-container#L286), you can see we do various configuration of the container.. I think we pretty much have complete control over what settings we use.

prakashsurya:
And you can see [here](https://github.com/delphix/appliance-build/blob/master/upgrade/upgrade-scripts/upgrade-container#L385), we even override the `systemd-nspawn` command used to start the container.

prakashsurya:
Closing in favor of https://github.com/delphix/appliance-build/pull/611 and https://github.com/delphix/appliance-build/pull/612.

PR #321, opened 2021-08-04
DLPX-76874 delphix-platform: add dependency on ntp

pzakha:
See JIRA for context

## Testing
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5860/
- Manual upgrade testing to focal

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -54,6 +54,7 @@ DEPENDS += ansible, \
 	   debsums, \
 	   dmidecode, \
 	   net-tools, \
+	   ntp, \
 	   open-iscsi, \
 	   openssh-server, \
 	   openssl, \


prakashsurya (commented):
I'm not necessarily opposed to this change, but now that we've landed https://github.com/delphix/appliance-build/pull/585 .. I think a better fix (assuming it'd work), would be to move the installation of the "delphix-virtualization" package to a variant specific package list in appliance-build, rather than using Ansible like we do now.

Would you be interested in trying out that solution? I'm willing to open a PR to do this, if you'd be willing to test it out for me (or let me know how to test it myself).

pzakha:
You can test changes to the focal branch by running `ab-pre-push`.

As a tip if you are planning to move the installation of `delphix-virtualization` into the package lists, it _might_ be necessary to add a `Pre-depends` to `delphix-virtualization` on `delphix-platform` but I'm not sure if even that would be sufficient. For instance, we have the following code in delphix-platform's `postinst`:
```
	if ! id -u postgres >/dev/null; then
		# When installing postgres, a postgres user is created unless it
		# already exists. To have a consistent UID accross installations
		# and with systems that were migrated from Illumos, we create a
		# user with the same UID as in Illumos. The Ubuntu postgres
		# package requires that the postgres user is also part of the
		# postgres group. Again, for consistency, we create the postgres
		# group with a pre-determined GID, which the same as the
		# postgres UID. Note that we put this code here instead of the
		# delphix-platform service as we need it to be executed before
		# the postgres package gets installed.
		addgroup postgres --gid 65437
		adduser --home /var/lib/postgresql --no-create-home \
			--shell /bin/bash --ingroup postgres \
			--gecos "PostgreSQL administrator" --uid 65437 \
			--disabled-password postgres
	elif [[ $(id -u postgres) -ne 65437 ]]; then
		echo "ERROR: postgres id $(id -u postgres) != 65437" >&2
		exit 1
	fi
```
Basically we need to ensure that the postgres user is created before `postgresql-common.postinst` is ran. This could be done fairly simply by adding a `Depends: delphix-platform` on postgres but we are not currently building postgres ourselves.

prakashsurya (approved):
I'll go ahead and approve this as-is.. I think the shift to package-lists vs. installing packages via Ansible in appliance-build might be a bit larger endeavor than I was expecting.

pcd1193182 (approved):

PR #323, opened 2021-08-10
DLPX-76874 delphix-platform: add dependency on ntp (re-integrate)

pzakha:
This is a follow-up to #321 which was reverted in #322 because it caused DLPX-76954.

The problem with #321 was that previously ntp was installed by delphix-virtualization which took care of masking and disabling the service before it could run, but now that the installation of ntp was done before delphix-virtualiation's preinst/postinst scripts could run, the ntp service was never disabled.

This fix copies the logic from delphix-virtualization to deal with ntp.

One caveat with this new logic is that now on the `internal-dcenter` variant we would no longer run neither systemd-timesyncd nor ntp. Not sure how much of an issue that is though.

An alternative would be to not integrate this and to instead handle DLPX-76872 differently, however I think that having appliance-build first install systemd-timesyncd and then have it later replaced by ntp is not great.

A better approach in my opinion if we care about keeping the time on `internal-dcenter` syncrhonized would be to continue with this change, and then add additional code in the `internal-dcenter` ansible scripts to  re-enable ntp and set it up with some servers to sync-up with.

## Testing
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5919/ (pass)

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -53,6 +53,33 @@ configure)
 	systemctl enable systemd-networkd.service
 	systemctl enable iscsi-name-init.service
 
+	#
+	# There's 3 types of services that we account for:
+	#
+	#  1. Services that are always enabled.
+	#  2. Services that are always disabled.
+	#  3. Services that are disabled by default, but may be
+	#     dynamically enabled.
+	#
+	# For more info, see comment in the preinst script.
+	#
+
+	#
+	# Here we account for services of type (3)
+	#
+	# The systemctl is-enabled will return false if the service is masked,
+	# disabled, or if the service doesn't exist yet. By checking this we will
+	# only unmask and disable the service if it wasn't already enabled.
+	#
+	while read -r svc; do
+		if ! systemctl is-enabled "$svc"; then
+			rm -f "/etc/systemd/system/$svc"
+			systemctl disable "$svc"
+		fi
+	done <<-EOF
+		ntp.service
+	EOF
+
 	if ! id -u postgres >/dev/null; then
 		# When installing postgres, a postgres user is created unless it
 		# already exists. To have a consistent UID accross installations
--- a/debian/preinst
+++ b/debian/preinst
@@ -1,0 +1,54 @@
+#!/bin/bash -eux
+#
+# Copyright (c) 2021 by Delphix. All rights reserved.
+#
+
+#
+# Note that the logic below was copied from delphix-virtualization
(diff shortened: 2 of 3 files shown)


basil (commented):
I think disabling NTP on DCenter hosts would likely lead to clock skew and operational problems. For example, we run `awscli(1)` on DCenter hosts, and some preliminary searching shows that `aws s3` commands can return `RequestTimeTooSkewed` errors if there is clock skew (see [this blog post](https://aws.amazon.com/blogs/developer/clock-skew-correction/)). We also have cleanup jobs that run on DCenter hosts that rely on filesystem timestamps being accurate.

pzakha:
Okay, so it seems like this would have to be addressed before pushing this change then.

pzakha:
@basil We should be safe integrating this PR as a time daemon will keep running even once this is integrated. This won't be the case after we move to Ubuntu 20.04, so I've also posted https://github.com/delphix/appliance-build/pull/594 (more details are in there).

prakashsurya:
> I think disabling NTP on DCenter hosts would likely lead to clock skew and operational problems.

Where does NTP get installed an enabled on DCOL systems right now?

pzakha:
> Where does NTP get installed an enabled on DCOL systems right now?
ntp doesn't currently get installed on DCOL systems. They use systemd-timesyncd right now, which comes enabled out of the box with systemd.

prakashsurya:
I don't even see the package installed on dcol2:
```
$ dpkg -l | grep ntp | wc -l
0
```
are we sure we're using NTP on dcenter?

prakashsurya:
is the conern that, this change will move dcenter from systemd-timesyncd, to ntp.. and this will also leave ntp disabled.. so that this change will leave dcenter with no time-syncing mechanism enabled out-of-the-box?

basil (approved):

prakashsurya (approved):
LGTM.. are you planning to also remove NTP from the list of virtualization dependencies, and remove this logic from that package's postinst/preinst hooks?

pzakha:
> LGTM.. are you planning to also remove NTP from the list of virtualization dependencies, and remove this logic from that package's postinst/preinst hooks?

Yeah, I'll open another bug for it after this lands

pcd1193182 (approved):

PR #324, opened 2021-09-21
DLPX-77515 rpool should use lz4 compression

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6149/)

I specifically chose not to put the `zpool upgrade` call in the existing `delphix-platform` service, to keep it from running inside of the upgrade verification container (and modifying the pool's format from within the container). As a result, the new pool format won't take affect until the first boot, since this means it also will not run at build-time, when building the initial VM disk images. I think that should be fine, and perhaps preferable, since the initial VM disk images are generated on an Ubuntu system, rather than a Delphix system, so the pool features available might be different.

One thing to note when upgrading the pool's on-disk format, is that we need to ensure that GRUB is compatible with all the features that we enable. I've verified that running `zpool upgrade rpool` works, in that I can boot from `rpool` after I do that, on a trunk-based system. It's unclear if this will continue to be the case, as new pool features are added. Still, I chose to use `zpool upgrade` here with the assumption that this won't be a practical problem for us, and the notion that it's likely better to use new features as soon as they become available; if this does become a problem, we should detect the issue during our testing cycle, and we could address it at that time.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -50,6 +50,7 @@ configure)
 	systemctl enable auditd.service
 	systemctl enable delphix.target
 	systemctl enable delphix-platform.service
+	systemctl enable delphix-rpool-upgrade.service
 	systemctl enable systemd-networkd.service
 	systemctl enable iscsi-name-init.service
 
--- a/debian/prerm
+++ b/debian/prerm
@@ -27,6 +27,7 @@ upgrade | remove)
 
 	systemctl disable delphix.target
 	systemctl disable delphix-platform.service
+	systemctl disable delphix-rpool-upgrade.service
 
 	#
 	# We also need to remove the "ansible-done" file, to ensure that
--- a/files/common/lib/systemd/system/delphix-rpool-upgrade.service
+++ b/files/common/lib/systemd/system/delphix-rpool-upgrade.service
@@ -1,0 +1,29 @@
+#
+# Copyright 2021 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+[Unit]
+Description=Delphix "rpool" Upgrade Service
+ConditionVirtualization=!container
+PartOf=delphix.target
+After=local-fs.target
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/zpool upgrade rpool
+RemainAfterExit=yes
+
+[Install]
+WantedBy=delphix.target


ahrens (approved):

prakashsurya:
I verified this works on a trunk based VM on AWS.. the service log shows this:
```
-- Logs begin at Tue 2021-09-28 16:53:15 UTC, end at Tue 2021-09-28 17:48:38 UTC. --
Sep 28 17:45:43 ip-10-110-244-244 systemd[1]: Starting Delphix "rpool" Upgrade Service...
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]: This system supports ZFS pool feature flags.
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]: Successfully upgraded 'rpool' from version 28 to feature flags.
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]: Enabled the following features on 'rpool':
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   async_destroy
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   empty_bpobj
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   lz4_compress
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   multi_vdev_crash_dump
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   spacemap_histogram
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   enabled_txg
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   hole_birth
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   extensible_dataset
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   embedded_data
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   bookmarks
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   filesystem_limits
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   large_blocks
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   large_dnode
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   sha512
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   skein
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   edonr
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   userobj_accounting
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   encryption
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   project_quota
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   device_removal
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   obsolete_counts
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   zpool_checkpoint
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   spacemap_v2
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   allocation_classes
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   resilver_defer
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   bookmark_v2
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   redaction_bookmarks
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   redacted_datasets
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   bookmark_written
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   log_spacemap
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   livelist
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   device_rebuild
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   zstd_compress
Sep 28 17:45:49 ip-10-110-244-244 zpool[977]:   draid
Sep 28 17:45:49 ip-10-110-244-244 systemd[1]: Started Delphix "rpool" Upgrade Service.
Sep 28 17:47:08 ip-10-110-244-244 systemd[1]: Stopped Delphix "rpool" Upgrade Service.
-- Reboot --
Sep 28 17:48:01 ip-10-110-244-244 systemd[1]: Starting Delphix "rpool" Upgrade Service...
Sep 28 17:48:01 ip-10-110-244-244 zpool[1041]: This system supports ZFS pool feature flags.
Sep 28 17:48:01 ip-10-110-244-244 zpool[1041]: Pool 'rpool' already has all supported and requested features enabled.
Sep 28 17:48:02 ip-10-110-244-244 systemd[1]: Started Delphix "rpool" Upgrade Service.
```
i.e. I installed the new package, rebooted.. verified rpool was upgraded on the next boot.. rebooted again.. verified system booted, and no new features got enabled on the next boot.

prakashsurya:
Similarly, I verified this on a 6.0/stage based VM on AWS..
```
 sudo journalctl -u delphix-rpool-upgrade
-- Logs begin at Tue 2021-09-28 17:51:48 UTC, end at Tue 2021-09-28 17:59:32 UTC. --
Sep 28 17:53:04 ip-10-110-227-126 systemd[1]: Starting Delphix "rpool" Upgrade Service...
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]: This system supports ZFS pool feature flags.
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]: Successfully upgraded 'rpool' from version 28 to feature flags.
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]: Enabled the following features on 'rpool':
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   async_destroy
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   empty_bpobj
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   lz4_compress
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   multi_vdev_crash_dump
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   spacemap_histogram
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   enabled_txg
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   hole_birth
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   extensible_dataset
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   embedded_data
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   bookmarks
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   filesystem_limits
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   large_blocks
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   large_dnode
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   sha512
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   skein
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   edonr
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   userobj_accounting
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   encryption
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   project_quota
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   device_removal
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   obsolete_counts
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   zpool_checkpoint
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   spacemap_v2
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   allocation_classes
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   resilver_defer
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   bookmark_v2
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   redaction_bookmarks
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   redacted_datasets
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   bookmark_written
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   log_spacemap
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   livelist
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   device_rebuild
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   zstd_compress
Sep 28 17:53:12 ip-10-110-227-126 zpool[7232]:   draid
Sep 28 17:53:12 ip-10-110-227-126 systemd[1]: Started Delphix "rpool" Upgrade Service.
Sep 28 17:57:55 ip-10-110-227-126 systemd[1]: Stopped Delphix "rpool" Upgrade Service.
-- Reboot --
Sep 28 17:58:38 ip-10-110-227-126 systemd[1]: Starting Delphix "rpool" Upgrade Service...
Sep 28 17:58:38 ip-10-110-227-126 zpool[3810]: This system supports ZFS pool feature flags.
Sep 28 17:58:38 ip-10-110-227-126 zpool[3810]: Pool 'rpool' already has all supported and requested features enabled.
Sep 28 17:58:38 ip-10-110-227-126 systemd[1]: Started Delphix "rpool" Upgrade Service.
```
This time, I restarted `delphix.target` instead of rebooting to run the service the first time, and get the pool upgraded.. then I rebooted, and verified the system booted up properly (i.e. with the upgraded pool feature set).

grwilson (approved):

pzakha:
Once we upgrade rpool, would that in some cases prevent us from rolling back to the previous version? If that's the case, shouldn't we ensure we do that once we are past the no-rollback point?

prakashsurya:
Maybe? @ahrens can you comment on that?

prakashsurya:
FWIW, if rollback is an issue here, I'm not sure I'm convinced pushing this until after the point-of-no-return is how we should address it. IMO, we should be trying to move in a direction where the point-of-no-return doesn't exist, rather than adding more reasons to keep it.

So, rather, id suggest we modify the script to only enable "rollback safe" features. Which, I think means, only enabling features understood by the initial 6.0.0.0 release. The policy to only support upgrades from 12-month old releases would help us here, moving forward.

pzakha:
Yeah, that's a good idea. I don't remember if Linux allows to upgrade to the "feature flags" version and selectively enable feature flags though.

ahrens:
At one point, I think that when we upgraded, we would run `zpool upgrade domain0` on the old bits, so that we are sure to be able to fall back to the old bits.  I'm not sure if we still do that, but we could take that approach for rpool.

That said, yes you should be able able to name exactly which features to enable (`zpool set feature@...=enabled`).  But it's unlikely that we will regularly update that list.

prakashsurya:
> we would run zpool upgrade domain0 on the old bits

I think this is roughly equivalent to waiting until after the point-of-no-return to upgrade the pool.. i.e. after the point-of-no-return, we could immediately upgrade the pool, or wait until the next upgrade is initiated.. either way, after that pool upgrade, rollback is no longer possible.

> I'm not sure if we still do that

IIRC, today we upgrade the domain0 pool on the new bits, but within the checkpoint, so that we can rewind that operation on rollback.

We don't take a checkpoint of rpool, so I opted not to follow this approach for rpool; further, I'm not sure we could rewind to the checkpoint even if we did create one; e.g. for domain0, we export, then import with rewind, which I don't think would work the same for rpool as it's always mounted.

> But it's unlikely that we will regularly update that list.

Yea, that's the drawback of setting each feature explicitly; although, I could see that as a benefit as well, as I'm not sure we really want to be manipulating rpool's feature set all that often (e.g. due to compatibility with the bootloader).

It's not clear to me which approach is best; @ahrens @pzakha do either of you have a preference?

If we really want to try to enable all available features, then I'm slightly leaning towards the "upgrade after point-of-no-return" approach.

If we're OK having a more static list of features that we explicitly enable, that are known to be compatible with both rollback and the bootloader, then I like the "list all feature explicitly in the script" approach.

pzakha:
I'd probably be fine with having a list of features we enable, even if we (almost) never update it. We can update that list whenever there's a new feature we want to have. Given that we've been running with rpool at version 28 for all those years without issues, not running the latest and greatest features would not be an issue IMO.

The advantage of controlling the feature list is that it will also allow us to keep the pool importable by a stock Ubuntu VM (even though I'm not sure we really have a use case for that).

ahrens:
I think either approach is fine.  I agree with Pavel that having the list of features will probably work well enough.

prakashsurya:
OK, I've opened: https://github.com/delphix/delphix-platform/pull/328

PR #327, opened 2021-09-29
Upgrade of rpool needs to depend on rollback service

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6200/)

prakashsurya:
@pzakha I'm not yet convinced this is the right solution, as I said in #324, but I figured I'd open the PR in case we decide to take this approach.

pzakha:
Looking at what the rollback service does, it seems to launch the monitor script asynchronously, meaning that it will be "started" before we reach the point of no return. Am I reading this correctly?

But in any case, I think our discussion in #324 will lead to a better approach.

PR #333, opened 2021-10-14
DLPX-77901 Prevent services from being re-enabled on upgrade

pzakha:
See https://github.com/delphix/appliance-build/pull/615 for more context.


prakashsurya (approved):
I approve, but can you clarify (or perhaps add documentation where appropriate) where and how we should be doing service manipulation for both, initial VMs and for upgrade? E.g. why do we handle "ntp" via appliance-build and upgrade-scripts, but we handle other services in this post-install script? I want to make sure it's clear when to use the packaging hooks, and when to use appliance-build + upgrade-scripts, and I'm not sure that delineation is clear to me.

PR #335, opened 2021-10-20
DLPX-78004 Add script to retrieve minimum version property from root …

vimalraj-anbarasu:
Manually tested the script.

### DIFF
--- a/files/common/usr/bin/get-minimum-version.sh
+++ b/files/common/usr/bin/get-minimum-version.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+#
+# Copyright 2021 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


files/common/usr/bin/get-minimum-version.sh
```diff
@@ -0,0 +1,36 @@
...
+	exit 2
+}
+
+if [[ $# -gt 0 ]]; then
+	echo "No arguments are supported."
+	usage
+fi
+
+output=$(zfs get -Hpo value "com.delphix:minimum-version" "$(dirname "$(zfs list -Hpo name /)")")
+if [[ "$output" != "-" ]]; then
```
prakashsurya, line 34: I think we should return an error, if the property is not found (e.g. `exit 1` or equivalent).. since unlike the hotfix property, this one should always exist.. (since we're adding this script to 6.0.11, and the property should always be there on versions 6.0.11 or greater)

PR #337, opened 2021-11-08
TOOL-12429 delphix-platform fails to build using latest ansible version

pzakha:
This will allow us to keep building delphix-platform with the latest version of ansible provided by Ubuntu. See JIRA for more details.

## Testing
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6516/
- manually tested that this works by upgrading ansible on bootstrap image and then building delphix-platform on it.

### DIFF
--- a/bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
@@ -26,6 +26,8 @@
     - python3-docker
 
 - docker_image:
-    path: "{{ toplevel.stdout }}/docker"
+    build:
+      path: "{{ toplevel.stdout }}/docker"
     name: delphix-platform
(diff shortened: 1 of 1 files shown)


sebroy (approved):

prakashsurya (approved):
LGTM.. maybe we should open a bug to remove the use of docker, now that linux-pkg and other stuff has been solidified...

pzakha:
> LGTM.. maybe we should open a bug to remove the use of docker, now that linux-pkg and other stuff has been solidified...

Yeah, we could do that. I wanted to do it at first, but it requires a bit more work since we build the package for several different platforms and so it would also require some changes in linux-pkg & potentially a flag day. I've opened TOOL-12443.

PR #338, opened 2021-11-09
DLPX-78306 [Backport of DLPX-76874 to 6.0.12.0] delphix-platform: add dependency on ntp

pzakha:
This essentially backports #323 and then also #333 which undoes a bunch of the complexity introduced in #323.

This will be pushed at the same time as the remaining work for DLPX-78304

## Testing
See https://github.com/delphix/appliance-build/pull/627

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -54,6 +54,7 @@ DEPENDS += ansible, \
 	   debsums, \
 	   dmidecode, \
 	   net-tools, \
+	   ntp, \
 	   open-iscsi, \
 	   openssh-server, \
 	   openssl, \


debian/rules
```diff
@@ -54,6 +54,7 @@ DEPENDS += ansible, \
 	   debsums, \
 	   dmidecode, \
 	   net-tools, \
+	   ntp, \
```
dbj-delphix, line 57: Does this address [TOOL-12422](https://jira.delphix.com/browse/TOOL-12422)?

pzakha: No, it doesn't, it's just a backport from master. It does address the current failure to build the internal-dcenter variant though.

prakashsurya (approved):

sdimitro (approved):

PR #347, opened 2021-12-07
DLPX-77871 Need to install the usrmerge package on Ubuntu 20.04

sdimitro:
## Motivation:

This is needed so engines with Ubuntu 18.04 that are migrated to 20.04 have the exact same root filesystem structure/hierarchy as fresh install of 20.04.

## Testing:

ab-pre-push: TBA

Manual Testing TBA - upgrade an 18.04 VM with the upgrade image of the above link and ensure that the structure has indeed changed.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -110,6 +110,14 @@ DEPENDS += $(DEPENDS.$(TARGET_PLATFORM))
 DEPENDS += delphix-build-info,
 
 #
+# The usrmerge package modifies the layout of directories under root (/) upon
+# installation, to ensure that a Delphix Engine upgraded to Ubuntu 20.04 has
+# the same directory layout as a Delphix Engine that initially came on
+# Ubuntu 20.04 (or later).
+#
+DEPENDS += usrmerge,
+
+#
 # These packages are tools that are intended for human convenience. The
 # product should not rely on them programmatically. They may be updated
 # or replaced without regard for backward compatibility.


prakashsurya (approved):
I'm not familiar with this package, but this code change is simple enough.

sdimitro:
> I'm not familiar with this package, but this code change is simple enough.

In 20.04 the `/{bin,lib,sbin}` and `/usr/{bin,lib,sbin}` directories where merged by having `/{bin,lib,sbin}` being a symlink to `/usr/{bin,lib,sbin}`:
```
:~$ ls -lh /
total 49K
lrwxrwxrwx   1 root root    7 Nov 30 11:05 bin -> usr/bin
drwxr-xr-x   3 root root   14 Dec  7 21:10 boot
drwxr-xr-x  16 root root 3.5K Dec  7 21:07 dev
drwxr-xr-x 122 root root  219 Dec  7 21:09 etc
...cropped...
lrwxrwxrwx   1 root root    7 Nov 30 11:05 lib -> usr/lib
lrwxrwxrwx   1 root root    9 Nov 30 11:05 lib32 -> usr/lib32
lrwxrwxrwx   1 root root    9 Nov 30 11:05 lib64 -> usr/lib64
lrwxrwxrwx   1 root root   10 Nov 30 11:05 libx32 -> usr/libx32
...cropped...
drwxr-xr-x  28 root root 1.1K Dec  7 23:59 run
lrwxrwxrwx   1 root root    8 Nov 30 11:05 sbin -> usr/sbin
...cropped...
drwxrwxrwt  11 root root   13 Dec  8 00:00 tmp
drwxr-xr-x  15 root root   15 Nov 30 11:09 usr
drwxr-xr-x  17 root root   19 Nov 30 11:34 var
```

Ubuntu is not the only distro that has done this, quite the opposite apparently a lot of distros have already moved to this hierarchy already and we are catching up. The motivation and more details on why this merging happens can be found here: https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/ - TL;DR; the main reason (in my understanding) is compatibility between different Linux distros and Unixes (apparently Solaris was the first "Unix" to do this) and having a simpler hierarchy so build tools can have an easier time finding dependencies.

Ubuntu 20.04 images themselves don't need the `usrmerge` package because their hierarchy is created like this. The package was specifically created for folks upgrading from older Ubuntu versions to 20.04 such that they can make their hierarchy consistent with this new standard. My understanding is that the `usrmerge` debian package is simply a post-hook installation shell script that does the hierarchy migration for you.

Do the above help? @prakashsurya

sebroy (approved):

PR #350, opened 2021-12-10
DLPX-78424 Increase ulimit for open files

naveen-singh-delphix:
### Context:
This pull request attempts to increase the soft limit for "number of open file descriptors" for the users "delphix" and "root" on our engine.

### Problem:
The performance team encountered "out of file descriptors" errors on some of their test cases in the fio test suite. The fio suite itself has a bug which leads to leaked file descriptors during the test runs.

### Solution:
We are increasing the soft limit for "open file descriptors" from the current value of 1024 to 32768. This value was decided during consultations with the performance team. Since we do not see any reason for users other than "delphix" and "root" ever running into this problem, we want to keep the change limited to the affected users only. We are thus going for the minimum set of changes that are needed to fix this problem.
The hard limit is kept unchanged from its current value of 1048576 because we do not expect to hit this limit in any scenario in the near future.

### Testing:
1. The fix was tested using a simple C program that keeps creating new file descriptors until it hits the current soft limit which can be checked using "ulimit -Sn".
2. ab-pre-push results -- http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/89

### Implementation:
The fix is implemented by adding the relevant lines via ansible playbook in /etc/security/limits.conf.

### Notes To Reviewers:
Any extra information a reviewer may need to know before reviewing
your change. For example here you might want to describe which files
should be looked at first or which files are auto-generated.

### Deployment Plan:
Some changes get more complicated and may need changes in multiple
repositories or may require infrastructure changes. Describe how these
changes will be smoothly deployed.

### Future work:
NA

### Bonus:
NA


files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -173,6 +173,14 @@
     - 'root'
     - '*'
 
+- lineinfile:
+    create: yes
+    path: /etc/security/limits.conf
```
prakashsurya, line 178: Rather than modifying the existing file, can we instead add a new file in `/etc/security/limits.d` to do this? If so, then we can add that file to the repository like this:
```
± cat files/common/etc/security/limits.d/delphix.conf
root soft nofile 32768
delphix soft nofile 32768
```

I'm not sure if files under `limits.d` will take effect, but if they do, then I think this would be a better approach; e.g. less chance to conflict with upstream changes to the default `limits.conf` file, and easier to change or remove these settings on upgrade.

prakashsurya: Although.. now I see we already modify `limits.conf` just a few lines above.. so I'll go ahead and approve this, since it's consistent with our existing logic..

I still think both should probably be moved to a new file under `limits.d`, but we can take that on in a follow up change.

prakashsurya (approved):

sdimitro (approved):
LGTM!

As a final test, could you maybe get @jwk404 or Kevin Green to run their fio workload and make sure it doesn't reproduce?

The above ab-pre-push that you ran has generated an image in AWS that you can clone like this: `dc clone-latest dlpx-naveen.singh-master <vm name>`

After creating a VM with the above command, I would log in to that VM make sure that the new lines have been added to the config file and have one of the aforementioned performance folks verify that the we no longer hit the file limit with their fio workload.

prakashsurya:
FWIW, I'd suggest being explicitly about the snapshot when cloning from the group.. otherwise, other test runs could update the group, and one could wind up testing with the wrong image.. i.e.
```
dc clone -s ami-0d7518ea2a02d0eec dlpx-naveen.singh-master <vm name>
```

naveen-singh-delphix:
> FWIW, I'd suggest being explicitly about the snapshot when cloning from the group.. otherwise, other test runs could update the group, and one could wind up testing with the wrong image.. i.e.
> 
> ```
> dc clone -s ami-0d7518ea2a02d0eec dlpx-naveen.singh-master <vm name>
> ```

Great point, thanks for suggesting it.

naveen-singh-delphix:
> LGTM!
> 
> As a final test, could you maybe get @jwk404 or Kevin Green to run their fio workload and make sure it doesn't reproduce?
> 
> The above ab-pre-push that you ran has generated an image in AWS that you can clone like this: `dc clone-latest dlpx-naveen.singh-master <vm name>`
> 
> After creating a VM with the above command, I would log in to that VM make sure that the new lines have been added to the config file and have one of the aforementioned performance folks verify that the we no longer hit the file limit with their fio workload.

Yes, I have confirmed the presence of the new entries in /etc/security/limits.conf and requested Perf team to validate the changes using their performance tests.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
naveen-singh-delphix: https://delphix.atlassian.net/browse/DLPX-78757 has been created to track this improvement.

naveen-singh-delphix:
> 

Got confirmation from Performance team that the fix works fine.

PR #366, opened 2022-04-11
DLPX-80427 zcache crashes after a deferred upgrade

grwilson:
This overrides our `zfs-object-agent` systemd service so that it's included as part of `delphix.target`.


prakashsurya (approved):
I'm OK doing it this way, but can you add some details to commit message (or somewhere) about why we're choosing this over adding it directly to the `zfs-object-agent.service` file itself, in the zfs repository.. so we can capture that information and answer future questions about why we chose to do it this way..

manoj-joseph (approved):

PR #370, opened 2022-04-13
DLPX-80625 delphix-platform build failure

prakashsurya:


prakashsurya:
building the package is still failing, so presumably the issue is not due to #368

prakashsurya:
So far, I think the issue is actually due to the following set of curcumstances:

1. we run the package build inside a Docker container
2. we mount the git repository in that container as a Docker volume
3. we run `git rev-parse`

It looks like `git rev-parse` is failing like so:
```
fatal: unsafe repository ('/opt/delphix-platform' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /opt/delphix-platform
```

I think this workflow used to work, but now might not due to us using a newer version of `git`. This change perhaps driven by this security issue: https://github.blog/2022-04-12-git-security-vulnerability-announced/ (not quite sure, just speculating based on some google searching around). Since we don't pin the version of `git` that we install in the Docker environment, I wouldn't be surprised if this behavior changed, and we're now/automatically using a newer (incompatible) version of `git` than we used previously.

I also verified that all prior of our delphix-platform pakcage builds that were successful, occurred before April 12th, when that vulnerability was posted. Further, our `master` builds appear to be suffering from this as well as the reported `6.0/stage` builds.

The solution I'm taking here, is to simply avoid running `git rev-parse` from within the Docker container, as I don't think it's necessary to do so. As a follow up, we might want to remove the use of the Docker container entirely, or somehow configure the permissions such that `git rev-parse` works in that environment.

grwilson (approved):

sebroy (approved):

PR #375, opened 2022-04-21
DLPX-81199 don't run zfs-mount service within upgrade container

prakashsurya:
Note, this depends on http://reviews.delphix.com/r/81395/ landing first.

### DIFF
--- a/files/common/lib/systemd/system/zfs-import-cache.service.d/override.conf
+++ b/files/common/lib/systemd/system/zfs-import-cache.service.d/override.conf
@@ -1,0 +1,11 @@
+#
+# During upgrade verification, the root filesystem will be booted as a
+# container using systemd-nspawn. We don't "sandbox" the container, so
+# when zfs-import-cache.service runs, it could potentially try to import
+# a pool, which is not desired.
+#
+# To prevent this behavior, we explicitly disable this service from
+# running when inside of the container.
+#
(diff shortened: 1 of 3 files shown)


sebroy:
In the product, upgrade verification relies on domain0 filesystems being present in the container, right (specifically `/mds`, which is in the domain0 pool)?

prakashsurya:
> (specifically /mds, which is in the domain0 pool)?

ah, right.. presumably the upgrade testing is going to catch that.. I think it might be better to mount `domain0/mds` specifically, rather than everything from `domain0`.. I think the mounting everything approach, also adds latency upgrade verification when it's run on customer systems with lots of VDBs/etc..

so, yea, I think this change as is might not be enough.. I'll think about where/how to mount `domain0/mds` (assuming that's not handled already in some other script)..

prakashsurya:
@sebroy I've opened http://reviews.delphix.com/r/81395/ to address your concern; it looks like that is all that's needed for the upgrade verification clone of postgres to work with the changes being proposed in this PR.

dbj-delphix (approved):
This LGTM

prakashsurya:
See #375

prakashsurya:
I seem to have dropped the ball on this.. re-opening here: https://github.com/delphix/delphix-platform/pull/465

PR #385, opened 2022-08-05
DLPX-82936 Handle generation of MAC address based interface names on new engine installations.

naveen-singh-delphix:
# Context:
This PR attempts to switch from kernel assigned network interface names to MAC address based interface names in AWS environments only. It specifically handles creation of a script that gets invoked post upgrade to convert the interface names to MAC address based names, and also takes care of creating MAC address based interface names on new engine installations.
This PR is directly related to the following two reviews which handle other parts of the overall fix
a. `https://github.com/delphix/appliance-build/pull/690` -- This part of the fix addresses the conversion of old canonical network interface names in the netplan file to MAC address based names on the upgraded engine.
b. `http://reviews.delphix.com/r/83630` -- Enable the generation of new MAC address based interface names cloned instances in AWS

# Problem:
In AWS environment, changing instance types renders the new instance without any network configuration. This is due to the fact that the network interface names are assigned by the kernel based on the path on which they are enumerated during the boot, and this enumeration order can change while booting into the instance after changing its instance type. Depending on the enumeration path, the interface names can be different to the existing ones in the netplan file. The stale network interface cleaner thread in osadmin service ends up cleaning the existing interfaces because they do not match with the new interface names generated on the instance, thus leaving the instance without any network configuration. 

# Solution:
The fix being proposed is to switch to MAC based interface names in our netplan file, which will remain consistent across instance change cycles.

# Testing
The following scenarios have been manually tested --
1. Fresh Install
2. Upgrade
3. Rollback
4. Cloned Instance

# Implementation:
The part of the fix inside delphix-platform has two sub-parts --
1. To handle the conversion of the interface names from the canonical names assigned by the kernel to MAC based names, we run a conversion script as part of the PreExecStart stage of systemd-networkd daemon.
2. To handle the correct generation of MAC based interface names in fresh installed engines, we leverage the debian/postinst file and push our own custom version of the 80-net-setup-link.rules file under /etc/udev/rules.d directory. This custom version overrides the default version of the 80-net-setup-link.rules file under /usr/lib/udev/rules.d.


debian/postinst
```diff
@@ -72,6 +72,29 @@ configure)
...
+  # We are switching to MAC address based network interface names ini AWS environment
+  # from 6.0.10.0 onwards. To handle this change, we leverage the default
+  # 80-net-setup-link.rules udev rules file. We make a copy of this file under /etc
+  # and then change the naming policy to MAC address based names. Please see DLPX-75209
+  # for more context on this change.
+  platform_type=$(cat /var/lib/delphix-appliance/platform)
+  if [[ "$platform_type" == "aws" ]]; then
+    cp /usr/lib/udev/rules.d/80-net-setup-link.rules /etc/udev/rules.d
+    rc=$?
+    if [[ "$rc" -ne 0 ]]; then
```
sebroy, line 85: No need for an extra variable here, you can test $? directly.

debian/postinst
```diff
@@ -72,6 +72,29 @@ configure)
...
+      echo "ERROR: Could not copy the /usr/lib/udev/rules.d/80-net-setup-link.rules to /etc"
+      exit 1
+    fi
+    sed -i 's/ID_NET_NAME/ID_NET_NAME_MAC/g' /etc/udev/rules.d/80-net-setup-link.rules
+    rc=$?
```
sebroy, line 90: same here

debian/postinst
```diff
@@ -72,6 +72,29 @@ configure)
 		echo "ERROR: postgres id $(id -u postgres) != 65437" >&2
 		exit 1
 	fi
+
+  # We are switching to MAC address based network interface names ini AWS environment
+  # from 6.0.10.0 onwards. To handle this change, we leverage the default
+  # 80-net-setup-link.rules udev rules file. We make a copy of this file under /etc
+  # and then change the naming policy to MAC address based names. Please see DLPX-75209
+  # for more context on this change.
+  platform_type=$(cat /var/lib/delphix-appliance/platform)
```
prakashsurya, line 81: can we use the `get-appliance-platform` script instead?
```
$ get-appliance-platform 
aws
```

e.g.
```
if [[ "$(get-appliance-platform)" == "aws" ]]; then
```

debian/postinst
```diff
@@ -72,6 +72,29 @@ configure)
...
+    if [[ "$rc" -ne 0 ]]; then
+      echo "ERROR: Could not copy the /usr/lib/udev/rules.d/80-net-setup-link.rules to /etc"
+      exit 1
+    fi
+    sed -i 's/ID_NET_NAME/ID_NET_NAME_MAC/g' /etc/udev/rules.d/80-net-setup-link.rules
+    rc=$?
+    if [[ "$rc" -ne 0 ]]; then
+      echo "ERROR: Could not update the /etc/udev/rules.d/80-net-setup-link.rules file"
+      exit 1
+    fi
```
prakashsurya, line 94: I think this is redundant, and won't even get exercised, since we run the script with `-e`/`errexit`.. right? so lets remove this?

same for the `rc` checking above..

files/aws/lib/systemd/system/systemd-networkd.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStartPre=
+ExecStartPre=+/opt/delphix/server/bin/update_netplan_to_mac_address
```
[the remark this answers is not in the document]
prakashsurya: yea, I think it might be better to create this file via the app-gate, due to the dependency issue.. presumably, the `postinst` logic could go in the app-gate's script too..

debian/postinst
```diff
@@ -72,6 +72,29 @@ configure)
...
 	fi
+
+  # We are switching to MAC address based network interface names ini AWS environment
+  # from 6.0.10.0 onwards. To handle this change, we leverage the default
+  # 80-net-setup-link.rules udev rules file. We make a copy of this file under /etc
+  # and then change the naming policy to MAC address based names. Please see DLPX-75209
+  # for more context on this change.
+  platform_type=$(cat /var/lib/delphix-appliance/platform)
+  if [[ "$platform_type" == "aws" ]]; then
+    cp /usr/lib/udev/rules.d/80-net-setup-link.rules /etc/udev/rules.d
```
prakashsurya, line 83: do we want to copy and modify each time we upgrade? presumably we do, in case the script changes in a new version.. we want to use the new version of the script.. right?

debian/postinst
```diff
@@ -72,6 +72,29 @@ configure)
...
+    rc=$?
+    if [[ "$rc" -ne 0 ]]; then
+      echo "ERROR: Could not copy the /usr/lib/udev/rules.d/80-net-setup-link.rules to /etc"
+      exit 1
+    fi
+    sed -i 's/ID_NET_NAME/ID_NET_NAME_MAC/g' /etc/udev/rules.d/80-net-setup-link.rules
```
prakashsurya, line 89: if no replacement is made, will `sed` throw an error? how do we want to handle such an event? e.g. maybe the script is modified, and the replacement doesn't match anymore...

debian/postinst
naveen-singh-delphix: Yes, this has been fixed.

debian/postinst
naveen-singh-delphix: Yes, this has been fixed.

debian/postinst
naveen-singh-delphix: Yes, this looks much better and has been fixed.

debian/postinst
naveen-singh-delphix: Yes, thanks for pointing out. This has been fixed.

debian/postinst
naveen-singh-delphix: Yes, the copy on each build would take care of copying the latest version of the default 80-net-setup-link.rules file.

debian/postinst
naveen-singh-delphix: In case of any failure in execution of 'sed', we now explicitly fail. 
Does this address your concern ?

files/aws/lib/systemd/system/systemd-networkd.service.d/override.conf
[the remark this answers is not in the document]
naveen-singh-delphix: The dependency has been broken by moving the script update_netplan_to_mac_address to delphix-platform repository (it was earlier insider dlpx-app-gate repo). This should allow the override.conf file and the update_netplan_to_mac_address script to co-exist without any dependency on delphix-virtualization.

The default override.conf file (common across all the platforms) has been renamed to 10-override.conf. The new override.conf file with the PreExecStart directive has been renamed to 20-override.conf file. With these two changes, we ensure that the default and common 10-override.conf file always executes first, and the 20-override.conf file will only execute only in AWS environment.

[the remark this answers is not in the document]
naveen-singh-delphix: Do these changes address all the concerns ?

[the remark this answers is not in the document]
sebroy: Yes, but if we're renaming these, I'd prefer to give them meaningful names (e.g. you could keep the default override file name the same, and name the new one aws-override.conf or update-interface-names.conf, etc.). I don't think that the order matters in this case (i.e. I'm not sure why we'd use a "10" or "20" prefix for each).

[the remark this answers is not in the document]
naveen-singh-delphix: To test the current changes, ,I added a dummy override.conf file to the systemd-networkd service and the systemctl cat command lists it at the end --

delphix@ip-10-110-237-180:~$ sudo systemctl cat systemd-networkd | grep "^# /"
# /lib/systemd/system/systemd-networkd.service
# /usr/lib/systemd/system/systemd-networkd.service.d/10-override.conf
# /usr/lib/systemd/system/systemd-networkd.service.d/20-override.conf
# /usr/lib/systemd/system/systemd-networkd.service.d/override.conf

I believe the override files will get processed in the order above. We would ideally want the default override.conf file to be processed before any platform specific override files. Prefixing the filenames with digits makes it easy to achieve so.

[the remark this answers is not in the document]
sebroy: I thought that we had done an experiment a few weeks ago to determine that the order of the sections included in these files doesn't matter...(?)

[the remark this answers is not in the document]
sebroy: FWIW, I have no problem with giving them an explicit ordering, I'm just making sure I understand the reasoning. My point was also about the naming; I think you can name them anything as long as they have a `.conf` suffix.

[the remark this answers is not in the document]
naveen-singh-delphix: The aws specific override configuration file has been renamed to aws-override.conf file.
The reasoning behind prefixing override configuration files was discussed with Seb and we agreed to continue the names as suggested in the PR. The naming scheme will ensure that the default override.conf file will be processed before any platform specific override files.

sebroy (approved):

debian/postinst
```diff
@@ -72,6 +75,34 @@ configure)
...
+	# We are switching to MAC address based network interface names only in AWS environment
+	# from 6.0.10.0 onwards. To handle this change, we leverage the default 80-net-setup-link.rules udev
+	# rules file. We make a copy of this file under /etc and then change the naming policy to MAC address
+	# based names. Please see DLPX-75209 for more context on this change.
+	# Though unlikely, it is still remotely possible that the /usr/lib/udev/rules.d/80-net-setup-link.rules
+	# file either gets removed from the base OS distribution or gets replaced with another file to serve
+	# the exact/similar purpose. We must detect and inspect such a change and update the renaming logic if
+	# needed. Ideally, we should also detect changes in the contents of this file but we doubt there is an
+	# obvious way of accomplishing this today.
+	if [[ ! -e "$DEFAULT_UDEV_NET_SETUP_LINK_DIR"/80-net-setup-link.rules ]]; then
```
prakashsurya, line 88: are we sure this file will exist on non-aws platforms? what package provides it?

debian/postinst
```diff
@@ -72,6 +75,34 @@ configure)
...
+		echo "$DEFAULT_UDEV_NET_SETUP_LINK_DIR/80-net-setup-link.rules found missing, investigation needed to understand " \
+			"if its location has changed or if the file itself has been replaced with another udev rules file" >&2
+		exit 1
+	fi
+	if [[ "$(get-appliance-platform)" == "aws" ]]; then
+		# If there are any changes in the specific line of the rules file that we are modifying, we fail early
+		# to allow us the opportunity to investigate the changes and update our own changes accordingly.
+		grep "^NAME==\"\", ENV{ID_NET_NAME}\!=\"\", NAME=\"\$env{ID_NET_NAME}\"" \
+			"$DEFAULT_UDEV_NET_SETUP_LINK_DIR"/80-net-setup-link.rules
+		if [[ ! -d "$MODIFIED_UDEV_NET_SETUP_LINK_DIR" ]]; then
```
prakashsurya, line 98: can't we `mkdir` even if the directory already exists? doing so might make the code more simple.. i.e. no need for a conditional here.

debian/postinst
```diff
@@ -15,6 +15,9 @@
 # limitations under the License.
 #
 
+DEFAULT_UDEV_NET_SETUP_LINK_DIR="/usr/lib/udev/rules.d"
+MODIFIED_UDEV_NET_SETUP_LINK_DIR="/etc/udev/rules.d"
```
prakashsurya, line 19: I don't feel strongly, but IMO, I'd define these closer to the logic that'll actually use them..

debian/postinst
```diff
@@ -72,6 +75,34 @@ configure)
...
+		# If there are any changes in the specific line of the rules file that we are modifying, we fail early
+		# to allow us the opportunity to investigate the changes and update our own changes accordingly.
+		grep "^NAME==\"\", ENV{ID_NET_NAME}\!=\"\", NAME=\"\$env{ID_NET_NAME}\"" \
+			"$DEFAULT_UDEV_NET_SETUP_LINK_DIR"/80-net-setup-link.rules
+		if [[ ! -d "$MODIFIED_UDEV_NET_SETUP_LINK_DIR" ]]; then
+			mkdir -p "$MODIFIED_UDEV_NET_SETUP_LINK_DIR"
+		fi
+		cp "$DEFAULT_UDEV_NET_SETUP_LINK_DIR"/80-net-setup-link.rules "$MODIFIED_UDEV_NET_SETUP_LINK_DIR"
+		sed -i 's/ID_NET_NAME/ID_NET_NAME_MAC/g' "$MODIFIED_UDEV_NET_SETUP_LINK_DIR"/80-net-setup-link.rules
+		update-initramfs -u -t -k "$(uname -r)"
```
prakashsurya, line 103: can you add a comment about why we need to call this? it's relatively unusual to need to update the initramfs, so a quick comment might help the reader..

debian/postinst
```diff
@@ -72,6 +75,34 @@ configure)
...
+	# obvious way of accomplishing this today.
+	if [[ ! -e "$DEFAULT_UDEV_NET_SETUP_LINK_DIR"/80-net-setup-link.rules ]]; then
+		echo "$DEFAULT_UDEV_NET_SETUP_LINK_DIR/80-net-setup-link.rules found missing, investigation needed to understand " \
+			"if its location has changed or if the file itself has been replaced with another udev rules file" >&2
+		exit 1
+	fi
+	if [[ "$(get-appliance-platform)" == "aws" ]]; then
+		# If there are any changes in the specific line of the rules file that we are modifying, we fail early
+		# to allow us the opportunity to investigate the changes and update our own changes accordingly.
+		grep "^NAME==\"\", ENV{ID_NET_NAME}\!=\"\", NAME=\"\$env{ID_NET_NAME}\"" \
```
prakashsurya, line 96: instead of using `grep`.. could we instead modify the `sed` command to be more specific about the line we're expecting? I don't feel strongly, so I'm fine as-is.. but just curious if we could simplify the logic and remove `grep`, but still maintain the "fail if the line isn't what we expect" semantics..

files/aws/lib/systemd/system/systemd-networkd.service.d/20-override.conf
```diff
@@ -0,0 +1,29 @@
+#
```
prakashsurya, line 1: it looks like this file is present twice.. `20-override.conf` and `aws-override.conf`.. is this intentional?

files/aws/lib/systemd/system/systemd-networkd.service.d/aws-override.conf
```diff
@@ -0,0 +1,28 @@
+#
```
prakashsurya, line 1: it looks like this file is present twice.. `20-override.conf` and `aws-override.conf`.. is this intentional?

files/aws/usr/bin/update_netplan_to_mac_address
```diff
@@ -0,0 +1,57 @@
...
+	cur_timestamp=$(date +%y%m%d_%h%m%s)
+	cp -p "$DLPX_NETPLAN_DIR""$DLPX_NETPLAN_NAME" /var/tmp/"$DLPX_NETPLAN_NAME"_"$cur_timestamp"
+
+	# Update the netplan file with MAC based interface names
+	while IFS= read -r line; do
+		ifname=$(echo "$line" | awk '{print $1}')
+		macname=$(echo "$line" | awk '{print $2}' | tr -d ":")
+		macname="enx""$macname"
+		if grep -q "$ifname:" "$DLPX_NETPLAN_DIR""$DLPX_NETPLAN_NAME"; then
+			echo "Modifying $ifname to $macname in the netplan file" >>/var/log/syslog 2>&1
```
prakashsurya, line 44: I think `logger` is preferred to writing to the system logic.. can we use that instead?

files/aws/usr/bin/update_netplan_to_mac_address
```diff
@@ -0,0 +1,57 @@
...
+			sed -i "s/$ifname/$macname/g" "$DLPX_NETPLAN_DIR""$DLPX_NETPLAN_NAME" ||
+				die "Could not modify $ifname to $macname in the netplan file"
+		fi
+	done <"$INTERFACE_TO_MACADDRESS_MAP_FILE"
+	echo "Updated all the interface names to mac address based names" >>/var/log/syslog 2>&1
```
prakashsurya, line 49: I think `logger` is preferred to writing to the system logic.. can we use that instead?

Also, why not just write to stdout instead, which should get captured by the systemd service journal.. so we could see it when we do things like `journalctl -u systemd-networkd`.. e.g. all the `die` statements will not go to syslog, they'll go to the service journal.. so maybe it'll be more consistent to just emit all the output to the service journal?

debian/postinst
naveen-singh-delphix: I checked on ESX and this file is present. I could check on OCI and Azure but will need to figure out how to create an engine on these platforms.

The dkpg query for this file does not return a package name. Is there another way to check out the package name ?
```
delphix@DLPX-75209-TEST-20220912-1:~$ dpkg -S /usr/lib/udev/rules.d/80-net-setup-link.rules 
dpkg-query: no path found matching pattern /usr/lib/udev/rules.d/80-net-setup-link.rules
```

prakashsurya: Hm, yea.. I'd have to do some digging to see how/when it's generated.. seems to be some dynamically generated file.. I guess I'm OK landing without a double check on the other platforms.. presumably, if it doesn't exist, we'll catch that via a failed upgrade test on those platforms.

sebroy: @naveen-singh-delphix , the dpkg command isn't finding that path because the packaging defines it as `/lib/udev/...` (`/lib` is a symbolic link to `/usr/lib`):
```
root@guild:/export/home/delphix# dpkg -S /lib/udev/rules.d/80-net-setup-link.rules
udev: /lib/udev/rules.d/80-net-setup-link.rules
```

files/aws/lib/systemd/system/systemd-networkd.service.d/aws-override.conf
naveen-singh-delphix: Thanks for pointing this. 20-override.conf has been removed. I think I had deleted it earlier but looks like my steps were not correct.

debian/postinst
naveen-singh-delphix: Thanks @sebroy.

debian/postinst
naveen-singh-delphix: Yes, I have explained the reasoning behind calling update-initramfs. Does the comment look good in your opinion ? Or should more context be provided ?

debian/postinst
naveen-singh-delphix: Yes, this makes sense and we avoid an unnecessary check as you suggested.

debian/postinst
naveen-singh-delphix: yes, I have moved the initialization closer to the actual logic.

files/aws/usr/bin/update_netplan_to_mac_address
naveen-singh-delphix: Sure. I will study the usage of logger command and update.

files/aws/usr/bin/update_netplan_to_mac_address
naveen-singh-delphix: Sure, I will give this a try and update.

files/aws/lib/systemd/system/systemd-networkd.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
```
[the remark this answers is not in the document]
naveen-singh-delphix: Yes. this should not be visible now. My earlier attempts at deleting this file were not correct and hence this kept showing up.

files/aws/lib/systemd/system/systemd-networkd.service.d/20-override.conf
naveen-singh-delphix: The 20-override.conf has removed. Sorry for the confusion.

files/aws/usr/bin/update_netplan_to_mac_address
naveen-singh-delphix: As suggested in another comment, I have updated all the echo statements to write to STDOUT to enable all the output to be visible in one place, i.e. output of systemd journal file.

files/aws/usr/bin/update_netplan_to_mac_address
naveen-singh-delphix: I have updated all the echo statements to write to STDOUT to enable all the output to be visible in one place, i.e. output of systemd journal file.

prakashsurya (approved):

naveen-singh-delphix:
The following shellcheck error seems spurious to an extent. It suggests to remove the single quote before the search pattern in sed command. Replacing this single quote with double quotes results in the replacement NOT happening at all. Hence I am ignoring this recommendation for now.

In debian/postinst line 99:
		sed -i 's/NAME=="", ENV{ID_NET_NAME}!="", NAME="$env{ID_NET_NAME}"/NAME=="", ENV{ID_NET_NAME_MAC}!="", NAME="$env{ID_NET_NAME_MAC}"/' \
                       ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.

PR #395, opened 2022-11-07
DLPX-83249 Uploading upgrade image requires available space around 3x the size of the image

palash-gandhi:
### Problem
The current upload + unpack process involves these steps:
1. The front-end uploads the upgrade image in chunks. (1x space)
2. Once all chunks are uploaded, the backend stitches them together (2x space
since the stitch is not done in place)
3. The `unpack-image` script is called on the resulting tar file from step 2.
    a. This script first untars the file contents which include another tar file called
`payload.tar.gz`.
    b. The script then calls the `prepare` script from the unpacked contents which
is what untars `payload.tar.gz`
Step 3 causes space usage to increase to 3x due to the final untar operation.

While there is [logic in the java code](https://gitlab.delphix.com/app/dlpx-app-gate/blob/6.0/stage/appliance/server/upg/src/main/java/com/delphix/appliance/server/upg/impl/UpgradeManagerImpl.java#L1950) to delete the upgrade tar, it is only executed
after unpack-image finishes, which means that at some point during this process
the space consumption is almost 3x the size of the upgrade image.

### Solution
Delete the uploaded tar file after the initial unpack and after the checksums have been
verified but before calling the `prepare` script.

### Testing Done
Full upgrade blackbox tests on a VM with my bits: http://selfservice.jenkins.delphix.com/job/blackbox-chained/3555/

`git-ab-pre-push --test-upgrade-from=6.0/release`: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3715/

### Future Work
The work for [DLPX-81182](https://delphix.atlassian.net/browse/DLPX-81182) will involve adding a check before uploading the image to fail fast if there isn't enough space (at least 2x) in `rpool/update`.

### DIFF
--- a/files/common/usr/bin/unpack-image
+++ b/files/common/usr/bin/unpack-image
@@ -127,6 +127,8 @@ sha256sum -c SHA256SUMS >/dev/null ||
 
 popd &>/dev/null || die "'popd' failed"
 
+rm "$UPGRADE_IMAGE_PATH" || die "failed to remove $UPGRADE_IMAGE_PATH"
+
 report_progress_inc 50 "Handoff unpack to prepare script."
 
 "$UNPACK_DIR"/prepare ${opt_f:+"-f"} ${opt_s:+"-s"} ${opt_x:+"-x"} "$UNPACK_DIR"


prakashsurya (approved):
seems reasonable to me

sdimitro (approved):

sebroy:
Won't the error at https://gitlab.delphix.com/app/dlpx-app-gate/blob/6.0/stage/appliance/server/upg/src/main/java/com/delphix/appliance/server/upg/impl/UpgradeManagerImpl.java#L1954 always be emitted since the expected behavior of dx-unpack is to remove the file? Is there a separate review to remove the Java code that removes the file?

palash-gandhi:
> Won't the error at https://gitlab.delphix.com/app/dlpx-app-gate/blob/6.0/stage/appliance/server/upg/src/main/java/com/delphix/appliance/server/upg/impl/UpgradeManagerImpl.java#L1954 always be emitted since the expected behavior of dx-unpack is to remove the file? Is there a separate review to remove the Java code that removes the file?

Seb and I spoke offline and while the exception mentioned above only logs to the error file and is not thrown due to [throwIfInterrupt](https://gitlab.delphix.com/app/dlpx-app-gate/blob/6.0/stage/appliance/server/upg/src/main/java/com/delphix/appliance/server/upg/impl/UpgradeManagerImpl.java#L1955), he suggested that we should log something IF the upgrade tar is not deleted by the unpack-image script because that would mean that unpack-image did not successfully complete. Here's the app-gate change: http://reviews.delphix.com/r/85373/

PR #396, opened 2022-11-08
TOOL-16594 Don't remove delphix user from existing groups

prakashsurya:
This change modifies the Ansible logic that is used to configure the
delphix user, such that the Ansible task will no longer remove the user
from any groups it already exists in.

This change is needed so that appliance-build can modify the groups the
user is in, based on the variant the build is producing. Without this
change, the build could assign the delphix user to a non-standard group,
but that configuration would get overwritten by this Ansible task.


prakashsurya:
For a bit more context, in https://github.com/delphix/appliance-build/pull/705 we made it so that the `delphix` user would be part of the `docker` group.. it turns out, that configuration gets overwritten as a result of this delphix-platform Ansible task.. so that appliance-build change isn't working as intended.

sebroy (approved):

sdimitro (approved):

PR #399, opened 2022-11-17
TOOL-16805 Configure swap device on buildserver systems

prakashsurya:
We only want to run this when the system is actually booting, not during the build.


prakashsurya:
- `git-ab-pre-push -v internal-buildserver --no-tests` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3806/)

sebroy (approved):

sdimitro (approved):

PR #400, opened 2022-11-18
TOOL-16816 need a wrapper around the aws command that retries in case of segmentation faults

sebroy:
# Problem & Solution
See the Jira issue for details. After this PR is merged, the next step will be for @prakashsurya to finish the work that enables builds on a Delphix build server. After that, we can modify build automation to use this new command instead of the `aws` command directly.

# Testing Done

### Successful execution of an aws command:
```
delphix@guild:~$ ./dlpx-aws s3 ls dev-de-images/builds/jenkins-ops/appliance-build/6.0/stage/pre-push/431/upgrade-artifacts/
2022-11-18 07:48:32        181 MD5SUMS
2022-11-18 07:48:32        277 SHA256SUMS
2022-11-18 07:48:32 4817264640 external-standard.upgrade.tar
2022-11-18 07:48:32 4907243520 internal-dev.upgrade.tar
2022-11-18 07:48:32 4879472640 internal-qa.upgrade.tar
```

### Successful execution of an aws command after one or more segmentation faults:
This was simulated by sending signal 11 to the running aws command as it was running.
```
delphix@guild:~$ ./dlpx-aws s3 cp s3://dev-de-images/builds/jenkins-ops/appliance-build/6.0/stage/pre-push/431/upgrade-artifacts/internal-dev.upgrade.tar .
./dlpx-aws: line 24: 165468 Segmentation fault      (core dumped) aws "$@"
Attempt 1 failed, trying again after a small nap.
./dlpx-aws: line 24: 165524 Segmentation fault      (core dumped) aws "$@"
Attempt 2 failed, trying again after a small nap.
download: s3://dev-de-images/builds/jenkins-ops/appliance-build/6.0/stage/pre-push/431/upgrade-artifacts/internal-dev.upgrade.tar to ./internal-dev.upgrade.tar
delphix@guild:~$ ls -l ./internal-dev.upgrade.tar
-rw-r--r-- 1 delphix staff 4907243520 Nov 18 07:48 ./internal-dev.upgrade.tar
```

### Failure of an aws command after 5 attempts to run it (persistent segmentation faults):
```
delphix@guild:~$ ./dlpx-aws s3 cp s3://dev-de-images/builds/jenkins-ops/appliance-build/6.0/stage/pre-push/431/upgrade-artifacts/internal-dev.upgrade.tar .
./dlpx-aws: line 24: 165886 Segmentation fault      (core dumped) aws "$@"
Attempt 1 failed, trying again after a small nap.
./dlpx-aws: line 24: 165919 Segmentation fault      (core dumped) aws "$@"
Attempt 2 failed, trying again after a small nap.
./dlpx-aws: line 24: 165954 Segmentation fault      (core dumped) aws "$@"
Attempt 3 failed, trying again after a small nap.
./dlpx-aws: line 24: 165990 Segmentation fault      (core dumped) aws "$@"
Attempt 4 failed, trying again after a small nap.
./dlpx-aws: line 24: 166064 Segmentation fault      (core dumped) aws "$@"
Too many failed attempts, aborting.
```

### Retry loop only engages for segmentation faults, and not other failures:
In this case the exit status of the script must match the exit status of the aws command that was run.
```
delphix@guild:~$ ./dlpx-aws
usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help
aws: error: the following arguments are required: command
delphix@guild:~$ echo $?
2
```
In this example a SIGINT was sent to the process (ctrl-c):
```
delphix@guild:~$ ./dlpx-aws s3 cp s3://dev-de-images/builds/jenkins-ops/appliance-build/6.0/stage/pre-push/431/upgrade-artifacts/internal-dev.upgrade.tar .
cancelled: ctrl-c received
delphix@guild:~$ echo $?
1
```


prakashsurya:
As you pointed out, this only helps our build if we're building on Delphix.. which we're planning to do, so I guess that's OK.

But.. what about any time we're not running on Delphix? For example, our Jenkins logic that also uses `aws` and currently has a different mechanism for retries? Do we want a solution that'd work in that case too?

For example, perhaps we could write a python wrapper, and wrap it in a python package, so we don't necessarily need to run on Delphix to get this benefit? The python package would allow easy installation, via `pip` or equivalent (and stored in artifactory).

The Jenkins logic looks like this:
```
 54 def runAwsCommandWithRetries(awsCommand, awsCredentialsId, numRetries = 10) {   
...                                                        
 59   venvHelpers.withVenv('venv') {  
 ```
 i.e. it already runs inside a python virtual enviroment.. presumably that could be modified to use our python wrapper package.. and then, we could also modify our other tools (like linux-pkg and appliance-build) to also use the python package, instead of the normal `aws` tool (we might need to switch from installing from a debian package, to installing from a python package.. but I'd hope that's not too cumbersome)..
 
 What do you think? I'm OK with a solution that doesn't necessarily solve the issue for all use cases, but if we could, I think that'd be nice...
 
 With all that said, I don't really have experience creating, maintaining, and/or distributing an internal python package.. so that'd be a learning curve...
 
I'll go ahead and approve this as-is.. we can always adopt this for now, and move to a different solution if we wanted to change directions later.. if we stick with this for now, is there any way we could replace the previous `aws` command with ours? so consumers don't have to change anything (i.e. they could still use `aws` instead of `dlpx-aws`? e.g. maybe with `dpkg-divert` if necessary?

PR #404, opened 2023-01-05
DLPX-83991 Crash dump process provides no feedback

don-brady:
### Problem
When makedumpfile is generating a dump file, no progress is observed in
the system console.  Since this operation can take a long time, we should
provide some progress output.
### Solution
We can configure makedumpfile to provide progress output using the
message-level option.
### Testing Done
Tested by generating a panic with lots of dirty memory and observed that
the progress output is seen in the console.

Note that this was tested with DLPX-84173 which modifies makedumpfile to
output line feeds with the progress output so that the line buffering into
the console will show the progress.

console-progress

### Deployment Plan
This PR should land after DLPX-83991 which addresses the buffering of the
progress output in the console.


prakashsurya (approved):
nice!

sdimitro (approved):

tonynguien (approved):

PR #410, opened 2023-02-17
DLPX-84733 Add "apt-configure" so dev variant can install packages

prakashsurya:
On developer systems, the intention is for us to be able to install
packages via the APT package manager, by communicating with Ubuntu's
servers. We regressed on this functionality when landing 584d5ac, so
this change is adding it back.

### DIFF
--- a/files/common/etc/cloud/cloud.cfg.d/91-delphix-platform.cfg
+++ b/files/common/etc/cloud/cloud.cfg.d/91-delphix-platform.cfg
@@ -131,6 +131,7 @@ cloud_config_modules:
  - locale
  - grub-dpkg
  - apt-pipelining
+ - apt-configure
  - ntp
  - runcmd
 


prakashsurya:
FWIW, we control this for external systems via this setting in `91-delphix-platform.cfg` (to disable the use of Ubuntu's servers):
```
 40 #
 41 # Disable cloud-init from manipulating the APT sources. We don't want
 42 # the Delphix appliance to rely on packages from any 3rd party sources
 43 # (e.g. Ubuntu's APT repositories), so we need to ensure cloud-init does
 44 # not try to configure the APT sources to point to Ubuntu's servers.
 45 #
 46 apt:
 47   preserve_sources_list: true
 ```
 
 And then we override this, to enable it for developer systems in `99-delphix-development.cfg`:
 ```
 apt:
  preserve_sources_list: false
```

don-brady (approved):

sebroy (approved):

PR #415, opened 2023-03-29
DLPX-85348 Performance degradation observed in the time taken by MSSQL VDB operation jobs

sebroy:
The sgdisk command calls the sync() system call, which simply waits for dirty data to be written. It does not actually trigger zfs_sync() (i.e. a txg sync) to happen, which causes the sgdisk command to take at most 50 seconds to complete on an otherwise idle system (whereas before it would take at most 5 seconds).

This PR reverts the change that bumped the txg wait sync time from 5 to 50 seconds. We may be able to re-introduce that change if we figure out how to trigger a txg sync as part of the sync system call.

### DIFF
--- a/files/common/lib/modprobe.d/10-zfs.conf
+++ b/files/common/lib/modprobe.d/10-zfs.conf
@@ -115,10 +115,3 @@ options zfs zvol_threads=256
 # link creation can get up to the order of a minute or more.
 #
 options zfs zfs_vdev_open_timeout_ms=180000
-
-#
-# Reduce the frequency of txg sync from its default of 5 seconds. This avoids
-# sending unnecessary I/O to the underlying storage when the system is mostly
-# idle.
-#
-options zfs zfs_txg_timeout=50


grwilson (approved):

pcd1193182 (approved):
Why does the sgdisk command need to wait for a TXG to sync, but doesn't force a more prompt sync the way `snapshot` or other admin operations do? That seems like a better long-term solution, though of course this is a good short-term one.

sebroy:
@pcd1193182  The `sgdisk` command calls the `sync()` system call, which has the side-effect of sitting there and waiting for a txg sync. The question of why `sgdisk` needs to use the `sync()` syscall is a good one, who knows? The bigger problem to me is not figuring out how to fix `sgdisk`, but it's figuring out how to fix the `sync()` system call, as that seems broken across the board, and we just happened to notice that with our use of `sgdisk`.

tonynguien (approved):

PR #429, opened 2023-05-18
DLPX-85748 delphix-startup-screen does not display management service if it becomes inactive

nupur-agrawal-delphix:
<details open>

Delphix startup screen doesn't display inactive services. It displays failed services but not inactive services. Sometimes, this information is important for `enabled` services (services which always start while booting).
</details>

<details open>

Always display the state of enabled services (even though they are inactive).
</details>

<details open>

ab-pre-push: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/5502/console
Tested my manually stopping postgres service and ensuring postgres and mgmt service both are showing as inactive. 
Also tested my simulating a postgres service failure, which shows postgres service as failed and mgmt service as inactive.
</details>
Screen Shot 2023-05-19 at 1 47 09 PM

### DIFF
--- a/files/common/usr/bin/delphix-startup-screen
+++ b/files/common/usr/bin/delphix-startup-screen
@@ -186,31 +186,29 @@ def set_common_variables(stdscr) -> Tuple[int, int]:
     return X, Y
 
 
-def getstatus_pair(lst) -> Generator:
-    """
-    Process the status and return a list which is comprised of
-    ['service description', 'service status'].
-    """
-    for i in range(0, len(lst), 2):
-        yield lst[i:i + 2]
-
-
 def getstatus() -> str:
     """
-    Each service returns 2 lines -- one with the description and the
-    next line with the state.
+    Gets the status of each service, with each line containing service
+    state and description.
     """
     status = ""
-    for i in getstatus_pair(list(run_svcs("Description"))):
+
+    # `svcs_status` contains 3 elements for each service:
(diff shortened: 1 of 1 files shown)


grwilson (approved):
This looks good. I do wonder if we want to mask the rollback service since I believe it is mostly inactive. Might be worth checking with Anthony Checkas and Prakash.

prakashsurya:
> I do wonder if we want to mask the rollback service since I believe it is mostly inactive.

We discussed offline, and I think the path forward should be to remove the `delphix-rollback` service from the `delphix` target.

The `delphix` target is useful as a way to restart all "delphix" services easily, without having to list each service individually. This is especially useful for deferred upgrades; in that case, though, we don't actually want to (potentially) trigger a rollback, thus I don't think it really makes sense for the `delphix-rollback` service to get restarted.

Thus, if we remove `delphix-rollback` from the `delphix` target, it'll address that "problem".. while also removing it from the list shown here.. further, we wouldn't need any special handling of the `delphix-rollback` service in this script.

I think this change is good to land as-is, and I can open a follow on PR to remove `delphix-rollback` from `delphix.target`.

@grwilson does that sound good to you?

prakashsurya (approved):

prakashsurya:
> and I can open a follow on PR to remove delphix-rollback from delphix.target.

See here: https://github.com/delphix/dlpx-app-gate/pull/752

PR #434, opened 2023-05-30
DLPX-85893 run upgrade "execute" script from separate service

prakashsurya:
### Problem

When an upgrade fails while running the "execute" script, we don't have
a mechanism for the product to automatically recover.

For example, we've had cases where the OOM killer will kick in, killing
the virtualization service while it's running the "execute" script,
causing the entire upgrade to fail and requiring support intervention on
the system to recover.

### Solution

The solution we've discussed is to run the "execute" portion of the
upgrade in a standalone service, such that it wouldn't be killed when
the OOM killer targets the virtualization service, and such that it
could be automatically restarted in the event that it dies for some
other reason (e.g. a kernel panic).

While this change doesn't provide the full solution, it's a step in that
direction. This change adds a new script that can be used to run the
"execute" portion of the upgrade manually via the command line, but also
via the virtualization service, or a new upgrade-specific service that
we have yet to create.

The intention is for this new script to be used by the virtualization
service for now, and then eventually used by the yet-to-be-created
upgrade service. Until the upgrade service is created, this new script
can be kicked off using "systemd-run", such that it can be started by
the virtualization service much like today, but decoupled from the
virtualization service's cgroup limits (which will help prevent OOMs).

### Why?

This new script is required for a couple of reasons:

1. The current "execute" script doesn't handle things like updating the
   GRUB bootloader, nor does it handle restarting services. So, if we
   wanted to have a new service to "execute" the upgrade, the service
   would need to handle these additional steps. IMO, it makes sense to
   encapsulate all of these steps in a script like I've done here, so
   that when the new service is created, it can simply call this script.

2. Likewise, the virtualization service currently calls the "execute"
   script and needs to wait for it to complete, such that it can perform
   these additional tasks mentioned in the point above. By moving these
   additional tasks into this script, the virtualization service can now
   the script asynchronously via "systemd-run" (since it no longer needs
   to wait for the script to complete); which allows us to solve the
   problem incrementally.

### Related Work

- https://github.com/delphix/appliance-build/pull/725
- https://github.com/delphix/dlpx-app-gate/pull/773

### DIFF
--- a/files/common/usr/bin/execute-in-place-upgrade
+++ b/files/common/usr/bin/execute-in-place-upgrade
@@ -1,0 +1,222 @@
+#!/bin/bash
+#
+# Copyright 2023 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+UPDATE_DIR=${UPDATE_DIR:-/var/dlpx-update}
+
+function die() {
+	echo "$(basename "$0"): $*" >&2
+	exit 1
+}
+
+function usage() {
+	echo "$(basename "$0"): $*" >&2
+	echo "Usage: $(basename "$0") [-f] [-p] [-l | -v version]"
+	exit 2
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/5627/) and [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/5641/)

sebroy (approved):

palash-gandhi (approved):
Lgtm, the only concern I have with the reboot/restart happening in this new script is the difference in alerting which I have left a comment about in the dlpx-app-gate PR

prakashsurya:
Closing this after the discussion and changes being made here: https://github.com/delphix/appliance-build/pull/725

PR #436, opened 2023-06-14
DLPX-86534 CIS: idle shell timeout

skinner-m-c:
<details open>
Not having an idle timeout for a shell session creates an opertunity for attackers to make use of unattended shell sessions. While we do not provide a real shell offically for customers, many customers may unofficaly gain shell access. Support engineers can gain shell access. So with this change support engineers will be kicked out of idle shells.

</details>

<details open>
Set TMOUT in /etc/profile. Make the setting readonly so that it cannot be overriden to extend the idle timeout. 
</details>

<details open>
http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/5818/

This job created multiple images. The internal dev qa image was imported as part of the usual appliance build orchestrator job. I manually verified that `/etc/profile` did not contain a timeout setting. I then re-ran `delphix-build-and-snapshots` job at http://selfservice.jenkins.delphix.com/job/delphix-build-and-snapshots/job/ami-snapshots/4175/console
to import the external image variant into DCoA. I logged into the shell via the CRA process (register the engine at register.delphix.com, and then getting the challenge response code from auth.delphix.com) confirmed that `/etc/profile` in this variant does have the timeout setting. I then proceeded to wait five minutes and was kicked out of my SSH session.
</details>

<details open>
<summary><h2> Notes to Reviewers </h2></summary>

This five minute timeout may affect how support does their work and so the actual timeout may change after you have looked at this review depending on feedback from support.
</details>

### DIFF
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -262,7 +262,7 @@
 
 - blockinfile:
     path: /etc/profile
-    insertafter: BOF
+    insertafter: EOF
     block: |
       # Set a 5 minute idle timeout for all shells. After prompting for input
       # this timeout timer begins.
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -260,6 +260,23 @@
     #
     - variant is regex("external-.*")
 
+- blockinfile:
+    path: /etc/profile
+    insertafter: BOF
+    block: |
+      # Set a 5 minute idle timeout for all shells. After prompting for input
+      # this timeout timer begins.
+      TMOUT=300
+      readonly TMOUT
+      export TMOUT
+  when:
+    #
+    # For developer convenience, we only enable the shell timeout
+    # for external variants. The idle timeout can be a burden when we
+    # need to run long running processes over SSH on our internal
+    # systems (e.g. for development, testing, etc).
+    #
+    - variant is regex("external-.*")
 #
 # Harden the appliance by disabling SFTP.
 #
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ Log into that VM using the "ubuntu" user, and run these commands:
 
     $ git clone https://github.com/delphix/delphix-platform.git
     $ cd delphix-platform
+    $ sudo apt-get update && sudo apt-get install python3-docker
     $ ansible-playbook bootstrap/playbook.yml
     $ ./scripts/docker-run.sh make packages
 


sebroy (approved):

prakashsurya (approved):

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -260,6 +260,22 @@
     #
     - variant is regex("external-.*")
 
+- blockinfile:
```
prakashsurya, line 264: FWIW.. I think I'd prefer we deliver a file in `/etc/profile.d`, rather than using ansible to modify `/etc/profile`.. we could deliver that file via Ansible, but even better would be to deliver that file via dpkg (e.g. add it as a file delivered by our package)..

I'm not necessarily opposed to this approach, just wanted to mention perhaps a better way to do this.

skinner-m-c: I chose this way to avoid having a chance of the setting being overridden by other config files. Making it a file in `/etc/profile.d` would introduce a config file load order than would have to be maintained. I might might switch to using a separate config file  if we want to make it configurable in each shell session. Because the purpose of the setting comes from increasing security and passing CIS scans, making it configurable probably is not desirable though. I can foresee reasons for making it configurable but not for any security purposes.

prakashsurya: OK, sounds good.

sebroy: @matt-skinner-delphix , the reason why a `.d` snippet might be preferable doesn't have to do with configurability, but has to do with leaving the base packaged file intact so that it can continue to be updated by the package that provides it. If the base file is changed (e.g. as we're doing here with ansible), then if the package that provides `/etc/profile` ever changes that base file, those changes won't get applied to the file.

As you mention, though, one downside to the `.d` file is that the code that checks for this setting in the CIS benchmark might only look at /etc/profile, and not the fully baked environment (once all of the confguration snippets have been read).

sebroy: All of that said, I'm fine either way (and already approved what you have). There are pros and cons either way.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -260,6 +260,22 @@
...
+    path: /etc/profile
+    insertAfter: BOF
```
prakashsurya: should this be `EOF`, not `BOF`?

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
prakashsurya: having a file in `.d` for our config, also makes it easier for us to change that configuration.. e.g. the logic as written here, only adds the lines.. so, if we wanted to modify the configuration, we'd have to write new logic to remove the existing lines, and then use this logic to add the new lines.. doing this in Ansible becomes tedious, and easy to get wrong.. but only matters once/if we need to change it..

further, w.r.t. the ordering of `.d` files.. that ordering still exists with the `/etc/profile` file itself.. we "solved" this ordering by simply appending to the end of the `/etc/profile` file, but (IMO, at least) we could do something similar by adding a file named `99-delphix`.. i.e. the "99" prefix helps ensure our config is applied "last"..

generally, I think we prefer using `.d` files, as we've done for other configuration files.. e.g.
```
> find . -name '*.d' | head
./files/azure/etc/cloud/cloud.cfg.d
./files/oci/etc/cloud/cloud.cfg.d
./files/common/usr/lib/sysctl.d
./files/common/etc/sudoers.d
./files/common/etc/update-motd.d
./files/common/etc/default/grub.d
./files/common/etc/cloud/cloud.cfg.d
./files/common/lib/udev/rules.d
./files/common/lib/sysctl.d
./files/common/lib/modprobe.d
```

but I don't feel strongly.. and I'm good with this as-is.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
skinner-m-c: fixed

PR #441, opened 2023-06-21
TOOL-21734 appliance-build-stage1 fails because package locales is missing

sebroy:
This re-adds the `locales` package that was removed by #439. It looks like the
package needs to be on the build server (which we don't test), or on the engine
itself for not-in-place upgrades (which we don't test). The lack of testing of
both of these use cases is why this didn't get caught pre-commit in #439.

ab-pre-push: http://selfservice.jenkins.delphix.com/job/github/job/delphix/job/delphix-platform/job/appliance-build-orchestrator/job/pre-push/254/

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -65,6 +65,7 @@ DEPENDS += ansible, \
 	   kbd, \
 	   kmod, \
 	   less, \
+	   locales, \
 	   logrotate, \
 	   lsb-release, \
 	   mount, \


sumedhbala-delphix (approved):
does it make sense to test the change with all the variants and platforms similar to the nightly job http://ops.jenkins.delphix.com/job/appliance-build/job/develop/job/nightly/87/parameters/ ?

prakashsurya:
> or on the engine itself for not-in-place upgrades (which we don't test)

IIRC, we do run not-in-place upgrade tests via some cadence.. it's not pre-push, but I'm pretty sure they get run via some of the regression runs..

prakashsurya (approved):

prakashsurya:
for posterity's sake..

```
 40 #
 41 # In order for this locale to be used (e.g. by virtualization) we need
 42 # to ensure it's available. Generally this locale will be generated by
 43 # appliance-build for new VMs, but we can't rely on that since that
 44 # doesn't work for not-in-place upgrades. Thus, we must explicitly
 45 # generate the locale here.
 46 #
 47 - locale_gen:
 48     name: en_US.UTF-8
 49     state: present
 ```
this task isn't specific to the build.. but it looks like for a normal engine, we pull it in via the virtualization package:
```
$ aptitude why locales
i   delphix-virtualization Depends postgresql-12
i A postgresql-12          Depends locales | locales-all
```
hence why this task works when it's run via our normal pre-push testing..

PR #447, opened 2023-07-11
DLPX-86902 failed to upgrade from 13.0.0.0 to 14.0.0.0 because engine should allow deferred

prakashsurya:
In 25ebf3493ddbd6bd27887e72a527716a1c181ca4 we removed the `update-notifier-common` package, which provides the `/usr/share/update-notifier/notify-reboot-required` script. This script is used by kernel packages to create the `/var/run/reboot-required` file, which is used by our product's upgrade logic to determine if a `FINISH_DEFERRED` upgrade is required to reach the `RUNNING` state.

As a result, without the `notify-reboot-required` on the engine, the `/var/run/reboot-required` does not get created anymore.

This change adds the `update-notifier-common` back.

### DIFF
--- a/files/common/usr/share/update-notifier/notify-reboot-required
+++ b/files/common/usr/share/update-notifier/notify-reboot-required
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+if [ "$0" = "/etc/kernel/postinst.d/update-notifier" ]; then
+	DPKG_MAINTSCRIPT_PACKAGE=linux-base
+fi
+
+echo "*** System restart required ***" >/var/run/reboot-required
+echo "$DPKG_MAINTSCRIPT_PACKAGE" >>/var/run/reboot-required.pkgs
--- a/files/common/etc/kernel/postinst.d/update-notifier
+++ b/files/common/etc/kernel/postinst.d/update-notifier
@@ -1,0 +1,1 @@
+/usr/share/update-notifier/notify-reboot-required


palash-gandhi (approved):

prakashsurya:
Looks like `update-notifier-common` pulls in some unwanted dependencies, so I'm not sure we want to depend on that package like I was originally thinking..
```
$ apt depends update-notifier-common
update-notifier-common
...
  Depends: update-manager-core (>= 1:17.04.2)
...

$ apt depends update-manager-core
update-manager-core
...
  Depends: ubuntu-advantage-tools
...
```

The script we need looks simple, though, so I've tried to copy/paste it into the platform package.. lets see if this works..

david-mendez1 (approved):

files/common/usr/share/update-notifier/notify-reboot-required
```diff
@@ -0,0 +1,50 @@
...
+export TEXTDOMAINDIR=/usr/share/locale
+
+case "$DPKG_MAINTSCRIPT_PACKAGE::$DPKG_MAINTSCRIPT_NAME" in
+    linux-image-extra*::postrm)
+        exit 0;;
+esac
+
+if [ "$0" = "/etc/kernel/postinst.d/update-notifier" ]; then
+    DPKG_MAINTSCRIPT_PACKAGE=linux-base
+fi
```
palash-gandhi: Will this code ever run? I know this PR is simply copying over the script from upstream as-is but I wonder what happens if future upstream changes break this script.

prakashsurya: as is, probably not.. but with the `update-notifier-common` package installed, there's a `/etc/kernel/postinst.d/update-notifier` file that's a symlink to `/usr/share/update-notifier/notify-reboot-required`.. and I think the scripts in `/etc/kernel/postinst.d` get execute when the kernel is upgraded (which is how the `reboot-required` file is created when a new kernel is installed)..

in this PR so far, I'm not creating that symlink.. but maybe I should? I thought about creating it, to remain consistent, and probably let it work for kernel upgrades too (the ZFS postinst hook calls `/usr/share/update-notifier/notify-reboot-required` directly, so practically, it probably doesn't matter for our appliance since we update ZFS every release)..

prakashsurya:
I did an upgrade from 13, to 14 + this change, and verified the `/var/run/reboot-required` file was present, and the delphix version was correctly showing `DEFERRED`.. I also verified without this change, doing the same test, the version would show as `RUNNING`.

```
$ cat /var/run/reboot-required.pkgs
zfs-modules-5.15.0-1031-dx2023062919-7f5e6edd6-aws
linux-base

$ cat /var/run/reboot-required
*** System restart required ***

$ get-appliance-version
14.0.0.0-snapshot.20230712152225225+jenkins-selfservice-appliance-build-develop-pre-push-788
```

files/common/usr/share/update-notifier/notify-reboot-required
palash-gandhi: But should we be concerned about copying a script from a third party package instead of actually depending on the package? By copying the script, aren't we opening ourselves up to the possibility of an upstream change breaking the script? Such breakages wouldn't happen if we simply depended on the third party package.
I guess both options have trade-offs. Depending on the third-party package bloats our product, but not adding it adds risk of breakages. If I were to pick, the added bloat is more acceptable than the latter.

prakashsurya: > If I were to pick, the added bloat is more acceptable than the latter.

it's not just bloat.. if we pull in `update-notifier-common`, we'll also pull in `ubuntu-advantage-tools`.. and the `ubuntu-advantage-tools` package has caused bugs, e.g. ESCL-4493..

but even if we fix those bugs, such that it is just "bloat".. I'd argue implementing a solution that works and doesn't have such bloat, would be a better solution.. and implementing our own script to create the `/var/run/reboot-required` file, and not having the bloat, is better..

but I'm open to other input here.. if others also think depending on the package, instead of rolling our own script, and inheriting the bloat that comes with that, is a better solution.. I'd be OK with that.

I was assuming the bloat was not acceptable, since we've been trying to reduce that bloat; e.g. https://github.com/delphix/delphix-platform/pull/439 (the change that introduced this breakage).

> By copying the script, aren't we opening ourselves up to the possibility of an upstream change breaking the script?

generally, sure.. in this specific case, though, I'm not concerned..

for example, we're adding a script at a specific location on the root filesystem.. since we're not pulling in the package that would have also (tried to) put the script at that location, we don't have any conflicts, so I think we're safe to install our version of the script.. arguably, the `delphix-platform` package should now conflict with the `update-notifier-common` package, to help the package manager prevent both package from being installed..

so I think we're safe to install our own script at the same location..

now the contents of the script are currently being copy/pasted as you're pointing out.. but I don't see anything in the script that would break over time.. is there any specific parts of the script that you're concerned about? e.g. anything you think will break due to an upstream change?

with that said, I also think most of the script is useless for us, and really all we need is the last two lines.. i.e.
```
echo "*** $(eval_gettext "System restart required") ***" > /var/run/reboot-required
echo "$DPKG_MAINTSCRIPT_PACKAGE" >> /var/run/reboot-required.pkgs
```
so if you feel strongly, I could modify our version of the script to just write to these two files.. I think that'd also accomplish our goals..

even further, I think we could modify the ZFS package's `postinst` script to create the `/var/run/reboot-required` file directly, and not call this script.. since we always update the ZFS kernel modules in each release, I think this would work.. although, might not be the most elegant solution.. if we did this, we wouldn't even need to create this script..

Let me know what you think.. I see a few valid options:

1. depend on `update-notifier-common`, causes bloat (assume bugs are fixed)
2. copy/paste like I've done here, might break over time (I'm still doubtful, but breakage could be fixed)
3. don't copy/paste, rewrite script to be more simple; could this also break over time?
4. modify ZFS postinst script to create `reboot-required` directly (kind of awkward, as it's delphix specific)

IMO, if we're really concerned that copy/paste won't work moving forward, I think I'd opt for (3) rather than (1).. but I'm open to other opinions.

palash-gandhi: >anything you think will break due to an upstream change?

I was mostly referring to any upstream change that changes the functionality or the command that this script depends on. Nothing specific.
I forgot about ESCL-4493. I personally like option 3 which at least reduces chances of breakages.

prakashsurya: > I was mostly referring to any upstream change that changes the functionality or the command that this script depends on.

sure.. I just don't see anything that the script depends on that'd break..

prakashsurya: > I personally like option 3 which at least reduces chances of breakages.

I pushed an update, let me know what you think.

palash-gandhi (approved):

PR #449, opened 2023-07-13
DLPX-86533 CIS: default umask

abhi2196:
<details open>

```
(5.76) 4726 Current 'UMASK' setting for the '/etc/bashrc or /etc/bash.bashrc' file

The 'default UMASK permissions' determine what [default] privilege level will be set upon directories and files created by the user. The usual manufacturer default is '022.' If set at this value, when creating a new file, the resulting default permissions will be '644' (666 minus 022, i.e. -rw-r--r--). When creating a new directory, these default permissions will be 755 (drwxr-xr-x), which sets the access level to (rwx r-x r-x): for owner (rwx), group (read/execute), other (read/execute) access on file access. If users are not properly restricted, sensitive system or business information may be improperly disclosed. The most restrictive setting is 077 and your default file permissions would be 600 (-rw-------) and your default directory permissions would be 700 (drwx------), thus becoming (rwx --- ---): for owner (rwx), group (no access), other (no access). Also, as a malicious user could lay the groundwork for a privilege escalation attack by changing the UMASK value in this configuration file, access to this file and its UMASK setting should be restricted appropriately.

Remediation: # Edit file '/etc/bash.bashrc' to configure 'UMASK' setting according to the business needs and organization's security policies.

UMASK <number>

# Example UMASK 027

(5.77) 12884 Status of 'umask' setting in /etc/profile and /etc/profile.d/*.sh files

The default umask determines the permissions of files created by users. The user creating the file has the discretion of making their files and directories readable by others via the chmod command. Users who wish to allow their files and directories to be readable by others by default may choose a different default umask by inserting the umask command into the standard shell configuration files ( .profile , .bashrc , etc.) in their home directories. Setting a very secure default value for umask ensures that users make a conscious choice about their file permissions. A default umask setting of 077 causes files and directories created by users to not be readable by any other user on the system. A umask of 027 would make files and directories readable by users in the same Unix group, while a umask of 022 would make files readable by every user on the system. This setting should be configured according to the needs of the business.

Remediation:

Run the following command to verify the 'umask' setting in '/etc/profile and /etc/profile.d/*.sh' files. $sudo grep "umask" /etc/profile /etc/profile.d/*.sh
umask 027

Edit the 'etc/profile and /etc/profile.d/*.sh' file to configure 'umask' setting according to the business needs and organization's security policies.
umask <permissions>

# Example umask 022
```
</details>

<details open>

- Added a file `/etc/profile.d/set-umask-for-all-users.sh` to set umask to `022` also added the same in `/etc/bash.bashrc` file as well.
</details>

<details open>

- `ab-pre-push`: http://selfservice.jenkins.delphix.com/job/github/job/delphix/job/delphix-platform/job/appliance-build-orchestrator/job/pre-push/282/console `Failed due to issue in yml syntax`
- `ab-pre-push`: http://selfservice.jenkins.delphix.com/job/github/job/delphix/job/delphix-platform/job/appliance-build-orchestrator/job/pre-push/291/ `Passed`
- `Commit 2`, `ab-pre-push`: http://selfservice.jenkins.delphix.com/job/github/job/delphix/job/delphix-platform/job/appliance-build-orchestrator/job/pre-push/300/ `Passed`
- Created an engine and validated that umask value is getting updated for the users.
- `Commit 3`, `ab-pre-push`: http://selfservice.jenkins.delphix.com/job/github/job/delphix/job/delphix-platform/job/appliance-build-orchestrator/job/pre-push/303/ `Passed`
</details>

### DIFF
--- a/files/common/etc/profile.d/set-umask-for-all-users.sh
+++ b/files/common/etc/profile.d/set-umask-for-all-users.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-umask 022
+umask 027
--- a/files/common/etc/profile.d/set-umask-for-all-users.sh
+++ b/files/common/etc/profile.d/set-umask-for-all-users.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+umask 022
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -682,20 +682,11 @@
     - variant == "internal-buildserver"
     - not ansible_is_chroot
 
+#
 # CIS: Set default umask (DLPX-86533)
-# We need to set default umask as 022 in the /etc/profile and /etc/bash.bashrc files,
+# We need to set default umask as 022 in the /etc/bash.bashrc file,
 # so that the same can be applied for all the users on the engine.
-- blockinfile:
-    path: /etc/profile.d/set-umask-for-all-users.sh
-    create: yes
-    block: |
-      #!/bin/sh
-      umask 022
-
-- file:
-    path: /etc/profile.d/set-umask-for-all-users.sh
-    mode: '0755'
-
+#
 - blockinfile:
     path: /etc/bash.bashrc
     block: |
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -681,3 +681,23 @@
   when:
     - variant == "internal-buildserver"
     - not ansible_is_chroot
+
+# CIS: Set default umask (DLPX-86533)
+# We need to set default umask as 022 in the /etc/profile and /etc/bash.bashrc files,
+# so that the same can be applied for all the users on the engine.
+- blockinfile:
+    path: /etc/profile.d/set-umask-for-all-users.sh
+    create: yes
+    block: |
+      #!/bin/sh
+      umask 022
+
+- file:
+    path: /etc/profile.d/set-umask-for-all-users.sh
+    mode: '0755'
+
+- blockinfile:
+    path: /etc/bash.bashrc
+    block: |
+      # Set default umask value.
+      umask 022
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -684,11 +684,11 @@
 
 #
 # CIS: Set default umask (DLPX-86533)
-# We need to set default umask as 022 in the /etc/bash.bashrc file,
+# We need to set default umask as 027 in the /etc/bash.bashrc file,
 # so that the same can be applied for all the users on the engine.
 #
 - blockinfile:
     path: /etc/bash.bashrc
     block: |
       # Set default umask value.
-      umask 022
+      umask 027


files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -681,3 +681,23 @@
...
+# so that the same can be applied for all the users on the engine.
+- blockinfile:
+    path: /etc/profile.d/set-umask-for-all-users.sh
+    create: yes
+    block: |
+      #!/bin/sh
+      umask 022
+
+- file:
+    path: /etc/profile.d/set-umask-for-all-users.sh
```
sebroy: This really should be delivered as a stand-alone file rather than constructed in an ansible task (e.g. `files/common/etc/profile.d/set-umask-for-all-users.sh`). The reason is primarily that once this logic is there, if we ever need to change the contents of the file, the ansible logic becomes more complex to handle upgrade cases. If you have to modify an existing file, then doing that in ansible seems like the logical choice, but in this case, this file doesn't exist yet...

sebroy (changes requested):

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
abhi2196: Sure, I have moved it to a separate file at `files/common/etc/profile.d/set-umask-for-all-users.sh`

files/common/etc/profile.d/set-umask-for-all-users.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+umask 022
```
sebroy: 022 is the current default umask. I think CIS checks are looking for 027. Can you verify that?

abhi2196: Actually, the CIS check report says,

```
Remediation: 
Edit file '/etc/bash.bashrc' to configure 'UMASK' setting according to the business needs and organization's security policies.

Edit the 'etc/profile and /etc/profile.d/*.sh' file to configure 'umask' setting according to the business needs and organization's security policies.
```

So we need to decide what's the best value from the security point of view and also from the usability point of view that will not break things.

 `027` looks good to me as it will provide access to groups as well.

So do I need to get this reviewed or approve? OR I can go ahead and change it to `027`?

sebroy: Let's run this by @cprice-delphix for confirmation, but my preference would be to ship with a more restrictive/secure default umask (`027`). This will also potentially address a number of other CIS findings around permissive file permissions (e.g. in `/var/log` and other places) for files that happened to be created using the default umask.

cprice-delphix: The CIS benchmark PDF defines the `umask` test as:
> 5.5.4 Ensure default user umask is 027 or more restrictive (Automated)

So I'd expect this to address the finding.

abhi2196: Thanks for the confirmation, I will make that change.

sebroy:
Note to @dbj-delphix and @prakashsurya, this change has a chance of negatively impacting variants that aren't covered in ab-pre-push automated testing:
1. dcol
2. the build server

If you have recommendations on how to mitigate the risks here, please do suggest them. I don't see an immediate risk to dcol given that this would only impact new deployments, but the build server image is frequently re-built. That's the one I'm most concerned about.

prakashsurya:
@sebroy if you're suggesting/asking if we could generate a buildserver with this change, and then rebuild all packages with that buildserver, I'm sure we can accomplish that.. Iet me see how we might do that..

prakashsurya:
> Iet me see how we might do that..

as-is, we'd have to build the variant via `git-ab-pre-push -v internal-buildserver`, and then use a modify devops-gate to test the builds on the resultant image.. I'll look into making changes to devops-gate, to allow the buildserver image to be passed into the jobs.. today, we don't support that, but no reason we couldn't, and enable us to more easily pre-push test the buildserver image and workload..

prakashsurya:
> I'll look into making changes to devops-gate, to allow the buildserver image to be passed into the jobs..

I've opened: https://github.com/delphix/devops-gate/pull/1524

prakashsurya:
`git-ab-pre-push -v internal-buildserver -p aws --no-tests` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/6279/)

prakashsurya:
Running the builds with this change [here](http://selfservice.jenkins.delphix.com/job/linux-pkg/job/develop/job/build-packages/job/pre-push/903/)

sebroy (approved):

VenkatanadhanG (approved):

PR #455, opened 2023-07-26
DLPX-87132 Revert "DLPX-86533 CIS: default umask"

colinwkingen:
…7-7a57-459e-af31-863ef36f1721"

This reverts commit 73297c9c320083ab19e7b4d77d8e0c3085c3d469, reversing changes made to 5a98530dafd6e05e7dfd7b073e4dcec523097058.

### Problem

https://delphix.atlassian.net/browse/DLPX-87132

Change caused masking update failures.

### DIFF
--- a/files/common/etc/profile.d/set-umask-for-all-users.sh
+++ b/files/common/etc/profile.d/set-umask-for-all-users.sh
@@ -1,3 +1,0 @@
-#!/bin/sh
-
-umask 027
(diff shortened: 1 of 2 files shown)


sonamkindy (commented):
This looks good, but you need to update the commit message to:
`DLPX-87132 Revert "DLPX-86533 CIS: default umask"` so the jira pr check passes.

You probably also want to update the solution to something like revert the change so masking engineers and automation are unblocked; the change can be reintroduced at a later date.

sebroy (approved):

PR #458, opened 2023-08-09
DLPX-87360 reduce deadman timer to quickly detect hung socket connection

grwilson:
This is the companion PR to https://github.com/delphix/zfs/pull/1097

### DIFF
--- a/files/common/lib/modprobe.d/10-zfs.conf
+++ b/files/common/lib/modprobe.d/10-zfs.conf
@@ -122,3 +122,12 @@ options zfs zfs_vdev_open_timeout_ms=180000
 # VDB errors (e.g. MSSQL VDB refresh failing).
 #
 options zfs zvol_open_timeout_ms=60000
+
+#
+# The deadman timer is typically used to detect if an I/O is hung.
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):

files/common/lib/modprobe.d/10-zfs.conf
```diff
@@ -122,3 +122,12 @@ options zfs zfs_vdev_open_timeout_ms=180000
...
 options zfs zvol_open_timeout_ms=60000
+
+#
+# The deadman timer is typically used to detect if an I/O is hung.
+# On object store pools, the deadman timer check for a hung socket
+# connections between the kernel and the agent. Reducing this
+# to 1 minute will allow the kernel to detect this condition quickly
+# and restart the agent if the connection is hung.
+#
+options zfs_deadman_synctime_ms=60000
```
prakashsurya: I don't think the syntax is right.. I think you need `options zfs zfs...`

sebroy (approved):
No concerns other than the thing that Prakash already pointed out.

PR #459, opened 2023-08-11
DLPX-87464 Add netcat package to list of dependencies

prakashsurya:


sebroy (commented):
What set of new dependencies does this bring in (if any)?

sebroy:
Based on internal discussions, it's not clear that there's sufficient justification for adding this to the image.

debian/rules
```diff
@@ -151,6 +151,7 @@ DEPENDS += aptitude, \
 	   memstat, \
 	   mtr-tiny, \
 	   ncdu, \
+	   netcat, \
```
sebroy: I think the netcat package is actually called `netcat-openbsd`. The information on the plain `netcat` package says:

> This is a transitional package that depends on the netcat-openbsd
> implementation of netcat, preferred in Ubuntu.  It may be safely removed.

sebroy (changes requested):

prakashsurya:
> Based on internal discussions, it's not clear that there's sufficient justification for adding this to the image.

Should I go ahead and close this as "won't fix"?

sebroy:
> Should I go ahead and close this as "won't fix"?

It looks like there may still be some use cases not addressed by our APIs. Let's go ahead and move forward with this, and we can file separate Jira issues representing the gaps in functionality.

prakashsurya:
I ran the following command before and after this change:
```
$ dpkg-query -Wf '${Package}\n' | sort
```
Here's the diff:
```
$ diff pkgs-before.txt pkgs-after.txt
495a496,497
> netcat
> netcat-openbsd
```
looks like it's not pulling in any other dependencies.

I'll make the change from depending on `netcat` to `netcat-openbsd`.. I think we're good otherwise.

david-mendez1 (approved):

sebroy (approved):

PR #474, opened 2024-02-29
DLPX-86530 CIS: delphix user lockout after failed login attempts

rupalimatkar:
Background
=========
CIS: delphix user lockout after failed login attempts

JIRA: https://delphix.atlassian.net/browse/DLPX-86530

```
Status of the 'deny' setting for pam_tally2.so module in /etc/pam.d/common-auth file
Located in the /etc/pam.d directory, the 'pam_tally2.so' module allows administrators to manage user login security policy and monitor user login activity. The 'deny' parameter in the 'pam_tally2.so' module sets the number of failed login attempts allowed prior to account lockout. As a malicious user can use brute force attacks to compromise user accounts, account lockout policies mitigate this risk by restricting failed login attempts. The 'deny' parameter in the 'pam_tally2.so' module should be set in accordance with needs of the organization.

Remediation: Edit the /etc/pam.d/common-auth file and add the auth line below:
auth required pam_tally2.so onerr=fail audit silent deny=5 unlock_time=900
```

Solution
======
Updating the pam modules `common-auth and common-account` to enforce `delphix` user lockout policies using `pam_tally2.so` 

Testing Done
======
https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/10988/

**With 4 unsuccessful login attempts and 5th successful login attempt with `delphix` user -**

image

**With 5 unsuccessful login attempts and entering correct password at 6th attempt with `delphix` user -** 

`delphix` user lockout already happened and after unlock time completion connection to engine will be successful with correct password.

image

**Re-login to engine with delphix user after unlock period is over -** 

image

### DIFF
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -337,6 +337,42 @@
     replace: '\1'
 
 #
+#
+# Lock out the user after an unsuccessful consecutive login attempts.
+#
+- lineinfile:
+    path: /etc/pam.d/common-auth
+    line: "{{ item }}"
+    insertbefore: '^auth\s+\[success=1\s+default=ignore\]\s+pam_unix\.so\s+nullok\s+try_first_pass'
+  with_items:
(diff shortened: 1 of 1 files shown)


sebroy:
Does this work on an external variant with CRA enabled?

PR #477, opened 2024-04-18
DLPX-89763 DLPX-86523 delphix-platform changes

justsanjeev:
<details open>

CIS is looking or a single home directory filesystem mounted at the `/home` location, currently we have the home dataset is mounted on `/export/home`

Due to that we see the below issues in the CIS Report

- (1.45) 7402 Status of the '/home' partition in the '/etc/fstab' file
- (1.46) 13248 Status of Mount Partition '/home' using mount command
- (1.47) 7403 Status of the 'nodev' mount option setting for the '/home partition' defined in the '/etc/ fstab' file
- (1.48) 14601 Status of the 'nodev' option for '/home' partition using 'mount' command

</details>

<details open>

- Mounting the home dataset to `/home`.
- Post upgrade creating a softlink `/export/home` to `/home` so that the exiting tests and dependencies have minimal impact.

With implemented a a script which takes care of a soft link creation with files/common/var/lib/delphix-platform/export-home. This ensure. a softlink from `/export/home` to `/home`. This helps to ensure that most of the flows are passed with this new implementation.

</details>

<details>

Refer testing section in https://github.com/delphix/appliance-build/pull/756

</details>


files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -22,7 +22,7 @@
 # it below; otherwise that task will fail.
 #
 - file:
-    path: /export/home
+    path: /home
```
prakashsurya, line 25: since we're changing the user's home directory location here, we might want to use this file as the place that we change the mountpoint for the home directory dataset too..

I have some concerns if we actually change the mountpoint "live" on the running systems, due to potential processes currently using `/export/home`..

we might want to only update `/etc/fstab` to point to the new location, while leaving the existing `/export/home` mountpoint intact.. and then leverage a future reboot to "move" the mountpoint from `/export/home` to `/home`.. i.e. just mount at `/home` after the reboot..

but, if we do that, and change the user's home directory to be `/home`, we might run into conflicts.. e.g. `/export/home` will be mounted, while the user expects it to be in `/home`.. at least until the next reboot..

might need to think about how to do the "transition" properly, while processes are still running..

justsanjeev: Resolved this by adding the `preinst` script as per our discussions. The mount point for home dataset is being changed to `/home` during upgrade and that too we have done by checking the existence  in the `/etc/fstab`, if this entry exist with `/export/home` .

debian/preinst
```diff
@@ -0,0 +1,48 @@
...
+		sed -i 's|/export/home|/home|g' "$fs_tab"
+		mount /home
+	fi
+
+	if [[ -e $auto_master ]]; then
+		if grep -q "\/home\s+auto_home\s+-nobrowse" "$auto_master"; then
+			sed -i 's|/home   auto_home    -nobrowse|#/home   auto_home    -nobrowse|g' "$auto_master"
+			systemctl restart autofs
+		fi
+	fi
```
sebroy: Why is this here?

debian/preinst
```diff
@@ -0,0 +1,48 @@
...
+# limitations under the License.
+#
+
+case $1 in
+upgrade)
+	# Checking the fstab file if the /export/home entry
+	# is present in the /etc/fstab, In case of container
+	# upgrade the file is already changed by the
+	# container-upgrade script and we dont need to do
+	# it again.
```
sebroy: I'm confused with the wording in this comment. There is also no script called "container-upgrade". If this is referring to the code from the "upgrade-container" script, that script sets up the container used for upgrade verification. I don't think it's relevant here. I might change this to something like:

Home directories were previously mounted under /export/home, and this was changed to /home. This is the upgrade logic that updates the /etc/fstab file to reflect that change. Home directories will be mounted in both /export/home and /home until the system is rebooted to ensure that running processes referencing the old /export/home paths continue to function while also enabling new logins under /home to work.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -738,3 +738,53 @@
...
+  ansible.builtin.shell: |
+    mount | grep /export/home
+  register: mount_status
+  ignore_errors: yes
+
+- name: Unmount the path if it is mounted
+  ansible.builtin.mount:
+    path: /export/home
+    state: unmounted
+  when: mount_status.rc == 0
```
sebroy: This is going to fail if there are any running processes with references to any files under /export/home. This feels very fragile. What I would recommend doing instead is:
1. Keep /export/home mounted. Since it's been changed in /etc/fstab, it simply won't get mounted at the next reboot.
2. At boot, since /export/home won't be mounted at that point, create the symbolic link then. I think this could be as simple as creating the symbolic link (if it doesn't already exist) in the delphix-platform service's start method.

This would remove the need for all of this new ansible logic, and harden the system against random upgrade failures due to unexpected processes holding the mount hostage.

sebroy (changes requested):

debian/preinst
justsanjeev: I have changed the comment, this is a better version. Thank you for the suggestion.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
justsanjeev: Thank you @sebroy for this point, Validated this and this is fragile.
- Created an app data unstructured vdb on a target
- Upgraded the engine
- delphix-mgmt did not come up due to delphix-platform

Screenshot 2025-04-09 at 5 24 33 PM

- delphix-platform did not come up because of the change where we tried to unmount /export/home

Screenshot 2025-04-11 at 11 25 16 AM

- Now rebooted the Engine to validate, after reboot
- Services came up successfully

- /export/home in not mounted

Screenshot 2025-04-11 at 11 26 59 AM

justsanjeev:
We have made the latest changes here and added the below logic to `delphix-platform.service` to take care of the `/export/home` mount point and soft link  `/export/home -> /home`
````
# Create /export/home symlink to /home if it doesn't already exist
ExecStartPost=/bin/sh -c '\
  echo "Ensuring /export/home is a symlink to /home..."; \
  if mountpoint -q /export/home; then \
    echo "/export/home is a mountpoint, unmounting..."; \
    umount /export/home || { echo "Failed to unmount /export/home"; exit 1; }; \
  fi; \
  if [ -e /export/home ] && [ ! -L /export/home ]; then \
    echo "Removing existing /export/home directory..."; \
    rm -rf /export/home; \
  fi; \
  if [ ! -d /export ]; then \
    mkdir /export; \
  fi; \
  if [ -d /export ] && [ ! -L /export/home ]; then \
    echo "Creating symlink: /export/home -> /home"; \
    ln -s /home /export/home; \
  fi;'

````

**We this change we got a build** : [appliance-build-orchestrator-pre-push/10842](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/10842/console). 

- Here we have the successful upgrade tests.
- But to verify @sebroy 's comment, I again the same testing 

-- Created an app data unstructured vdb on a target
-- Upgraded the engine
-- delphix-mgmt did not come up due to delphix-platform
Screenshot 2025-04-16 at 9 16 19 PM

Also my /var/log/syslog show 
````
Apr 15 12:04:11 ip-10-110-212-88 dynamic-debug[156701]: init/main.c:1218 [main]initcall_blacklisted =p "initcall %s blacklisted\012"
Apr 15 12:04:11 ip-10-110-212-88 dynamic-debug[156701]: init/main.c:1179 [main]initcall_blacklist =p "blacklisting initcall %s\012"
Apr 15 12:04:11 ip-10-110-212-88 sh[156702]: Ensuring /export/home is a symlink to /home...
Apr 15 12:04:11 ip-10-110-212-88 sh[156702]: /export/home is a mountpoint, unmounting...
Apr 15 12:04:11 ip-10-110-212-88 sh[156704]: umount: /export/home: target is busy.
Apr 15 12:04:11 ip-10-110-212-88 sh[156702]: Failed to unmount /export/home
Apr 15 12:04:11 ip-10-110-212-88 systemd[1]: delphix-platform.service: Control process exited, code=exited, status=1/FAILURE
Apr 15 12:04:11 ip-10-110-212-88 systemd[1]: delphix-platform.service: Failed with result 'exit-code'.
Apr 15 12:04:11 ip-10-110-212-88 systemd[1]: Failed to start Delphix Appliance Platform Service.
````

Upon executing below 
````
delphix@ip-10-110-212-88:~$ sudo systemctl enable delphix-platform.service
delphix@ip-10-110-212-88:~$ sudo systemctl restart delphix-mgmt
````

And my log shows that the services has started successfully

````
Apr 15 12:37:42 ip-10-110-212-88 dynamic-debug[165503]: init/main.c:1179 [main]initcall_blacklist =p "blacklisting initcall %s\012"
Apr 15 12:37:42 ip-10-110-212-88 sh[165504]: Ensuring /export/home is a symlink to /home...
Apr 15 12:37:42 ip-10-110-212-88 sh[165504]: /export/home is a mountpoint, unmounting...
Apr 15 12:37:42 ip-10-110-212-88 systemd[1]: export-home.mount: Succeeded.
Apr 15 12:37:42 ip-10-110-212-88 systemd[164396]: export-home.mount: Succeeded.
Apr 15 12:37:42 ip-10-110-212-88 sh[165504]: Removing existing /export/home directory...
Apr 15 12:37:42 ip-10-110-212-88 sh[165504]: Creating symlink: /export/home -> /home
Apr 15 12:37:42 ip-10-110-212-88 systemd[1]: Finished Delphix Appliance Platform Service.
Apr 15 12:37:42 ip-10-110-212-88 systemd[1]: Starting Delphix management service...
````

**Setup Present for the same setup**: [dlpx-qa-2025200-qar-170885-766f97e2.dlpxdc.co](dlpx-qa-2025200-qar-170885-766f97e2.dlpxdc.co)
**Upgraded from**: 2025.2.0.0
**Upgrade Tar:** [s3://dev-de-images/builds/jenkins-selfservice/appliance-build/develop/pre-push/4235/upgrade-artifacts/internal-qa.upgrade.tar](4235/upgrade-artifacts/internal-qa.upgrade.tar)
**An Engine retained in the Error State:** [dlpx-qa-2025200-qar-170888-766f97e2.dlpxdc.co](dlpx-qa-2025200-qar-170888-766f97e2.dlpxdc.co)

**Comment by justsanjeev** [comment]
Now here, we have created an alternative [PR](https://github.com/delphix/delphix-platform/pull/521) where we have added the logic to cleanup the `/export/home` mount point and creating the soft link `/export/home -> /home`. 

We have added another service [delphix-legacy-link.service](https://github.com/delphix/delphix-platform/pull/521/files#diff-bcb87666da339721dcc90e70264f681ff4e5a288a1469ce2fe608a47845a4d3b)



**This Build is Validated :** https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/10846/
**ManualUpgrade with an Unstructured database mounted:** ✅ 
**Manual Upgraded Engine, with deffered and Full Upgrade :** [dlpx-qa-2025200-qar-170959-766f97e2.dlpxdc.co](dlpx-qa-2025200-qar-170959-766f97e2.dlpxdc.co)
**Failed Pre Checkin Test Rebuild:** [pre_checkin](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/blackbox-self-service/171132/console)

File: debian/preinst (hunk 2 above)
Comment (justsanjeev) [reply debian/preinst]: Hi @sebroy

- We were working on a change to mount our home dataset on `/home` instead of `/export/home`. Initially, when the mount point was changed, there were some issues (or misconfigurations) during the process (whether it was an upgrade or building the image). This caused /home to be in use mid-process.

- The autofs service automatically mounted the /home file system. When we tried to mount it using /etc/fstab, it created a conflict and the build failed. To resolve this issue, we disabled autofs to auto mounting the dataset to  `/home` because we wanted the mounting to be handled via the static mount point defined in `/etc/fstab`.

- As we progressed with the changes across the repositories and ensured all configurations were correct, the automatic mounting issue was fixed. After thorough testing, we removed the block that disabled autofs, and our build was successful. Now This block of code , as per your comment is not required and we are removing that , Thanks for flagging this issue.


To validate this, I created a separate branch https://github.com/delphix/delphix-platform/commit/2cd0cbe0f285ad46d7b5e3d1a4f4371f7b267ef6 and removed the highlighted code block and Build.

https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/[appliance-build-orchestrator-pre-push/10878](appliance-build-orchestrator-pre-push/10878)/console

In the above build I see two failures- 
- Pre Checkin [171290](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/blackbox-self-service/171290)
- Upgrade [#7,412](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/blackbox-chained/7412)


**Solution to pre Checkin Failure**: The pre checkin failure is addressed in [qa-infra/pull/2270](https://github.com/delphix/qa-infra/pull/2270) and a rebuild of above job with our fix is [171474](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/blackbox-self-service/171474/)

**Solution to a test failure in Upgrade** - There is a test `test_validate_capacity` failed post upgrade. The test is failing due to https://perforce.atlassian.net/browse/DLPXQA-16439. Since, its a flaky test we rebuild the upgrade test here [blackbox-chained/7428/](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/blackbox-chained/7428/)


There is a manual upgrade also done and that was successful. We did this by adding a source and target environment with a unstructured database mounted on the target before upgrade.
**Upgraded Engine** : [dlpx-qa-2025200-qar-171213-766f97e2.dlpxdc.co](http://dlpx-qa-2025200-qar-171213-766f97e2.dlpxdc.co/login/index.html )- ✅ 

I am removing this code block(in this PR) and resolving this comment.

**Comment by justsanjeev** [comment]
Post changing my code as per above comments, I have Build [appliance-build-orchestrator-pre-push/10897](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/10897/) running.

File: files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml (hunk 4 above)
Comment (justsanjeev) [reply files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml]: I have posted new changes, those are dealing with the issues discussed here. Resolving this thread.

File: files/common/lib/systemd/system/delphix-legacy-link.service [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+      umount /export/home || { echo 'Failed to unmount /export/home'; exit 1; }; \
+    fi; \
+    if [ -e /export/home ]; then \
+      echo 'Removing existing /export/home directory...'; \
+      rm -rf /export/home; \
+    fi; \
+    if [ ! -d /export ]; then \
+      mkdir /export; \
+    fi; \
+    if [ -d /export ]; then \
```
sebroy: Lines 35-36 guarantee that `/export` exists.

files/common/lib/systemd/system/delphix-legacy-link.service
```diff
@@ -0,0 +1,45 @@
...
+
+[Service]
+Type=oneshot
+ExecStart=/bin/bash -c " \
+  if [ ! -L /export/home ]; then \
+    echo 'Ensuring /export/home is a symlink to /home...'; \
+    if mountpoint -q /export/home; then \
+      echo '/export/home is a mountpoint, unmounting...'; \
+      umount /export/home || { echo 'Failed to unmount /export/home'; exit 1; }; \
+    fi; \
```
sebroy: It is not safe to unmount all home directories while the system is running. The idea of creating boot-time logic for this is that after a reboot, `/export/home` cannot possibly be a mountpoint, as the filesystem will have been mounted at `/home` through the new entries in `/etc/fstab`.

files/common/lib/systemd/system/delphix-legacy-link.service
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+#
+
+[Unit]
+Description=Create /export/home symlink to /home after platform setup
+After=delphix-platform.service
+Requires=delphix-platform.service
+
+[Service]
+Type=oneshot
```
sebroy: A new systemd service may be overkill for this. A lighter-weight alternative would be to add another `ExecStart` script to the existing `delphix-platform.service`. The script could determine the right conditions in which to execute its logic (i.e. that home directories are mounted under `/home`), and not to run again if it's been run before.

sebroy (commented):
I think we're almost there with this. The only remaining detail is the mechanism used to run one-time boot-time logic.

files/common/lib/systemd/system/delphix-legacy-link.service
justsanjeev: Hi @sebroy In the latest revision. The logic is changed to check if 

- the mount point `/export/home` doesn't exist
- the Softlink /export/home is also not present

And then I am going ahead to remove the /export/home directory so that a softlink /export/home -> /home could be created.

As per my understanding this logic will be executed only after the reboot happens and it is also ensured that the mount point /export/home doesn't exist.

files/common/lib/systemd/system/delphix-legacy-link.service
justsanjeev: I have moved this to delphix-platform.service. Resolving this thread. Thanks for this suggestion.

files/common/lib/systemd/system/delphix-legacy-link.service
justsanjeev: Not creating the /export anymore. Resolving this

files/common/lib/systemd/system/delphix-platform.service
```diff
@@ -24,6 +24,23 @@ Before=rsync.service docker.service
...
+      fi; \
+      if [ ! -d /export ]; then \
+        mkdir /export; \
+      fi; \
+      if [ -d /export ]; then \
+        echo 'Creating symlink: /export/home -> /home'; \
+        ln -s /home /export/home; \
+      fi; \
+    fi; \
+  fi"
```
sebroy: Can we put this in a separate script (like the other `ExecStart` scripts) such as `home-symlink`? It will help readability (we should also take the opportunity to include comments to explain what this is doing).

files/common/lib/systemd/system/delphix-platform.service
```diff
@@ -24,6 +24,23 @@ Before=rsync.service docker.service
 Type=oneshot
 ExecStart=/var/lib/delphix-platform/ansible/apply
 ExecStart=/var/lib/delphix-platform/dynamic-debug
+ExecStart=/bin/bash -c " \
+  if ! mountpoint -q /export/home; then \
+    if [ ! -L /export/home ]; then \
+      echo 'Ensuring /export/home is a symlink to /home...'; \
+      if [ -e /export/home ]; then \
```
sebroy: I'm having a hard time imagining what might be in there. Did you find cases where the directory was non-empty?

files/common/lib/systemd/system/delphix-platform.service
```diff
@@ -24,6 +24,23 @@ Before=rsync.service docker.service
...
+        echo 'Removing existing /export/home directory...'; \
+        rm -rf /export/home; \
+      fi; \
+      if [ ! -d /export ]; then \
+        mkdir /export; \
+      fi; \
+      if [ -d /export ]; then \
```
sebroy: This check is redundant, as you created the directory if it didn't already exist above.

sebroy (changes requested):

files/common/lib/systemd/system/delphix-platform.service
justsanjeev: Added the logic to a separate file `files/common/var/lib/delphix-platform/export-home`.

files/common/lib/systemd/system/delphix-platform.service
justsanjeev: This issue is fixed.

files/common/lib/systemd/system/delphix-platform.service
justsanjeev: We discussed it in the meeting and this issues and as per explanation we agreed on the logic.

debian/postinst
```diff
@@ -124,6 +124,10 @@ configure)
 		update-initramfs -u -t -k "$(uname -r)"
 	fi
 
+  if [[ -f "/var/lib/delphix-platform/export-home" ]]; then
+    chmod 755 "/var/lib/delphix-platform/export-home"
+  fi
```
sebroy: This shouldn't be necessary. Note that there is no equivalent code needed for the other executables delivered by this package. You should be able to set the permissions on the file directly in the repo.

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,24 @@
...
+# This script ensures that the /export/home is a symlink
+# to /home.
+#
+
+if ! mountpoint -q /export/home; then
+  if [ ! -L /export/home ]; then
+    echo 'Ensuring /export/home is a symlink to /home...'
+    if [ -e /export/home ]; then
+      echo 'Removing existing /export/home directory...'
+      rm -rf /export/home
```
sebroy: This seems risky. If there's some bug in our code and we fail to unmount home directories for whatever reason, this will delete all home directories. At this point, if everything works as designed, `/export/home` should be an empty directory (it's the mountpoint for a filesystem that should have been unmounted already), and so `rmdir` should succeed.

sebroy (changes requested):

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,63 @@
...
+    echo '/home is also mounted. Since during the next boot /export/home will not be mounted, exiting safely.'
+    exit 0
+  else
+    echo '/home is not mounted. Aborting!!'
+    exit 1
+  fi
+fi
+
+# /export/home is not mounted, check if /home is mounted
+# Check if /home is mounted
```
sebroy: nit: "check if /home is mounted" appears twice in this comment

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,63 @@
...
+  echo "/home is mounted. Proceeding with /export/home cleanup."
+  # If /export/home exists
+  if [ -d /export/home ]; then
+    echo "/export/home exists. Attempting to remove it..."
+    rmdir /export/home 2>/dev/null
+    if [ $? -eq 0 ]; then
+      echo "/export/home directory removed successfully."
+    else
+      echo "/export/home is not empty. Please clean it manually before running this script."
+      exit 1
```
sebroy: Part of me thinks that we may end up with customers experiencing upgrade failures due to unexpected files in there (for reasons we haven't anticipated). To harden this part of the migration, perhaps we should move anything that's still in here elsewhere (with output identifying that we did this) and move on with removing the directory.

sebroy (changes requested):

files/common/var/lib/delphix-platform/export-home
justsanjeev: I have removed the extra comment.

files/common/var/lib/delphix-platform/export-home
justsanjeev: Thank you for the feedback regarding hardening the migration process. I have incorporated your suggestion to handle unexpected files in `/export/home` during upgrades.

The current implementation follows a defensive approach:
**First attempt**: Try rmdir /export/home to remove an empty directory
**Fallback with backup**: If rmdir fails (directory not empty), automatically move all contents to a timestamped backup location (/export/home.backup.YYYYMMDD_HHMMSS)
**User notification**: Clear output messages inform users about the backup location

This ensures:
**Zero data loss**: All unexpected files are preserved in the backup
**Upgrade resilience**: No upgrade failures due to non-empty /export/home
**Transparency**: Users are informed about what was moved and where
**Seamless experience**: Upgrade completes successfully regardless of directory state

The implementation addresses the concern about unexpected upgrade failures while maintaining data loss.

sebroy (approved):

debian/preinst
```diff
@@ -0,0 +1,42 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+case $1 in
+upgrade)
```
prakashsurya: I don't think I like doing this here, since the the location/directory of user's home directly isn't really something that's specific to the platform package..

I think a better place to do this might be in the `execute` upgrade-script, in appliance-build.

I'm not necessarily opposed to doing it here, just would like a high level explanation w.r.t. why it makes the most sense from a package arch. perspective, if we want to do it like this. E.g. why this package, vs. another package?

Alternatively, perhaps we should do this outside of upgrade execution context entirely (e.g. outside of the `execute` script, and outside of a package hook), and instead in the new `export-home` script, and run that script in a new standalone service.. one that's idempotent, and disables/masks itself after it's done with the migration..

files/common/lib/systemd/system/delphix-platform.service
```diff
@@ -24,6 +24,7 @@ Before=rsync.service docker.service
 Type=oneshot
 ExecStart=/var/lib/delphix-platform/ansible/apply
 ExecStart=/var/lib/delphix-platform/dynamic-debug
+ExecStart=/var/lib/delphix-platform/export-home
```
prakashsurya, line 27: Why do we want to run this script every time we start the platform service? At some point, any such migration from `/export/home` to `/home` should be "done", and running this script is no longer necessary; right?

If so, would it be better to execute this migration logic in a way that it can be explicitly disabled after it's completed? e.g. perhaps a new service (if it needs to be run outside of upgrade context), that can disable/mask itself after it runs to completion..?

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,69 @@
...
+		else
+			# If rmdir fails, it means /export/home is not empty
+			# Move contents to a backup location for a seamless user experience.
+			echo "/export/home is not empty. Moving contents to /export/home.backup before cleanup..."
+			backup_dir="/export/home.backup.$(date +%Y%m%d_%H%M%S)"
+			mv /export/home "$backup_dir"
+			if [ $? -ne 0 ]; then
+				echo "Failed to move /export/home contents to backup. Manual intervention required."
+				exit 1
+			fi
```
prakashsurya, line 49: this will result in downtime for the API, and an escalation.. is that a sufficient failure mode for the `mv` command failing here?

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,69 @@
...
+			echo "Contents of /export/home moved successfully to: $backup_dir"
```
prakashsurya, line 50: Perhaps I missed the explanation somewhere, if so, feel free to point me at it..

Why do we move to a backup location like this? what do we we expect to do with the back up location? if nothing, how do we manage it; and/or is it supposed to live indefinitely at the back up location without much of a purpose?

Tangential question, is the intention to clear user home directories after this migration? E.g. after the migration, a user's home directory will be empty, even if it previously was not empty?

Similarly, if there were user home directories present in `/export/home` (e.g. `/export/home/delphix` or `/export/home/masking`), do those need to be recreated in `/home` (e.g. `/home/delphix` or `/export/home/masking`)? I'm not sure there's anything on the system to recreate user home directories, but perhaps I'm wrong.

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,69 @@
...
+	fi
+fi
+
+# /export/home is not mounted, check if /home is mounted
+if mountpoint -q /home; then
+	echo "/home is mounted. Proceeding with /export/home cleanup."
+	# If /export/home exists
+	if [ -d /export/home ]; then
+		echo "/export/home exists. Attempting to remove it..."
+		rmdir /export/home 2>/dev/null
```
prakashsurya, line 37: Won't this fail after we move `/export/home` to the backup location below (e.g. fail the second time this script runs)?

E.g.
```root@ip-10-110-209-239:~# mkdir /export/home-test
root@ip-10-110-209-239:~# mv /export/home-test /export/home-test.backup
root@ip-10-110-209-239:~# rmdir /export/home-test
rmdir: failed to remove '/export/home-test': No such file or directory
root@ip-10-110-209-239:~# echo $?
1
```

In which case, each time this script runs (after the initial move), it will attempt to do the back up again, and result in the `mv` failing, and this script exiting with an error?

E.g.
```
root@ip-10-110-209-239:~# mv /export/home-test /export/home-test.backup
mv: cannot stat '/export/home-test': No such file or directory
root@ip-10-110-209-239:~# echo $?
1
```

Am I not following this logic correctly? Do we have a system where we've run with this script configured as-is, and done multiple service restarts or reboots, and verified the system/stack recovers each time successfully?

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,69 @@
...
+mkdir -p /export
+
+# Create symlink
+echo "Creating symlink: /export/home -> /home"
+ln -s /home /export/home
+if [ $? -eq 0 ]; then
+	echo "Symlink created successfully."
+else
+	echo "Failed to create symlink. Please check permissions and try again."
+	exit 1
```
prakashsurya: This is a failure that will lead to API downtime, and an escalation to resolve.. is this appropriate for such a failure? How do we expect such a failure to be resolved; e.g. what permissions would we need to check, and what would need to be retried?

debian/preinst
```diff
@@ -0,0 +1,42 @@
...
+	fs_tab=/etc/fstab
+
+	if grep -q "\/export\/home" "$fs_tab"; then
+		sed -i 's|/export/home|/home|g' "$fs_tab"
+		mount /home
+	fi
+
+	passwd_file=/etc/passwd
+	if grep -q "\/export\/home" "$passwd_file"; then
+		sed -i 's/\/export\/home/\/home/g' /etc/passwd
```
prakashsurya: do we need the `passwd_file` variable if we don't use it here?

debian/preinst
```diff
@@ -0,0 +1,42 @@
...
+case $1 in
+upgrade)
+	# Home directories were previously mounted under /export/home,
+	# and this was changed to /home. This is the upgrade logic that
+	# updates the /etc/fstab file to reflect that change.
+	# Home directories will be mounted in both /export/home and /home
+	# until the system is rebooted to ensure that running processes
+	# referencing the old /export/home paths continue to function
+	# while also enabling new logins under /home to work.
+	fs_tab=/etc/fstab
```
prakashsurya: IMO, the variable isn't necessary if we only use it in 2 consecutive lines of code..

debian/preinst
```diff
@@ -0,0 +1,42 @@
...
+
+	if grep -q "\/export\/home" "$fs_tab"; then
+		sed -i 's|/export/home|/home|g' "$fs_tab"
+		mount /home
+	fi
+
+	passwd_file=/etc/passwd
+	if grep -q "\/export\/home" "$passwd_file"; then
```
prakashsurya: Is the `grep` necessary? if `/export/home` doesn't exist, then `sed -i` won't modify the file, right? so it seems the `grep` can be removed without changing behavior.. unless I'm mistaken?

debian/preinst
```diff
@@ -0,0 +1,42 @@
...
+	# Home directories were previously mounted under /export/home,
+	# and this was changed to /home. This is the upgrade logic that
+	# updates the /etc/fstab file to reflect that change.
+	# Home directories will be mounted in both /export/home and /home
+	# until the system is rebooted to ensure that running processes
+	# referencing the old /export/home paths continue to function
+	# while also enabling new logins under /home to work.
+	fs_tab=/etc/fstab
+
+	if grep -q "\/export\/home" "$fs_tab"; then
```
prakashsurya: Is the `grep` necessary? if `/export/home` doesn't exist, then `sed -i` won't modify the file, right? so it seems the `grep` can be removed without changing behavior.. unless I'm mistaken?

files/common/var/lib/delphix-platform/export-home
```diff
@@ -0,0 +1,69 @@
...
+	echo "/home is not mounted. Aborting to avoid risk of data loss."
+	exit 1
+fi
+
+# Ensure /export exists
+mkdir -p /export
+
+# Create symlink
+echo "Creating symlink: /export/home -> /home"
+ln -s /home /export/home
```
prakashsurya: Why do we need this symlink?

files/common/var/lib/delphix-platform/export-home
prakashsurya: Thinking a bit more, why do we need to do this at all?

If our intention is to change the mountpoint of the home dataset, from `/export/home` to `/home`.. why do we do any data manipulation at all? Why not just swap the mountpoint option in `/etc/fstab`, and that's it..?

Sorry, am I overlooking something?

debian/preinst
justsanjeev: I am first trying to move this logic to appliance-build and making this atomic change and starting a build and will work on other comments.

justsanjeev: @prakashsurya As per suggestion

- This file has been removed and
- The logic is implemented in [appliance-build](https://github.com/delphix/appliance-build/pull/756/changes#diff-86a23a160c7b1a331c70e36b5fdb8be0ef45da285cc99c24a7ef97e49b6c5f8cR237)


Resolving this thread.

files/common/lib/systemd/system/delphix-platform.service
justsanjeev: Hi @prakashsurya,

Earlier we implemented a systemd service for this but then during review we agreed on a light weight alternative, Kindly refer  https://github.com/delphix/delphix-platform/pull/477#discussion_r2085686108

files/common/var/lib/delphix-platform/export-home
justsanjeev: While working on this change, I encountered a scenario where, after updating the dataset mount point, there was still data present under /export/home/delphix/.cache. This was caused by the installation of awscli in
https://github.com/delphix/appliance-build/pull/756/changes#diff-203852fe3ced2c760ad56ce84f1c641d4c7883ee4525ccf406798f32605d505cR37

We have already addressed that specific caching behavior in appliance-build with above change.

However, this also raised a broader concern that there could potentially be other user or customer-generated data present under /export/home that we may not be fully aware of.

During previous review discussions, it was suggested that instead of directly removing /export/home when it is non-empty, we should preserve the contents by moving them to a backup location. This provides a safer and more seamless user experience while also minimizing any potential risk of unintended data loss.

debian/preinst
justsanjeev: This file and the logic is deleted.

files/common/var/lib/delphix-platform/export-home
justsanjeev: Hi @prakashsurya , Lets take the scenario line by line -  
 
Lets say in first run (when /export/home is a real non-empty directory):

  - Line 12: [ -L /export/home ] → false (it's a real dir, not a symlink) → continues
  - Line 20: mountpoint -q /export/home → depends on state
  - Line 45: mv /export/home "$backup_dir" → succeeds, /export/home is gone
  - Line 63: ln -s /home /export/home → symlink is created

  Every subsequent run:
  - Line 12: [ -L /export/home ] → true (it IS a symlink now) → exits at line 14 with "Nothing to do."

  The mv at line 45 is never reached again after the symlink is created. So the script is idempotent in all cases.

debian/preinst
justsanjeev: I have deleted this file. This logic is no more present.

files/common/var/lib/delphix-platform/export-home
justsanjeev: Good point, @prakashsurya. 

Since export-home is executed by delphix-platform.service (running as root), a symlink creation failure here would be highly unusual — but you're right that
  the error message should be more actionable if it does occur.                                                                                                                             
                                                               
  I'll update the message to include context about what to check:                                                                                                                           
  - Whether /export exists and is writable                                                                                                                                                  
  - Whether a stale /export/home entry already exists (e.g. a broken symlink left from a previous partial run)                                                                              
  - The specific ln error output, so the operator knows exactly what failed

debian/preinst
justsanjeev: This file is permanently removed and the logic is taken care in the appliance-build.

debian/preinst
justsanjeev: This script is permanantly removed and the implementation is taken care in the `appliance-build` as per you initial comment.

files/common/var/lib/delphix-platform/export-home
justsanjeev: Hi @prakashsurya,
                                                                                                                                                                                            
  You raise a valid concern. A failure at this point would require manual intervention, but I believe this scenario is genuinely unlikely in practice. Let me share the testing I carried out to validate the upgrade path:                                                                                                                                                         
                  
  Test setup (pre-upgrade):                                                                                                                                                                 
  - Created /export/home/delphix/data/data.txt with content
                                                                                                                                                                                            
  Post-upgrade observations:
  - /export/home is now a symlink: lrwxrwxrwx 1 root root 5 May 6 11:11 home -> /home                                                                                                       
  - All data is preserved and accessible at /home/delphix/data/data.txt                                                                                                                     
  - Both the old path (/export/home/delphix/) and new path (/home/delphix/) resolve correctly

````
delphix@ip-10-110-238-129:~$ ls -ltr /export/home/delphix/
total 2
drwxr-x--- 2 delphix staff 3 May  6 09:07 data
delphix@ip-10-110-238-129:~$ ls -ltr data/
total 1
-rw-r----- 1 delphix staff 8 May  6 09:07 data.txt
delphix@ip-10-110-238-129:~$ Connection to dlpx-qa-2025300-qar-215588-766f97e2.dlpxdc.co closed by remote host.
Connection to dlpx-qa-2025300-qar-215588-766f97e2.dlpxdc.co closed.
sanjeev.rohilla@APAC-SANJEEV % ssh delphix@dlpx-qa-2025300-qar-215588-766f97e2.dlpxdc.co
(delphix@dlpx-qa-2025300-qar-215588-766f97e2.dlpxdc.co) Password: 
delphix@ip-10-110-238-129:~$ cd /export/
delphix@ip-10-110-238-129:/export$ ls -ltr
total 1
lrwxrwxrwx 1 root root 5 May  6 11:11 home -> /home
delphix@ip-10-110-238-129:/export$ 
delphix@ip-10-110-238-129:/export$ cd /home/
delphix@ip-10-110-238-129:/home$ ls -ltr
total 3
drwxr-xr-x 3 cli     staff  7 May 21  2025 cli
drwxr-xr-x 4 delphix staff 10 May  6 11:10 delphix
delphix@ip-10-110-238-129:/home$ cd delphix/
delphix@ip-10-110-238-129:~$ ls -ltr
total 2
drwxr-x--- 2 delphix staff 3 May  6 09:07 data
delphix@ip-10-110-238-129:~$ ls -ltr data
total 1
-rw-r----- 1 delphix staff 8 May  6 09:07 data.txt
delphix@ip-10-110-238-129:~$ pwd
/home/delphix
delphix@ip-10-110-238-129:~$ 
````

 Regarding the mv failure on lines 46-49: since this script runs as root, the most realistic cause of a mv failure would be insufficient disk space on /export. In that case, there is no safe automated recovery — proceeding without the backup risks data loss, so exiting with an error is the conservative and correct choice. Manual intervention (freeing disk space and re-running) is a reasonable resolution path.                                                                                                                                              
                  
That said, I'm happy to improve here with further suggestion.

SetUP : 
````
Upgrade from `2025.3.0.0` to develop

Upgrade.tar:  `s3://dev-de-images/builds/jenkins-selfservice/appliance-build/develop/pre-push/6501/upgrade-artifacts/internal-qa.upgrade.tar` 

Build: https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/13945/console

Appliance : http://dlpx-qa-2025300-qar-215588-766f97e2.dlpxdc.co 🟢 
````

File: files/common/var/lib/delphix-platform/export-home (hunk 25 above)
Comment (justsanjeev) [reply files/common/var/lib/delphix-platform/export-home]: Hi @prakashsurya ,
                                                                                                                                                                                            
The symlink serves as a compatibility bridge. The path /export/home is referenced across multiple repositories (dlpx-qa-gate, devops-gate, qa-infra), and updating all of them atomically alongside this change would significantly broaden the scope of this PR and increase the risk surface.                                                                    
   
  By creating /export/home -> /home, we ensure that:                                                                                                                                        
  - Existing tooling and infrastructure continue to work without modification
  - This change remains self-contained and easier to review, test, and roll back if needed                                                                                                  
  - The broader cleanup of those references can be done incrementally in follow-up PRs, with each repo updated and tested independently
                                                                                                                                                                                            
  This approach is a common pattern for path migrations — move first, update consumers gradually, remove the compatibility shim last once all references are cleaned up as its suggested by the reviewers.

File: files/common/var/lib/delphix-platform/export-home [hunk 26]
```diff
@@ -0,0 +1,68 @@
...
+	echo "/home is not mounted. Aborting to avoid risk of data loss."
+	exit 1
+fi
+
+# Ensure /export exists
+mkdir -p /export
+
+# Create symlink
+echo "Creating symlink: /export/home -> /home"
+if ln -s /home /export/home; then
```
prakashsurya, line 63: can you remind me why we want to preserve this symlink?

PR #485, opened 2024-06-17
DLPX-91396 All emails generated by specific cron job, are redirected to the user SMTP.

prakashsurya:
<details open>

Emails generated by "cron" are being sent externally via SMTP.

</details>

<details open>

Disable "nullmailer" service, so it doesn't send email's externally.

</details>

### DIFF
--- a/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
+++ b/files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
@@ -705,12 +705,3 @@
     block: |
       # Set default umask value.
       umask 027
-
-- file:
-    src: /bin/true
-    dest: /usr/lib/nullmailer/true
-    state: link
-
-- copy:
-    dest: "/etc/nullmailer/remotes"
-    content: "localhost true"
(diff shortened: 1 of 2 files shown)


files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -705,3 +705,14 @@
     block: |
       # Set default umask value.
       umask 027
+
+- file:
+    src: /bin/true
+    dest: /usr/lib/nullmailer/true
+    owner: foo
+    group: foo
```
sebroy: Should this be an actual user/group?

prakashsurya: shoot, I removed it, then editor crashed and didn't notice that.. let me remove it again :p

prakashsurya:
Got this idea from: https://github.com/bruceg/nullmailer/issues/64

prakashsurya:
Actually, looks like I might be able to just disable the service, which seems like a cleaner solution to me; updated the PR to reflect that.

david-mendez1 (approved):

sebroy:
Sounds good. It might also be worth looking into the cron configuration. Since we know sending email should never happen, cron may also have a configuration to suppress email. I think CRONDARGS can be set so that cron sends output to syslog instead of email?

sebroy (approved):

PR #501, opened 2024-11-05
APIGW-10470 Delphix startup screen should be updated for DCT appliances

eyalkaspi-delphix:
In this PR, we are updating the delphix-startup-screen to work for DCT appliances by making the following changes:

First, we are displaying the packaged-app-version (i.e DCT version) instead of appliance version when it is available.

Then, we are avoiding an exception for network intefaces without address (using python `get` instead of `[]`).

We are hiding the DCT appliance network interfaces (docker0 and docker bridges). This is currently done by name matching, if anyone knows of a better way to identify them that would be great.

Finally, it appears that the list of systemd services now may exceed the available vertical space. While we might want to redesign the screen, or identify which services are critical or not to be visible, I've implemented a much simpler fix by catching the exception and simply moving on (so for DCT appliances the state of the last service, in the order returned by the `systemctl show` command, is missing).

A dcol2 DCT appliance (note the DCT version, long list of services, and only the external IP address is visible)

image

A dcol2 standard engine
image

### DIFF
--- a/files/common/usr/bin/delphix-startup-screen
+++ b/files/common/usr/bin/delphix-startup-screen
@@ -126,12 +126,21 @@ def load_header(stdscr) -> Any:
     Display the header information for the main screen.
     """
     WIN_LEN, WIN_HT = set_common_variables(stdscr)
-    cmd = ['get-appliance-version', '--patch']
-    cp = subprocess.run(cmd,
+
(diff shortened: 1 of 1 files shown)


delphixBrandonB (approved):
Looks good to me. I'm using very similar logic to hide the network interfaces.

amit-rathaur (approved):

PR #514, opened 2025-03-18
DLPX-93729 LTS 24.04: unable to upgrade emacs-nox package

prakashsurya:
<details open>
Upgrade is failing due to errors when running emacs post-install hooks.
</details>

<details open>
The solution taken here is to remove the emacs package(s) from the appliance.
</details>


prakashsurya:
I'll go ahead and merge this on the project branch, so I can verify it even works.. we can always back it out if we decide we need to keep emacs on the appliance..

PR #527, opened 2025-04-30
DLPX-94085 LTS 24.04: update delphix-platform for Ubuntu 24.04 appliance

prakashsurya:
<details open>
We need to port the changes made on the os-upgrade branch to develop, for the 2025.4.0.0 release.
</details>

<details open>

- Note, this will need to be integrated simultaneously with all the other LTS changes on other repositories, after code-complete.
- This PR has all the changes made on the os-upgrade branch, cherry-picked and squashed into a single commit for integration on the develop branch.
  - DLPX-93974 LTS 24.04: dailyaidecheck.service is enabled and will fail when started by dailyaidecheck.timer
  - DLPX-93961 LTS 24.04: getty services failed due to missing python3-netifaces
  - DLPX-93748 LTS 24.04: various services masked, causing enable to fail during package installation
  - DLPX-93749 LTS 24.04: sshd service fails to restart via virtualization's postinst hook (#517)
  - DLPX-93731 LTS 24.04: delphix-platform package hook fails due to delphix.target being masked (#516)
  - DLPX-93731 LTS 24.04: delphix-platform package hook fails due to delphix.target being masked (#515)
  - DLPX-93729 LTS 24.04: unable to upgrade emacs-nox package (#514)
  - DLPX-93685 LTS 24.04: Upgrade fails due to libkdumpfile conflict with upstream libkdumpfile10 package (#512)
  - DLPX-93639 LTS 24.04: Upgrade fails due to drgn conflict with upstream python3-drgn package (#511)
  - DLPX-93500 LTS 24.04: Disable unneeded pmcd Performance Co-Pilot services (#510)
  - CP-9872 nfs-utils - remove obsolete 10-lockd.conf (#509)
  - DLPX-93328 LTS upgrade: delphix-platform - Replace ntp with ntpsec as part of DLPX-93325 (#508)
  - DLPX-92876 delphix-platform dependencies need to be updated to allow installation on 24.04 (#502)

</details>


.github/workflows/main.yml
```diff
@@ -2,11 +2,12 @@ on: [push, pull_request]
 
 jobs:
   check-packages:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
sebroy: This doesn't look correct to me (here and throughout this file). Changing this from `ubuntu-latest` (which basically always should work) to `ubuntu-24.04` presumably puts us on a never-ending tech-debt treadmill that forces us to update this every few years. Am I wrong about that?

prakashsurya: yea, I agree.. let me see why it did that.. and I'll change it back..

prakashsurya: looks like this comes from: https://github.com/delphix/delphix-platform/pull/502

prakashsurya: I could see it going both ways.. i.e. if we use latest, then things can break without us due to the image changing.. if we use a pinned version, things should remain stable as the image doesn't change, but we have to play the game of updating it every-so-often.. I'm good changing it back to latest.. think that should be fine..

sebroy (approved):

tonynguien (approved):

PR #541, opened 2025-08-01
DLPX-94983 Alternative changes for the DLPX-94879

justsanjeev:
<details open>

Provide a clear description of the high-level problem you are trying to
solve. The problem statement should be written in terms of a specific
symptom that affects users or the business. The problem statement should
not be written in terms of the solution. If possible, include a minimal
reproducible example (MRE) with steps to reproduce, expected results,
and actual results.
</details>

<details open>

Provide a clear description of the high-level solution you have chosen.
If there were other possible solutions that you considered and rejected,
mention those along with the corresponding reasoning. Do not describe
implementation details when writing about the solution; these should go
into the implementation section instead.
</details>

<details>

Provide a clear description of how this change was tested. At minimum
this should include proof that a computer has executed the changed
lines. Ideally this should include an automated test or an explanation
as to why this pull request has no tests.
</details>

### DIFF
--- a/debian/preinst
+++ b/debian/preinst
@@ -1,0 +1,18 @@
+#!/bin/bash -eux
+#
+# Copyright 2025 Delphix
+#
+
+case $1 in
+upgrade)
+  # Create the working directory for the dxrt utility if it doesn't exist
+  if [ ! -d "/var/delphix/sfdc/dxrt" ]; then
+    mkdir -p "/var/delphix/sfdc/dxrt" || die "failed to mkdir /var/delphix/sfdc/dxrt"
+    chmod o+r "/var/delphix/sfdc/dxrt" || die "Failed to add permissions to /var/delphix/sfdc/dxrt"
+    chown masking:staff "/var/delphix/sfdc" || die "Failed to chown /var/delphix/sfdc"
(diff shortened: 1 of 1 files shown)


debian/preinst
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -eux
+#
+# Copyright 2025 Delphix
+#
+
+case $1 in
+upgrade)
+  # Create the working directory for the dxrt utility if it doesn't exist
+  if [ ! -d "/var/delphix/sfdc/dxrt" ]; then
+    mkdir -p "/var/delphix/sfdc/dxrt" || die "failed to mkdir /var/delphix/sfdc/dxrt"
```
prakashsurya, line 10: I'm confused, these commands are specific to the masking package, so why are we putting this in a masking-agnostic package, vs. putting them in the masking package?

prakashsurya (changes requested):

debian/preinst
justsanjeev: @Prateek-kumar-delphix I am closing this review as its no more required.

justsanjeev:
This change is no more required.

PR #554, opened 2026-01-14
CP-13126 Update delphix-platform to install docker from official docker repo instead of docker.io

SumoSourabh:
<details open>

We are using docker.io which is the default ubuntu package for installing docker, but the version is not in sync with the actual docker-ce version.

To be in sync with the docker-ce version we need to install the docker from official docker repositories. The docker repositories are already downloaded to ubuntu mirror and are being pulled during sudo apt-get update using https://github.com/delphix/linux-pkg/pull/381

</details>

<details open>

Updated the docker installation with docker-ce. Details about each package

- docker-ce -> Needed for the installation of docker community addition.
- docker-ce-cli -> Needed for running commands via CLI
- containerd.io -> docker-ce and docker-ce-cli are dependent on the components for containerd.io
- nftables -> With the release of docker version 29.x.x, it is necessary requirement as docker had started supporting iptables rules via nftables also.

Also, once all the components are installed, we have to start the service via systemctl command. What I found out is that first time installation on a machine with docker.io installed, the service needed to be started multiple times. So, I have to add retry mechanism
  - Also, running the ansible file again does not have the same issue as the docker is started without issues.
  - So, we should not face the same issue once the machine has docker-ce installed.

I tried multiple approaches like stopping the docker service, uninstalling all docker components and then starting the service and waiting for it to come online. But faced the same issue again and again.

</details>

<details open>
<summary><h2> Merging Details </h2></summary>
Still checking if it can be merged alone..
</details>

<details open>

- Tried on a server by cloning dlpx-internal-buildserver-develop.
- As per screenshot, running it first time has retries.
- Running it again does not have retries.
Screenshot 2026-02-09 at 7 49 52 PM

</details>

### DIFF
--- a/bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
@@ -21,9 +21,19 @@
 - apt:
     name: "{{ item }}"
     state: present
-  with_items:
-    - docker.io
+  loop:
+    - nftables
+    - docker-ce
+    - docker-ce-cli
+    - containerd.io
     - python3-docker
+
+- name: Execute 'systemctl start docker' command
+  ansible.builtin.command: systemctl start docker
+  register: docker_start_result
+  retries: 10
+  delay: 30
+  until: docker_start_result.rc == 0
 
 - docker_image:
     build:


bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -22,9 +22,43 @@
     name: "{{ item }}"
     state: present
   with_items:
-    - docker.io
+    - nftables
+    - docker-ce
+    - docker-ce-cli
+    - containerd.io
+    - docker-compose-plugin
```
sebroy: What is the purpose of all of these dependencies?

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -22,9 +22,43 @@
...
     - python3-docker
 
+- name: Execute 'sudo systemctl start docker' command
+  ansible.builtin.command: sudo systemctl start docker
+  register: docker_status_output
+  changed_when: false # This command only retrieves info, doesn't change state
+  retries: 10
+  delay: 30
+  until: docker_status_output.rc == 0
```
sebroy: Can you add a comment explaining why this is needed?

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -22,9 +22,43 @@
...
+  register: docker_version_output
+  changed_when: false # This command only retrieves info, doesn't change state
+
+- name: Print the full command output
+  ansible.builtin.debug:
+    var: docker_version_output.stdout
+
+- name: Print only the Docker client version
+  ansible.builtin.debug:
+    msg: "Docker Client Version: {{ docker_version_output.stdout_lines[1] }}"
```
sebroy: What is all of this for?

sebroy:
nit: delphix is mis-spelled in your PR description.

Can you fill in the PR description with meaningful context and testing details please? It's important to note that this is not a stand-alone PR, and must be merged along with PRs in other repos as well.

sebroy (changes requested):

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -22,9 +22,19 @@
     name: "{{ item }}"
     state: present
   with_items:
-    - docker.io
+    - nftables
+    - docker-ce
+    - docker-ce-cli
+    - containerd.io
     - python3-docker
```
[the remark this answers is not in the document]
SumoSourabh: The distro repositories are already added as part of some other project which is pulled when running the ansible tasks.

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
SumoSourabh: Added the details in the description.

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
SumoSourabh: Added the details in the description.

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
SumoSourabh: This was only done for testing purpose. Removed it.

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -21,10 +21,20 @@
 - apt:
     name: "{{ item }}"
     state: present
-  with_items:
-    - docker.io
+  loop:
+    - nftables
+    - docker-ce
+    - docker-ce-cli
+    - containerd.io
```
sebroy, line 28: If we don't directly depend on this, but rather it's a transitive dependency of `docker-ce`, then we shouldn't include it here. It will get installed via that package dependency.

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -21,10 +21,20 @@
 - apt:
     name: "{{ item }}"
     state: present
-  with_items:
-    - docker.io
+  loop:
+    - nftables
```
sebroy, line 25: I'm still unclear as to why we are introducing `nftables` here. This ansible playbook is used to configure the build server for building the `delphix-platform` package. The packages listed here are only installed on the build server when building (using the `linux-pkg` build job for `delphix-platform`), and do not have any impact on the packages included in the product. Do we really require installing a firewall for building a package?

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -21,10 +21,20 @@
...
+    - docker-ce-cli
+    - containerd.io
     - python3-docker
 
+- name: Execute 'systemctl start docker' command
+  ansible.builtin.command: systemctl start docker
+  register: docker_start_result
+  retries: 10
+  delay: 30
+  until: docker_start_result.rc == 0
```
sebroy, line 36: What's the cause of the failures to start the docker service? This should be in the service log (`journalctl -u docker`). Do we know the root-cause, and have determined that a retry is the best way to work around that, or do we not know the root-cause?

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -21,10 +21,20 @@
 - apt:
     name: "{{ item }}"
     state: present
-  with_items:
-    - docker.io
+  loop:
+    - nftables
+    - docker-ce
+    - docker-ce-cli
```
sebroy, line 27: You mentioned in the PR description that docker commands are used. What commands from the `docker-ce-cli` does the `delphix-platform` package build use?

sebroy, line 27: Is this really needed for the `delphix-platform` package build? What commands do we use?

SumoSourabh:
As delphix-platform docker dependency has been removed as part of https://github.com/delphix/delphix-platform/pull/555, this PR has become redundant and thus closing it.

PR #555, opened 2026-02-10
DLPX-96562 eliminate docker from the delphix-platform package build

sebroy:
See Jira for details. This PR is inter-dependent with linux-pkg PR https://www.github.com/delphix/linux-pkg/pull/382 and must be tested and integrated together.

linux-pkg delphix-platform build: https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/develop/job/build-package/job/delphix-platform/job/pre-push/1117/


.github/workflows/main.yml
```diff
@@ -1,13 +1,6 @@
...
-  check-packages:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - run: docker build -t delphix-platform:latest docker
-      - run: ./scripts/docker-run.sh make packages
-
   check-shellcheck:
     runs-on: ubuntu-latest
     steps:
```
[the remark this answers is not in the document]
sebroy: This was removed because package build is already tested in the ab-pre-push jenkins job that gets triggered as another PR check. This is thus redundant.

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Delphix
+# Copyright 2018, 2026 Delphix
```
prakashsurya: I'm OK with this as-is, as incremental improvement.. but mentioning, since we're here.. we could likely remove this ansible playbook as a whole, and put the packages listed below as `Build-Depends` in the `debian/control.in` file (or something like that, I might have the specific details wrong)..

I believe, linux-pkg already has logic backed in to install packages via those `Build-Depends`..

That'd also remove our reliance on ansible for the build (for something as simple as running `apt-get install ...`)..

But again, I'm good as-is, just mentioning some more clean up we _could_ do, if we wanted to..

.github/workflows/main.yml
```diff
@@ -1,13 +1,6 @@
 on: [push, pull_request]
 
 jobs:
-  check-packages:
```
prakashsurya, line 4: also mentioning for historical purposes.. we now have the `build-package` PR check, that ties into Jenkins and builds the package using our jenkins jobs (i.e. the same job that builds the package we install into the product)..

so we haven't lost any test coverage here.. we're just removing the duplicate (and inferior) coverage provided by this action..

prakashsurya (approved):
thank you, LGTM

bootstrap/roles/delphix-platform.bootstrap/tasks/main.yml
sebroy: Yes, I like that, and was pondering the same thing as I was re-reviewing these changes. Let me try that and update shortly.

prakashsurya (approved):

Makefile
```diff
@@ -26,13 +26,25 @@ VERSION := $(shell date '+%Y.%m.%d.%H')
 	check \
 	package
 
+build-deps:
+	#
+	# Create a fake debian/control file with a bogus platform for the
+	# putpose of installing build dependencies. The control files used for
```
lyriclake: s/putpose/purpose

lyriclake (approved):

PR #573, opened 2026-09-18
APIGW-27732 PUC upgrade image size may increed appliance allowance

eyalkaspi-delphix:
<details open>
We are working on merging the PUC appliance with the DCT appliance for the next DCT major release (2026.6) and anticipate the upgrade image
size of the combined appliance might exceed the current allowance of 15G.

</details>

<details>
<summary><h2> Evaluation </h2></summary>
All Delphix appliances currently have a 30g quota set to the /var/dlpx-update filesystem which is the location where upgrade images are
uploaded and expanded. Because this is a two setup process, the 30g allowance only allows upgrade images of size of up to 15g,
which is enforced by the mgmt-stack when the file is being uploaded (blocked if the file size is greater than half of the available space).

The DCT 2026.5 image size is already 12.Xg, and while we don't have a combined upgrade image size yet we know the DCT containers which
form the PUC application are currently of size ~6GB.

There will be an effort to reduce the combined image size, which could take the form of:
1. Being more efficient when packaging the application (better reuse of docker layers, sharing base image)
2. Removing virtualization specific packages which aren't necessary for DCT appliances (example the hosts jdk seem to take > 1GB, delphix-rust may not be necessary?)
Moreover, efforts could be made to combine the upload and unpack into a single step to avoid the double space issue.

But given that the upload of the upgrade image is implemented on the pre-upgrade version of the appliance code, we don't have any gurantee today
that the 2026.6 upgrade image size will be of size < 15GB.

</details>

<details open>

If we are unable to contraint the upgrade image size below 15GB for 2026.6 and later, we will need the pre-upgrade version to allow bigger image sizes. While we can't
go back in the past, we propose in this change to update the allowance from 15GB to 20GB (or the quota from 30GB to 40GB) to avoid being blocked later on. => Edit: later changed to 45G (22.5G image size)

We could not understand how the quota size was determined initially, but this seemed like an arbitrary value anyways, so we can't tell the exact impact of changing this
value. We know appliances must retain some zpool disk space (for core dumps, support bundles, ....), but can't tell exactly how much is too much for upgrade images.
This change, being a quota update, will not cause any of these issues unless the upgrade image size will actually exceed what is acceptable, which will be tested later on.
</details>

<details>
Build a dct-internal dcOA image with these changes and checked the quota

```
delphix@ip-10-110-192-164:~$ zfs get quota /var/dlpx-update
NAME          PROPERTY  VALUE  SOURCE
rpool/update  quota     40G    local
```

</details>


jmathur427 (approved):
If you have it handy, could you please share the reference or steps for building a dct-internal dcOA image with our changes? It would be useful for future reference.

files/common/var/lib/delphix-platform/ansible/10-delphix-platform/roles/delphix-platform/tasks/main.yml
```diff
@@ -110,7 +110,7 @@
     extra_zfs_properties:
       mountpoint: /var/dlpx-update
       compression: gzip
-      quota: 30g
+      quota: "{{ '40g' if variant is search('dct') else '30g' }}"
```
amit-rathaur: I think we discussed to make it 45g on every channel.

eyalkaspi-delphix: done

eyalkaspi-delphix:
> If you have it handy, could you please share the reference or steps for building a dct-internal dcOA image with our changes? It would be useful for future reference.

```
git ab-pre-push -v internal-dct -i internal-dct --no-tests
```

